PR #1033, opened 2014-09-26
[WIP] MapFish Print Server v3

procrastinatio:
This very WIP, it's only install the java .war in tomcat.

Documentation (broken on the main site)
http://mf-chsdi3.dev.bgdi.ch/~ltmom/mapfish-printV3/#/overview


procrastinatio:
So what do we do with Print v3 ?

kandre:
I suggest I look first at emapis to be more comfortable with it and then come back to you

procrastinatio:
Canton Basel-Land uses this Mapfish Print v3 (at least two presentation from c2c):

http://geoview.bl.ch/  (bottom-left to print)

procrastinatio:
Well, at last we have WMTS support...

And then in _http://mf-chsdi3.dev.bgdi.ch/{kurzel}/print/_, choose _Print app/Example_ `geoadmin3` (you may edit the `spec`before sending it to the server...

procrastinatio:
This is no longer needed

PR #1037, opened 2014-09-30
Testing varnish config

procrastinatio:
This PR defines the ressources that should be "protected" by varnish.

To test varnish configuration for production:

     ./run_nose.sh -p -a

with `-a`for all tests (varnish and mapproxy are deactivated by default) and `-p` for using the procution settings.


procrastinatio:
This needs to be hosted or used a proxy outside the KOMBV, because of the BIT Proxy IP.

gjn:
We can ssh into ssh0a, copy the relevant stuff and run the tests from there. I think the testscript(s) of our test project should take care of that for the varnish tests (probably all e2e tests should be run on both sides).

procrastinatio:
I leave the test here, but use a custom `http_proxy` for the requests testing varnish. I'll update the `http_proxy` as soon as I'll get the definitive address.

procrastinatio:
Using the new frankfurter proxy: ec2-54-93-109-29.eu-central-1.compute.amazonaws.com:80

chsdi/tests/mapproxy/test_wmtscapabilities.py
```diff
@@ -0,0 +1,112 @@
+#-*- coding: utf-8 -*-
+
+import sys
+import os
+import random
+from unittest import TestCase
+from testconfig import config as tc
+
```
gjn: `import requests` is missing here, meaning you can't run chsdi/tests/mapproxy/test_wmtscapabilities.py alone.

chsdi/tests/mapproxy/test_wmtscapabilities.py
```diff
@@ -0,0 +1,112 @@
...
+             4326: [(10, 6, 4), (15, 222, 123)],
+             4258: [(10, 6, 4), (15, 222, 123)]
+             }
+
+
+class TestWmtsCapabilitiesView(TestCase):
+
+    def setUp(self):
+        try:
+            self.url = "http:" + tc['vars']['mapproxy_url']
```
gjn: `mapproxy_url` is nowhere defined. Did you mean to use `mapproxyhost` here?

chsdi/tests/mapproxy/test_wmtscapabilities.py
```diff
@@ -0,0 +1,112 @@
...
+        except KeyError as e:
+            self.url = os.getenv('conf_to_use', "http://api3.geo.admin.ch")
```
gjn: The name 'conf_to_use' is confusing to me. Why not name it mapproxyhost? Or wmts_url?

chsdi/tests/mapproxy/test_wmtscapabilities.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def test_axis_order(self):
+        from urlparse import urlparse
+        import xml.etree.ElementTree as etree
+
+        for epsg in EPSGS:
+            resp = requests.get(self.url + '/rest/services/api/1.0.0/WMTSCapabilities.xml', params={'epsg': epsg, '_id': self.hash()},
+                                headers=self.get_headers())
+
+            print resp.encoding
```
gjn: This should be removed.

chsdi/tests/mapproxy/test_wmtsgettile.py
```diff
@@ -0,0 +1,86 @@
...
+    assert bits % 8 == 0
+    return os.urandom(bits / 8).encode('hex')
+
+
+def get_headers():
+    return {'User-Agent': 'WMTS Unit Tester v0.0.1', 'Referer': 'http://unittest.geo.admin.ch'}
+
+
+def check_status_code(path):
+    print path
```
gjn: Should be removed.

chsdi/tests/varnish/test_varnish.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def hash(self, bits=96):
+        assert bits % 8 == 0
+        return os.urandom(bits / 8).encode('hex')
+
+    def timestamp(self):
+        return random.randrange(20140101, 20141001)
+
+    def setUp(self):
+        os.environ["http_proxy"] = tc['vars']['http_proxy']
```
gjn: Can't we set a default proxy? So we could run it without loading a configuration?

e.g. running `buildout/bin/nosetests chsdi/tests/varnish/test_varnish.py' should also work, no?

gjn:
A few comments:

1) For such e2e tests, why not just use the official urls, as published in www.geo.admin.ch and http://api3.geo.admin.ch/services/sdiservices.html#wmts? If those are ok, we know backend/mapproxy are ok, no?
2) I couldn't run `chsdi/tests/mapproxy/test_wmtsgettile.py` succesfully. Had 306 failures.
3) For the tiles, shouldn't we mix the official adresses wmts, wmts1, wmts2, etc...(because that's what we need to support)?
4) We should be able to exclude those tests from standard nosetest runs (without any parameters, or specify to ignore them) as they are to be executed _after_ deploy. All other nosetests can be run _before_ the deploy. Then we need to adapt the deploy scripts accordingly (and also jenkis integration). If you don't mind, I'll create a PR to this branch for this.

gjn:
jenkins test

procrastinatio:
@gjn 

1) We test the varnish settings also on _dev_ and _int_ before _prod_, but _prod_ is by far more complicated than the other two (because of aliases). But why not.

2) The WMTS Capabilities reads the BOD, Mapproxy uses a generated config from the BOD. If both do not match, you have errors.

3) OK. Plus the various aliases.

4) For Varnish, sure it has to be done after deployment, and after every varnish changes. For WMTS, ideally it should be run **before**, but because of the staging mess, it has to be run **after**, for check.

gjn:
@procrastinatio I've create a PR to address parts of point 4. See https://github.com/geoadmin/mf-chsdi3/pull/1314

We can discuss the rest tomorrow. I want this in!

chsdi/tests/mapproxy/test_wmtscapabilities.py
```diff
@@ -0,0 +1,112 @@
...
+
+TILES_IDX = {3857: [(7, 67, 45), (10, 533, 360)],
+             21781: [(17, 5, 6), (18, 11, 13)],
+             2056: [(17, 5, 6), (18, 11, 13)],
+             4326: [(10, 6, 4), (15, 222, 123)],
+             4258: [(10, 6, 4), (15, 222, 123)]
+             }
+
+
+class TestWmtsCapabilitiesView(TestCase):
```
gjn: I think this could be moved into the 'integration' tests (https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/tests/integration/test_wmtscapabilities.py#L6) as we don't test anything varnish related. It's fast as well, so it can be run on standard nose tests.

chsdi/tests/mapproxy/test_wmtscapabilities.py
```diff
@@ -0,0 +1,112 @@
...
+    def check_status_code(self, url, referer, code):
+        headers = None
+        if referer:
+            headers = {'Referer': referer}
+            resp = requests.get(url, params={'_id': self.hash()}, headers=headers)
+        else:
+            resp = requests.get(url, params={'_id': self.hash()})
+        assert int(resp.status_code) == code
+
+    def test_bad_referer_get_capabilties(self):
```
gjn: Use address official (without rest/...)

chsdi/tests/mapproxy/test_wmtscapabilities.py
procrastinatio: OK, moved to *integrations* tests.

procrastinatio:
Updated according to comments/discussion:
* Using the only `run_nose.sh` with a supplementry parameter `-a` (for all tests)
* Moved some test to `integration`
* Testing only the official Urls

procrastinatio:
jenkins test

gjn:
Awewome I'll merge. Thanks alot for this much needed effort.

@daguer @alcapat @ponceta We will use these new tests on our ci instance.

PR #1044, opened 2014-10-03
[WIP] Static bod

gjn:
This is a proof of concept to use static file(s) instead of the db/bod for our metadata.

1) export_bod2json.py is the script that creates the static json files from the bod database. The source of truth remains the bod
2) The static files can be manaully updated with the `updatebod` target. `buildout/bin/buildout -c buildout_lt***.cfg install updatedbod`
3) On wsgi startup, the static files are read and filled into an in-memory sqlite database.
4) All requests to the bod are done on the sqlite database. This allows us to use basically the same model definitions as before. In fact, it's possible to switch between file-based backend and postgres backend by changing a variable in the bod.py file (https://github.com/geoadmin/mf-chsdi3/blob/gjn_static_bod/chsdi/models/bod.py#L18)

We need to discuss
1) if we want to go in this direction
2) what the dev process will be for those metadata (branches, deploys, etc.)
3) What do we do with auto-update tables in the bod (some timestamps are updated on prod directly...)

Input welcome.

Stuff:
- sorting in german does not work for umlaute. Remove_accents probably not correct.
- gain in preformance is about 30ms (basically, the round-trip to the db server, which is not needed anymore) for any service using the bod


gjn:
jenkins test

gjn:
jenkins test

gjn:
jenkins test

gjn:
jenkins test

gjn:
jenkins test

gjn:
Ok, I've rebase to latest master and updated the data. Does not currently work because of troubles importing the data to the sqlite db. Looks like primary key is dublicated for some data.

(for reference, here is the old master branch point from october: 3336ee902bc1e1b6be5ed9e736a7f3bb0c46f2b0)

gjn:
(the old branch of reference compared to current branch: https://github.com/geoadmin/mf-chsdi3/compare/3336ee902bc1e1b6be5ed9e736a7f3bb0c46f2b0...master)

ltclm:
impressive!

PR #1096, opened 2014-10-29
Add bulk tests to check presense of sphinxsearch indices

gjn:
This PR adds a test that checks the presense of sphinxsearch indices as they should be present as defined in the db. It was made in order to avoid having a situation where we have layers released that need indices to be present.

If the layer has the attribute `selectbyrectangle` set to true, a featureidentify request should be possible

If the layer has the attribute `searchable` set to true, a featuresearch request should be possible

Note:
While writing this PR, I noticed that we already have a mismatch between db definitions and physical presense of indices. More in comments.

### DIFF
--- a/chsdi/tests/integration/test_layers.py
+++ b/chsdi/tests/integration/test_layers.py
@@ -35,7 +35,7 @@ class LayersChecker(object):
         del self.testapp
         return False
 
-    def ilayers(self, queryable=None, hasLegend=None):
+    def ilayers(self, queryable=None, hasLegend=None, searchable=None):
         valNone = None
         query = self.session.query(distinct(LayersConfig.layerBodId)) \
             .filter(LayersConfig.staging == self.staging) \
@@ -44,6 +44,8 @@ class LayersChecker(object):
             query = query.filter(LayersConfig.queryable == queryable)
         if hasLegend is not None:
             query = query.filter(LayersConfig.hasLegend == hasLegend)
+        if searchable is not None:
+            query = query.filter(LayersConfig.searchable == searchable)
         for q in query:
             if self.onlylayer is None or q[0] == self.onlylayer:
                 yield q[0]
@@ -83,6 +85,12 @@ class LayersChecker(object):
         for lang in ('de', 'fr', 'it', 'rm', 'en'):
             assert ((layer + '_' + lang) in legendImages), layer + lang
 
+    def checkSearch(self, layer):
+        # If it fails here, it most probably means given layer does not have sphinx index available
+        link = '/rest/services/all/SearchServer?features=' + layer + '&bbox=600818.7808825106,197290.49919797093,601161.2808825106,197587.99919797093&type=featuresearch&searchText=dummy'
+        resp = self.testapp.get(link)
+        assert resp.status_int == 200, link
+
 
 def test_all_htmlpopups():
     with LayersChecker() as lc:
@@ -106,3 +114,9 @@ def test_all_legends_images():
     with LayersChecker() as lc:
         for layer in lc.ilayers(hasLegend=True):
             yield lc.checkLegendImage, layer, legendImages
+
+
+def test_all_searchable_layers():
+    with LayersChecker() as lc:
+        for layer in lc.ilayers(searchable=True):
+            yield lc.checkSearch, layer


AFoletti:
wow. I totally missed this one.
This needs a rebase and the def here [1] will no more be needed once the removal of the queryable indexes is merged [2]. Otherwise, it looks good.


[1] https://github.com/geoadmin/mf-chsdi3/pull/1096/files#diff-bc4339b7d81754dbbb2a3772541d7870R248
[2] https://github.com/geoadmin/service-sphinxsearch/tree/gjn_remove_queryablelayers

gjn:
I've re-activated this after the removal of all those select_by_rectangle indices.

When running locally, there's an error that `ch_bazl_luftfahrthindernis` layers is not searchable. It seems like it's searchable in bod but there's no sphinx index. @AFoletti @ltclm 
(it also doesn't work to search for features [on prod](http://map.geo.admin.ch/?X=90876.26&Y=663005.00&zoom=1&lang=fr&topic=aviation&bgLayer=ch.swisstopo.pixelkarte-grau&layers=ch.bazl.luftfahrthindernis&catalogNodes=1379,1381,1514))

gjn:
@AFoletti @ltclm Could you please look at the case mentioned in my previous comment?

ltclm:
This is exception has been done because of https://github.com/geoadmin/mf-geoadmin3/pull/2023

gjn:
In there, there is a reference for https://github.com/geoadmin/service-sphinxsearch/pull/187, but it seems like this index does not exist in sphinx. In fact, the source definition exists in the sphinx config (https://github.com/geoadmin/service-sphinxsearch/blob/master/conf/uvek.conf.part#L501), but there's no index definition.

Could this be done?

ltclm:
Do we still have to set a layer to "searchable" in the bod if we would like to use the db based query tool? If this is not needed anymore, we can remove it from sphinx and change the bod entry.

gjn:
Yes, the query tool filter still only displays searchable layers. (Reason given in the referenced Pull Request)

ltclm:
Ok, we can create the index, but we have to ask the datenherr first if this layer can be searchable in map.geo.admin.ch. @mariokeusen could you please ask them?

AFoletti:
@gjn or @loicgasser: does the querytool, which is activated for the luftfahrthindernis layer, need searchable=true? It's the only reason I see to have that boolean to TRUE in the BOD without the corresponding shpinx index.
If that's not the case and the querytool is unrelated to the BOD, I'll set the boolean to FALSE

ltclm:
> Yes, the query tool filter still only displays searchable layers. (Reason given in the referenced Pull Request) 

apparently yes, but what is this setting in the model definition for ``__queryable_attributes__ = ['ziegeleien']`` ?

AFoletti:
We are using a boolean (namely "searchable") for something it was not designed to. It now means "searchable" and "querytool=yes". That's in my opinion not so good

loicgasser:
yes, searchable layers means that they need queryable attrs in their models and are available in the object info.

AFoletti:
> yes, searchable layers means that they need queryable attrs in their models and are available in the object info.

I disagree with this definition.
Searchable layer: --> they need searchable=TRUE + sphinx index (nothing in the model)
Layers with querytool: --> they need __queryable_attributes__ in the model (OK) and searchable=TRUE in BOD (in my opinion, wrong, because as we have just seen, they are not really "searchable" as we intended them until now).

Then: either we make the querytool an extension of searchable (a layer with querytool MUST have an index) or we separate the two completely (a new boolean?)

ltclm:
we introduced the attributes searchable and queryable to document the different sphinx indexes in the bod. No we dont have queryable indexes anymore in sphinx, so we can get rid of this attribute, right?
I dont understand why the **query** tool only shows **searchable** layers?  The query tool should show all layers with **__queryable_attributes__** in their model.

AFoletti:
> The query tool should show all layers with queryable_attributes in their model.
exactly! And be independent from the searchable boolean.

> No we dont have queryable indexes anymore in sphinx, so we can get rid of this attribute, right?
I think we still need it for enabling the tooltip.

gjn:
Good discussion. I think @AFoletti is right.

a) searchable: features can be searched using the search tool of map.geo.admin.ch. Therefore, sphinx index must exist
b) queryable: layers have tooltips and can be found with select by rectangle action (part of the query tool)
c) having queryable attributes: query tool can be used to query specific attributes of layers

Problem: how can front-end know if c) is the case? This should be part of layersConfig. Your take @loicgasser ?

ltclm:
yes and we should probably update the following attributes in the bod: 
* *queryable* 
* *selectbyrectangle* 

Is it possible to put them together and replace them with an attribute *tooltips*?

gjn:
Maybe. In front-end code, we still differentiate between the two. We use 'queryable' for tooltip and 'selectbyrectangle' for bbox selection in query tool.

Are the all the same in the bod?

ltclm:
No, we have
```sql
$ psql -U www-data -h localhost -d bod_master -c "select queryable,selectbyrectangle,count(1) from re3.layers_js group by queryable,selectbyrectangle;"
 queryable | selectbyrectangle | count
-----------+-------------------+-------
 f         | f                 |   184
 t         | f                 |    18
 t         | t                 |   249
(3 Zeilen)
```
But we could rename queryable to tooltip...

ltclm:
@AFoletti and i propose to merge the *queryable* and *selectbyrectangle* attributes. The new attribute should have a more meaningful name like *tooltip [true|false]*.

The separation between *queryable* and *selectbyrectangle* has been done to save diskspace and index generation time on the sphinx infra - the arealstatistik layers have been the trigger for this change [1]. We think that it should not make a difference if you are accessing feature information by clicking a point or drawing a rectangle. 

[1] https://github.com/geoadmin/service-sphinxsearch/pull/126

gjn:
i fully agree.

ltclm:
I have added a new column ``tooltip`` to the following objects in the bod:
``bod_master.re3.layers_js``
``bod_master.re3.view_layers_js``
This column is triggered and updated automatcally by changes in the attributes ``queryable`` and ``selectbyrectangle``

Next steps would be:
* adapt code to use column tooltip instead of queryable and selectbyrectangle
* adapt code to use ``__queryable_attributes__`` from the model and provide this information in the layersconfig service

@gjn do you agree? Who will take care of these changes?

ltclm:
related changes in the bod
https://github.com/geoadmin/db/commit/c51726cd7b6acc38cb37d94026cdaf46a9e52d55#diff-c24334701204a238fe43fecd0b103903R10139

gjn:
@ltclm Does it make sense to say that each layer that is `searchable` should have __queryable_attributes__ in model definition. That way, we don't need step 2 that you proposed.

(We can add automated tests to assure that)

ltclm:
I think it would be better to keep 
``searchable --> sphinx index`` 
unlinked from 
``queryable_attributes --> query tool``

searchable attributes can / and will for sure be different from queryable attributes.

cedricmoullet:
Meeting to be organized

gjn:
@AFoletti @ltclm I've re-activated this PR. The points discussed have been adressed (see #1620) and searchable attribute will only be used for search in front end after today's deploy.

Anyhow, running [the tests shows](https://jenkins.ci.bgdi.ch/job/PR_chsdi3/708/console) that we have the searchable attribute to True in the db while not having any sphinx index for them. The following layers are affected:

```
ch_bazl_luftfahrthindernis
ch_bav_sachplan-infrastruktur-schifffahrt_ausgangslage
ch_bfe_sachplan-uebertragungsleitungen_anhoerung
ch_bav_sachplan-infrastruktur-schifffahrt_anhoerung
ch_bafu_neophyten-riesenbaerenklau
ch_bafu_neophyten-japanisches_geissblatt
ch_swisstopo_geologie-gravimetrischer_atlas_metadata
```

Maybe the had searchable attributed to true because the should be usable with the query tool. If so, this won't be necessary after today's deploy anmore.

Could you please check those layers?

AFoletti:
I'm having troubles with my connection today, but I'll check them all tomorrow

AFoletti:
Activity:
- ch_bafu_neophyten-riesenbaerenklau
- ch_bafu_neophyten-japanisches_geissblatt
- ch_swisstopo_geologie-gravimetrischer_atlas_metadata
were set to "searchable = false"
- ch_bfe_sachplan-uebertragungsleitungen_anhoerung
has a config (https://github.com/geoadmin/service-sphinxsearch/blob/master/conf/uvek.conf.part#L263) but the table is empty. Could it have generated the error?
- ch_bav_sachplan-infrastruktur-schifffahrt_ausgangslage
- ch_bav_sachplan-infrastruktur-schifffahrt_anhoerung
Indexes are added (PR: https://github.com/geoadmin/service-sphinxsearch/pull/235)

- ch_bazl_luftfahrthindernis
don't really know the status here. Maybe @ltclm ?

ltclm:
ch.bazl.luftfahrthindernis can be set to searchable=FALSE
It was an exception, see:
https://github.com/geoadmin/mf-geoadmin3/pull/2023

Since we dont use sphinx anymore for selectbyrectangle, we can get rid of this exception.

gjn:
jenkins test

gjn:
@AFoletti 

regarding `ch_bfe_sachplan-uebertragungsleitungen_anhoerung`, if there is not data, it should probably not be searchable. It's a workaround because sphinxsearch isn't capable of creating empty indices.

gjn:
jenkins test

AFoletti:
@gjn I don't see a big problem in setting empty layers temporarily to searchable=false (provided that we remember to set it again to true once needed :p). Will it solve your problem or should the config be removed too?

gjn:
searchable false should do it. Thanks alot for doing it and deploying to dev :)

AFoletti:
Done. It could be that another sachplan layer has only empty tables. If your test still fails, just tell me :)

gjn:
jenkins test

gjn:
Seems like this is going to be a never ending story.

`ch_swisstopo_geologie-geowege` is missing sphinx index...@AFoletti

gjn:
jenkins test

gjn:
ping @AFoletti 

Can you review state of `ch_swisstopo_geologie-geowege` (still failing)? Either searchable flag is wrong or index is missing.

AFoletti:
sry...
checked the status. Searchable flag war wrong. I set it to FALSE and deployed to dev

gjn:
jenkins test

gjn:
I will merge this and activate sphinx-index tests for searchable layers.

PR #1132, opened 2014-11-20
Filtering out geometries for unauthorized access

procrastinatio:
Carefully test, as I do'nt know exactly how the result, either a JSON or a GeoJSON is used in the makos. If it has to be a geometry, we should use a dummy one.

### DIFF
--- a/chsdi/views/features.py
+++ b/chsdi/views/features.py
@@ -20,6 +20,8 @@ from chsdi.models import models_from_name, oereb_models_from_bodid
 from chsdi.models.bod import OerebMetadata, get_bod_model
 from chsdi.views.layers import get_layer, get_layers_metadata_for_params
 
+PROTECTED_GEOMETRY_LAYERS = ['ch.bfs.gebaeude_wohnungs_register']
+
 
 # For several features
 class FeatureParams(MapServiceValidation):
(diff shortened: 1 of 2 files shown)


procrastinatio:
Rebased.

loicgasser:
I agree with @gjn See earlier comment:

@procrastinatio I think this is a tricky one. Feature requests and htmlpopu request have params.layerId and they work fine

But find and identify requests have params.layers (as it can contain multiple layers) and find requests have params.layer. So we would need to adapt this if statement to get it to work for those services.

gjn:
Here's a sample identify request that will fail with this PR applied:
`rest/services/ech/MapServer/identify?geometry=653199.9999999999,137409.99999999997&geometryFormat=geojson&geometryType=esriGeometryPoint&imageDisplay=1920,623,96&layers=all:ch.bfs.gebaeude_wohnungs_register&mapExtent=633200,132729.99999999997,671600,145189.99999999997&tolerance=5`

In fact, I suspect that all/most identify and find requests fail with this PR.

procrastinatio:
@loicgasser so we have a method with mutliple input parameters and output geometries, is it not a bit dangerous ?

gjn:
Yes, potentially. Could make sense to separate them.

procrastinatio:
OK, please test. The follwing three services are *protected*

    http://localhost:9001/rest/services/ech/MapServer/identify?geometry=653199.9999999999,137409.99999999997&geometryFormat=geojson&geometryType=esriGeometryPoint&imageDisplay=1920,623,96&layers=all:ch.bfs.gebaeude_wohnungs_register&mapExtent=633200,132729.99999999997,671600,145189.99999999997&tolerance=5

    http://localhost:9001/rest/services/ech/MapServer/find?layer=ch.bfs.gebaeude_wohnungs_register&searchText=berges&searchField=strname1

    http://localhost:9001/rest/services/ech/MapServer/ch.bfs.gebaeude_wohnungs_register/490830_0

loicgasser:
I propose one validation model per service and one test file per service.
@procrastinatio I am testing.

chsdi/views/features.py
```diff
@@ -473,9 +474,17 @@ def _process_feature(feature, params):
     # TODO find a way to use translate directly in the model
     if params.returnGeometry:
         f = feature.__geo_interface__
+        # Filter out this layer individually, disregarding of the global returnGeometry
+        # setting
+        if not params.varnish_authorized:
+            if hasattr(params, 'layers') and feature.__bodId__ == 'ch.bfs.gebaeude_wohnungs_register':
+                f = feature.__interface__
+            if hasattr(params, 'layer') and params.layer == 'ch.bfs.gebaeude_wohnungs_register':
```
loicgasser: minor `elif` would be slightly better.

loicgasser:
I would avoid wirting this layer name everywhere.
Instead I would do:
`PROTECTED_LAYERS=['ch.bfs.gebaeude_wohnungs_register']`
at the top or in the config and then
`bla in PROTECTED_LAYERS`

loicgasser:
Ok to me. FYI: I rebased your branch since we updated some libs today.

procrastinatio:
@loicgasser Your validate globally for a service, the geometry has to be removed for a specific layer, disregarding the value of the global *returnGeometry*

procrastinatio:
@loicgasser C'est rebasé

Issue #1006, opened 2014-09-11
When switching ol3 version


gjn:
When switching preview version to production version, the following needs to be adapted:

- Adapt the documentation (`View2D` is not used anymore and should be removed from the documentation)
- Adapt all examples to remove `View2D`
- Adapt `lubis_map.mako` (viewing of areal image is not working with latest version -> see also ga-tsm example in ol3). This mako is used in extended lubis tooltip and in the luftbild viewer as well.

PR #1168, opened 2014-12-17
The Preview is now the official API (no more view2D)

gjn:
This PR removes the preview of the loader and makes it the official version that will be served. The examples and documentation inside mf-chsdi3 are adapted accordingly (mainly replacing View2D with View). Note however, that old simple code using View2D should still work (we use alias), but might break in other areas.

I've kept the old versions (old-ga.js and old-ga-whitespace.js) around to be able to switch back to the old version quickly. (Year-end and all...)

Once this is deployed, we need to adapt the API documentation hosted on github (-> https://github.com/geoadmin/ol3/pull/112) and also our examples in codepen.$

Note that this also adresses #1006


gjn:
I've rebased.

@kandre Could you apply the css changes? Maybe best directly to https://github.com/geoadmin/ol3 and I'll update then this PR accordingly.

chsdi/templates/lubis_map.mako
```diff
@@ -36,7 +41,7 @@
                   }
 
                   curInstance = (++curInstance > MAX_INSTANCES) ? 0 : curInstance;
-                  return url.replace('{curInstance}', curInstance) + tileCoord.toString() + ".jpg";
+                  return url.replace('{curInstance}', curInstance) + tileCoordToString(tileCoord) + ".jpg";
```
oterral: You can use `tileCoord.join(',') ` because it's an array

oterral:
@gjn  I fix the attribution  css in the api --> https://github.com/geoadmin/ol3/pull/114/files

gjn:
@oterral I've adapted after merging your fix. But it does not look correct.

Please see [here](http://mf-chsdi3.dev.bgdi.ch/ltjeg/) or [here](http://mf-chsdi3.dev.bgdi.ch/ltjeg/examples/geoadmin_mapoverlay.html)

oterral:
ök i look into it

oterral:
Hope it's good now: https://github.com/geoadmin/ol3/pull/115/files

gjn:
Had to correct a little bit, but looks good now.

cedricmoullet:
@pauloamado Have you tested the usage of the API3 in the CMS with the preview mode ?

gjn:
@cedricmoullet As discussed yesterday, we want this for today's deploy, right?

PR #1188, opened 2015-01-06
Update aviation translations

loicgasser:
https://github.com/geoadmin/mf-geoadmin3/issues/2035


loicgasser:
RM is now in DE (as it should used to be EN)
EN translations are now up to date.
Trivial self merge

PR #1189, opened 2015-01-06
Print A3: go to full resolution

kandre:
150 dpis for both A4 and A3. The geoadmin part is not reverted as it can handle only one dpi value, see [#1899](https://github.com/geoadmin/mf-geoadmin3/pull/1899/files).

### DIFF
--- a/print/config.yaml
+++ b/print/config.yaml
@@ -1,5 +1,5 @@
 # allowed DPIs
-dpis: [150, 100]
+dpis: [150]
 
 #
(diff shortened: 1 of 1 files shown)


cedricmoullet:
@kandre @procrastinatio shouldn't we revert the https://github.com/geoadmin/mf-geoadmin3/pull/1899/files ?

oterral:
I suggest to kevin to remove it after the deploy.

PR #1200, opened 2015-01-09
replace oid with lknr

ltclm:
This pr will replace the oid in the model of the following layers:
* ch.bafu.sturm-staudruck_xxx
* ch.swisstopo.skitourenkarte-50.metadata


loicgasser:
LGTM

PR #1208, opened 2015-01-13
Include pkey in getAttributes when needed

loicgasser:
Would be nice to have it for tomorrow.

Some layers define pkeys as queryable attributes, this was not taken into account so far.

### DIFF
--- a/chsdi/models/vector/__init__.py
+++ b/chsdi/models/vector/__init__.py
@@ -168,22 +168,27 @@ class Vector(GeoInterface):
             return cls.__mapper__.columns.get(columnName)
         return None
 
-    def getOrmColumnsNames(self):
+    def getOrmColumnsNames(self, excludePkey=True):
         primaryKeyColumn = self.__mapper__.get_property_by_column(self.primary_key_column()).key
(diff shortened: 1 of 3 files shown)


procrastinatio:
OK, but maybe the name `excludePkey` is too restrictive.

loicgasser:
What do you suggest then?

loicgasser:
@gjn adapted to comments

gjn:
I still don't understand why you introduce a parameter that is always set to the same value (`False`) when the function is called...

loicgasser:
Per default we exclude the primary key from the attributes unless specified diffrently.

loicgasser:
You can read http://www.diveintopython.net/power_of_introspection/optional_arguments.html

gjn:
My fault. I didn't see https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/models/vector/__init__.py#L120

loicgasser:
@procrastinatio @gjn Good to you guys? Can I merge?

gjn:
Good for me.

Issue #1215, opened 2015-01-14
GetCap content and layer title


cedricmoullet:
http://wmts.geo.admin.ch/1.0.0/WMTSCapabilities.xml

-> ch.swisstopo.pixelkarte-farbe should not have >Landeskarte 1:25&amp;#39;000 | LK25 as title since it's not only a 1:25'000
-> content (title, description) should be checked in Arcgis. Is the content readable (special character, accent) ?

loicgasser:
In response header Content-Type:text/xml; charset=UTF-8

so yes should be able to treat it correctly

ltclm:
The content in the database is
``Landeskarte 1:25'000 | LK25 `` 
This string is somehow converted into:
``
Landeskarte 1:25&amp;amp;#39;000 | LK25 
``
@loicgasser Any idea, why is the apostrophe converted?

loicgasser:
@ltclm ok for title
But for ch.swisstopo.pixelkarte-farbe <Layer> is not present in test. Any idea?

ltclm:
@mariokeusen do we have a geocat entry for the base layers ch.swisstopo.pixelkarte-farbe|grau ?
If not we have to translate these layers manually in the bod to fix this issue in the wmts getcap.

loicgasser:
@ltclm I know why, it is not part of the inspire topic.

ltclm:
@loicgasser 
The base layer is not part of the topics ech and inspire. You can checkt the view bod.re.view_bod_wmts_getcapabilities_de.
But i wonder why this base layer is visible in the prod getcap then! What is the dev url of the getcap service?

loicgasser:
because of the all tag... Which make it not apply any filter... So I will fix in a separate PR.
Thanks, I wasn't sure about the base layers.

procrastinatio:
Side note, everywhere in title, abstract the apostrophe *'* (#39 or u+0027)  and the right single quotation mark  *’* (#8219 or u+2019) are used interchangeably, which is wrong.

cedricmoullet:
@loicgasser what are the next steps here ?

loicgasser:
@ltclm
In the view `view_bod_wmts_getcapabilities_themes_fr` base layers are visible in fk_dataset_id (list of layers separated with commas)
Should we always remove them from getCap?
For ech/inspire/all ?

loicgasser:
[...]

PR #1222, opened 2015-01-14
Remove default h filter from wmts getCap

loicgasser:
Fix for layer title:
https://github.com/geoadmin/mf-chsdi3/issues/1215#issuecomment-69910527

probably related to mako update

### DIFF
--- a/WMTSCapabilities.xml
+++ b/WMTSCapabilities.xml
@@ -1,0 +1,12267 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<Capabilities xmlns="http://www.opengis.net/wmts/1.0" xmlns:ows="http://www.opengis.net/ows/1.1" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:gml="http://www.opengis.net/gml" xsi:schemaLocation="http://www.opengis.net/wmts/1.0 http://schemas.opengis.net/wmts/1.0/wmtsGetCapabilities_response.xsd" version="1.0.0">
+    <!-- Revision: $Rev$ -->
+   <ows:ServiceIdentification>
+        <ows:Title>Federal Geodata Infrastructure of Switzerland</ows:Title>
+        <ows:Abstract>Some Geodata are subject to license and fees</ows:Abstract>
+        <ows:Keywords>
+            <ows:Keyword>FGDI</ows:Keyword>
+            <ows:Keyword>Pixelkarte</ows:Keyword>
+            <ows:Keyword>Switzerland</ows:Keyword>
+        </ows:Keywords>
+        <ows:ServiceType>OGC WMTS</ows:ServiceType>
+        <ows:ServiceTypeVersion>1.0.0</ows:ServiceTypeVersion>
+        <ows:Fees>yes</ows:Fees>
+        <ows:AccessConstraints>license</ows:AccessConstraints>
+</ows:ServiceIdentification>
+<ows:ServiceProvider>
+        <ows:ProviderName>swisstopo</ows:ProviderName>
+        <ows:ProviderSite xlink:href="http://www.swisstopo.admin.ch"/>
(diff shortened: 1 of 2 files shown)


gjn:
Officially [1][2] apostrophe character (`'`) needs to be escaped in xml. But instead of what @ltclm sees (`&amp;amp;#39;`) it should be converted to `&apos;`. I'm not sure if this PR results in a valid XML as per the specification.

IMO, we should rather find out why the x filter does not put the correct replacement in.

[1] http://stackoverflow.com/questions/1091945/what-characters-do-i-need-to-escape-in-xml-documents#1091953
[2] http://www.w3.org/TR/xml/#syntax

I'm on record saying that xml sucks...here's one reason.

loicgasser:
Mmm... it might be &apos; already. you browser reads it as XML so you wouldn't see apos anyway. right?

loicgasser:
check this one. http://www.w3schools.com/xml/note.xml you don't see apos anywhere..

loicgasser:
http://docs.makotemplates.org/en/latest/filtering.html
none of those filters make sense. Could you enlight me?

gjn:
It's `&#39;` in the raw response, which is apostrophe escaped for html, not for xml. I'm confused.

loicgasser:
maybe pyramid take care of it when you define the content-type now (update 2 weeks ago). In the internal machinery, I'll try to find a refrence.

loicgasser:
ok just saw your comment. previous one doesn't make sense then.

gjn:
yes, might be pyramids

cedricmoullet:
If we apply this PR, the XML is not valid anymore. @loicgasser is it not possible to generate a valid XML ?

loicgasser:
Please have a look at mako's source code `'` are replaced with `&#39;`

so it should be valid. Here we simply avoid to escape multiple times.

loicgasser:
See this link: https://bitbucket.org/zzzeek/mako/src/b6d0479034e4f379576caa6eb9ef3e32369c333f/mako/filters.py?at=master

Please merge if you are convinced otherwise, I'll make a utility function to transform `&#39;` into `&apos;`

procrastinatio:
We could also use `<![CDATA[  ...   ]]>`to escape all contents

procrastinatio:
Otherwise, all tests pass. So it is OK.

loicgasser:
`<![CDATA[ ... ]]>` is mainly use for script no?
+ `Everything inside a CDATA section is ignored by the parser.`
Which is not really what we want.
I think this can merged.

loicgasser:
@gjn Is it ok if I merge?

gjn:
Sorry for being late. Are we sure we don't have `&`, `<`, `>` in the source? Because those need to be escaped by the filter Not sure if we can be sure, especially since data comes from geocat, which might change without us knowing.

Problem is that mako assumes we want to create html, even though doc says `x` filter is for xml.

On the other hand, it seems like original problem is that it was escaped twice, so maybe pyramids escaped it as well and this is now ok?

Can we try to read this data coming from this branch into ArcGis?

loicgasser:
Found the line
default_filters = sget('default_filters', 'h')

So per default all mako template are escaped with h.
This is for security reasons apparently.
So I need to disable the default mako filter for xml and use x instead...

loicgasser:
Also n disable all default filters. Might just need to add this filter first every time we need to escape xml. Page filters are ignored for some obscure reasons..

loicgasser:
Ok PR updated. (Train rides can be useful)
It absolutely doesn't change a thing, but at least we fully understand what's going on.

It is really missing in the doc. As in developement.ini.in I did not specify anything for `default_filters` I did not excpect that.
Reference:
https://github.com/Pylons/pyramid_mako/blob/7e1c63e76c64c8599bd1d0563fce8665a91647ec/pyramid_mako/__init__.py#L189


Please merge.

gjn:
Ha! This explains the double escaping. Can't test now, but according to your refs, this should be good now and assures it's well xml escaped. Good job.

PR #1225, opened 2015-01-15
Remove duplicate column definition

gjn:
Colum('topics') is used twice in the GetCap model, resulting in warnings during the build. This removes one definition to get rid of the warning.

Looks like `topics` of this model is never used, but would be glad if someone could verify also.


loicgasser:
Good catch! No it is not used.

Issue #1220, opened 2015-01-14
Searchable attributes of a layer with several models 


pfanguin:
When a layer uses several models and different searchable attributes in the models are used, only the searchable attributes of the first model appear in the list of attribute (query tool).

Example : ch.bav.sachplan-infrastruktur-schiene_kraft

loicgasser:
solved via pull request #1226

PR #1226, opened 2015-01-15
Support multi models in attributes listing and preview services

loicgasser:
This provides a fix for https://github.com/geoadmin/mf-chsdi3/issues/1220

We now support mulit models layers with different attributes

Now in attributes listing values are sorted.

*Attribute Listing*
Prod at the moment:
http://api3.geo.admin.ch/rest/services/api/MapServer/ch.bav.sachplan-infrastruktur-schiene_kraft

not all searchable fields are displayed (or queryable attributes)

*VS*

Deployed branch
http://mf-chsdi3.dev.bgdi.ch/gal_attrs_mulit_models/rest/services/api/MapServer/ch.bav.sachplan-infrastruktur-schiene_kraft

which actually displays all the available searchable fields.

*Attribute values preview (use plname_de)*
Prod at the moment:
http://api3.geo.admin.ch/rest/services/api/MapServer/ch.bav.sachplan-infrastruktur-schiene_kraft/attributes/plname_de

ends up in a bad request (this field does not exist while it exists on the models)

*VS*

Deployed branch
http://mf-chsdi3.dev.bgdi.ch/gal_attrs_mulit_models/rest/services/api/MapServer/ch.bav.sachplan-infrastruktur-schiene_kraft/attributes/plname_de

Returns a set of preview values for plname_de

Additionnaly I updated six which is compatibility module to smooth differences between Python 2 and 3.

### DIFF
--- a/chsdi/tests/integration/test_mapservice.py
+++ b/chsdi/tests/integration/test_mapservice.py
@@ -409,12 +409,21 @@ class TestMapServiceView(TestsBase):
     def test_layersconfig_wrong_map(self):
         resp = self.testapp.get('/rest/services/foo/MapServer/layersConfig', status=400)
 
-    def test_features_attributes(self):
+    def test_layer_attributes(self):
         resp = self.testapp.get('/rest/services/ech/MapServer/ch.bafu.bundesinventare-bln', status=200)
         self.failUnless(resp.content_type == 'application/json')
 
-    def test_features_attributes_wrong_layer(self):
+    def test_layer_attributes_wrong_layer(self):
         resp = self.testapp.get('/rest/services/ech/MapServer/dummy', status=400)
+
+    def test_layer_attributes_multi_models(self):
+        resp = self.testapp.get('/rest/services/api/MapServer/ch.bav.sachplan-infrastruktur-schiene_kraft', status=200)
+        self.failUnless(resp.content_type == 'application/json')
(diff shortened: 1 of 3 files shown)


loicgasser:
I added a commit to take care of 1) of this issue: https://github.com/geoadmin/mf-geoadmin3/issues/2012

gjn:
Could you squash the commits.

loicgasser:
yep, squashed and rebased

loicgasser:
@procrastinatio @gjn Please review. As all labels are translated. We should try to fix all thoses bugs. I would like to have it for next week as well.

gjn:
@procrastinatio Could you give it a go?

loicgasser:
@procrastinatio ?

procrastinatio:
Well, I do not have major objection to this, but:
- we merge all attributes from different models. I think, for now, the diffent models are used at different scale, so only the model _valid_ at a specific scale should be used and its features returned. Why don't we used an ggregate layer in this case ? Or just simply swap the geomtry column ?
- no test.

loicgasser:
I think, for now, the different models are used at different scale
Not true, there several reasons for that.

Why don't we used an aggregate layer in this case ?
Explain me how you would do that.

No test
Will come please merge. And open issues if you feel like it.

loicgasser:
Rebased and tests have been added. An issue has been added with @procrastinatio proposal.

PR #1229, opened 2015-01-20
searchable attributes 20150116

pfanguin:


chsdi/locale/empty_chsdi.po
```diff
@@ -1,3 +1,127 @@
...
+
+
+
+
+
+
+
+
+
+
```
loicgasser: some extra space here. ;)

PR #1242, opened 2015-01-22
WMTS: add new Tilematrixset 17,19,24

ltclm:
this pr is adding new tilematrixset:  17 19 24

### DIFF
--- a/chsdi/templates/TileMatrixSet.mako
+++ b/chsdi/templates/TileMatrixSet.mako
@@ -179,13 +179,3 @@
     <MatrixWidth>19</MatrixWidth>
     <MatrixHeight>13</MatrixHeight>
 </TileMatrix>
-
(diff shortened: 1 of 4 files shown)


procrastinatio:
OK, for the WMTS GetCapabilities part, it looks good. I am testing with various clients. We probably also need to reconsider the tilematrix for these layers in the reprojected WMTS tiles.

procrastinatio:
@ltclm I a going to extent the test for WMTS GetCapabilities to document excatly what we want and to make sure we do not break anything:
https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/tests/integration/test_wmtscapabilities.py

cedricmoullet:
@ltclm is it still a WIP ?

procrastinatio:
OK, you may merge.

PR #1240, opened 2015-01-22
Fix height test because of changed DTM

gjn:
Last build in Jenkins failed because of height test [1].

It seems like the COMB DTM Model changed on our infrastructure, but the difference in resulting height for Bern is 3.6 meters, which seems alot to me. @ltclm Do you know the reason for this big of a change?

(Note that the other models didn't change and before COMB and DTM2 were the same)

https://jenkins.dev.bgdi.ch/job/chsdi3/253/console

PR #1244, opened 2015-01-23
Fix height tests to be not on strange border case

gjn:
This replaces https://github.com/geoadmin/mf-chsdi3/pull/1240, which uses a bordercase that is not consistent. We don't know the reasons.


loicgasser:
Ok I saw your discussion, shouldn't we open an issue with a low prio to check that. I agree we should simply change the tests for now.

gjn:
We investigated, and can't explain. See http://api3.geo.admin.ch/rest/services/height?easting=600000&northing=200000&layers=COMB

Go there 10 times and you won't always get the same result.

gjn:
Merged because of failing jenkins.

PR #1258, opened 2015-02-06
Add ability to cancel multi-page prints

gjn:
This is to support https://github.com/geoadmin/mf-geoadmin3/pull/2129


procrastinatio:
If you cancel, you'll get an error 'Cannot merge' because of missing PDFs in the error log.

But otherwise, very good.

gjn:
Strange. I only start merging when cancel file does not exist yet...

PR #1263, opened 2015-02-09
[WIP] Get the total number of features of an identify request

oterral:
For https://github.com/geoadmin/mf-geoadmin3/pull/2143

Add a maxFeatures paramater to the identify request and add the number of features available for this request in the response.

Still limited to 10000 features.

### DIFF
--- a/chsdi/views/features.py
+++ b/chsdi/views/features.py
@@ -263,6 +263,8 @@ def _identify(request):
     if maxFeatures is None:
         maxFeatures = 200
 
+    total = _get_nb_features_for_filters(params, models, maxFeatures=maxFeatures, where=params.where)
+
     features = []
     for feature in _get_features_for_filters(params, models, maxFeatures=maxFeatures, where=params.where):
         f = _process_feature(feature, params)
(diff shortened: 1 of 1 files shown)


gjn:
How does this perform on big layers (e.g wohnungsregister?)

oterral:
Not good :)  I have tested with luftbilder

oterral:
a "Show the next 200 results" button should be better

gjn:
Hmm. @mariokeusen do we need to know a) how many features we really have, or is it enough to b) know that there are more feaures?

If b) is enough, we could make a maxlimit query of 201, return 200 results and indicated if 201 were found. Then next query with offset, etc. This might perform considerably better.

cedricmoullet:
I also think that Gilbert's proposal makes sense

On Tue, Feb 10, 2015 at 8:37 AM, Gilbert Jeiziner <notifications@github.com>
wrote:

> Hmm. @mariokeusen <https://github.com/mariokeusen> do we need to know a)
> how many features we really have, or is it enough to b) know that there are
> more feaures?
>
> If b) is enough, we could make a maxlimit query of 201, return 200 results
> and indicated if 201 were found. Then next query with offset, etc. This
> might perform considerably better.
>
> —
> <https://github.com/geoadmin/mf-chsdi3/pull/1263#issuecomment-73656475>.
>



-- 

oterral:
Out of date with #1266

PR #1269, opened 2015-02-11
Fixing WMTS tests

procrastinatio:
Please note:
EPSG:4852 is used in many GIS software as 'ETRS89' but is unknow to the reference database http://www.epsg.org/, while EPSG:4258 it is the opposite.


oterral:
thx for fixing this mess

Issue #1074, opened 2014-10-21
Luftbild does not contain width/height


gjn:
Probably a data error. Luftbild under [1] does not contain widht/height, resulting in a not working luftbild viewer [2]. [2] is the link in the html popup. Note the 'none' values...

[1] http://map.geo.admin.ch/?ch.swisstopo.lubis-luftbilder_schwarzweiss=19790470160318&lang=de&topic=luftbilder&X=190000.00&Y=660000.00&zoom=1&bgLayer=ch.swisstopo.pixelkarte-grau&layers=ch.swisstopo.lubis-luftbilder_schwarzweiss&catalogNodes=1179,1180,1186
[2] http://api3.geo.admin.ch/luftbilder/viewer.html?lang=de&layer=ch.swisstopo.lubis-luftbilder_schwarzweiss&title=Bildnummer&height=None&width=None&bildnummer=19790470160318&datenherr=swisstopo

ltclm:
there is a width and height value in the database for this image.
```sql
select ebkey,image_width,image_height FROM luftbilder_swisstopo_bw where ebkey = '19790470160318';
     ebkey      | image_width | image_height
----------------+-------------+--------------
 19790470160318 |       16858 |        16857
```
If you are opening this fullresviewer the "normal way" the image is displayed correctly:
![image](https://cloud.githubusercontent.com/assets/5286659/5646679/225d7e9c-967f-11e4-9d2a-366813641725.png)

In the url above in the ticket the parameters ``width`` and ``height`` are missing, the url should look like this:
http://api3.geo.admin.ch/luftbilder/viewer.html?lang=de&width=16858&layer=ch.swisstopo.lubis-luftbilder_schwarzweiss&bildnummer=19790470160318&title=Bildnummer&rotation=0&datenherr=swisstopo&height=16857

cedricmoullet:
@gjn could you please check that it's still an issue ?

ltclm:
here some examples of recently generated lubis tiles on s3 by igis:
```bash
07.02.2015 20:04:46 20047020035329 Image folder created
07.02.2015 20:04:46 20047020035321 Image folder created
07.02.2015 20:04:46 19880610170171 Image folder created
07.02.2015 20:04:46 19950010043801 Image folder created
07.02.2015 20:04:46 19880610087146 Image folder created
07.02.2015 20:04:46 19950010033774 Image folder created
07.02.2015 20:04:46 20047020025333 Image folder created
[...]

PR #1271, opened 2015-02-12
Get Lubis widht/height from metafile if it's not in db

gjn:
https://github.com/geoadmin/mf-chsdi3/issues/1074

As described in the comments of the issues, we can have a situation where the lubis image is ready (tiled), but the corresponding information (widht and height) is not yet in db. This happens because Lubis image tiles are created constantly, while the db is updated every x month. This resulted in #1074

With this fix, when the widht and height is not yet in the db, it will be taken from the metafile that exists for each lubis image.

### DIFF
--- a/chsdi/templates/htmlpopup/lubis.mako
+++ b/chsdi/templates/htmlpopup/lubis.mako
@@ -12,9 +12,10 @@ import markupsafe
 def br(text):
     return text.replace('\n', markupsafe.Markup('<br />'))
 
+tileUrlBasePath = 'http://aerialimages0.geo.admin.ch/tiles'
+
 
 def determinePreviewUrl(ebkey):
-    tileUrlBasePath = 'http://aerialimages0.geo.admin.ch/tiles'
 
     def getPreviewImageUrl(ebkey):
         return tileUrlBasePath + '/' + ebkey + '/quickview.jpg'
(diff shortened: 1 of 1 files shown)


chsdi/templates/htmlpopup/lubis.mako
```diff
@@ -233,7 +271,7 @@ viewer_url = get_viewer_url(request, params)
   </div>
   <br>
 
-% if preview_url != "" and c['attributes']['image_width']:
+% if preview_url != "" and image_width != None:
```
loicgasser: can `image_width` still be equal to None at this point?

chsdi/templates/htmlpopup/lubis.mako
```diff
@@ -270,7 +308,7 @@ viewer_url = get_viewer_url(request, params)
       map.highlightFeature('${c['layerBodId']}', '${c['featureId']}');
       map.recenterFeature('${c['layerBodId']}', '${c['featureId']}');
 
-% if preview_url != "" and c['attributes']['image_width']:
+% if preview_url != "" and image_width != None:
```
loicgasser: same question here.

loicgasser:
looking good, 2 minor comments. As a side note we should eventually move all the functions in a `.py` file and avoid too much code in mako. (lubis_helpers.py? maybe)

chsdi/templates/htmlpopup/lubis.mako
gjn: yes, when the metafile does not exist (meaning lubis tile for this image not yet created, or in process of being created) or some exception thrown in the imagesize_from_metafile function.

loicgasser: ok thanks

gjn:
Thanks for the review!

PR #1272, opened 2015-02-12
More robust catalog view

procrastinatio:
This is a quick fix not to break adding a layer to an non existing category.

Additionnaly, the default behaviour of the catalog is changed as follow:
In the `view_catalog` of the `BOD`, if a category/branch is set to `test` it's leaves or sub-categories won't show, even if they are set to `prod` on the production environnment.

The whole catalog integration process should be made fool-proof, as now it is really hand-crafted in the Google Doc.


loicgasser:
Tested an works as excpected, nosetests ok.
Please merge.

gjn:
Thanks alot!

Issue #1219, opened 2015-01-14
Lang specific searchable attribute


loicgasser:
For instance:

https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/models/vector/uvek.py#L1364

cedricmoullet:
@loicgasser what are the next steps here ?

procrastinatio:
It would be nice to add a bit of context to the issue, otherwise, when rewewing, you have to figure out what is the problem, why it is a problem, additionnally to how this is solved

Searching for `galleria`in `description_it`: [7 items](http://api3.geo.admin.ch/rest/services/ech/MapServer/find?layer=ch.bav.sachplan-infrastruktur-schiene_ausgangslage&searchText=galleria&searchField=description_it&returnGeometry=false)

Searching for `tunnel`in `description_it`: [no items](http://api3.geo.admin.ch/rest/services/ech/MapServer/find?layer=ch.bav.sachplan-infrastruktur-schiene_ausgangslage&searchText=galleria&searchField=description_it&returnGeometry=false)

If I undesertand correctly, the goal is to use a neutral fiedl `description` (instead of `description_it`, `descriptions_de`,etc.) and to use the `lang`parameter ?

loicgasser:
The goal is to retrieve only the labels that correspond to the correct lang. With a fallback on `fr` for it, and `de` for `en` and `rm` if the corresponding label does not exist.

Open PR here https://github.com/geoadmin/mf-chsdi3/pull/1277

It should not affect the behavior of the find service.

It is only for the query tool. (featureAttributes -> to retrieve the correct list of attributes and avoid duplicated translated lables)

See https://github.com/geoadmin/mf-chsdi3/blob/gal_lang_spec_attr/chsdi/views/layers.py#L126

PR #1277, opened 2015-02-12
Support lang specific attrs

loicgasser:
This is a fix for:
https://github.com/geoadmin/mf-chsdi3/issues/1219

### DIFF
--- a/chsdi/models/vector/__init__.py
+++ b/chsdi/models/vector/__init__.py
@@ -1,8 +1,10 @@
 # -*- coding: utf-8 -*-
 
 from sys import maxsize
+import re
 import datetime
 import decimal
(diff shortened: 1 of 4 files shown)


loicgasser:
@procrastinatio could you have a look. Comments are in the code so it should be pretty easy to understand.

loicgasser:
With this approach we detect if an attribute name define in `__queryable_attributes__` ends with `_(de|fr|rm|en|it)`

If it is the case, based on the lang parameter of the request, we filter out attributes that don't belong to the request language (with fallbacks for `it`, `en` and `rm`)

procrastinatio:
Otherwise OK for me.

loicgasser:
thanks  @procrastinatio for the review

PR #1282, opened 2015-02-16
Data arround value in tt

alcapat:
wait data owner agreement


procrastinatio:
I though we should use <bodid><stringid> for translation ?

alcapat:
we can go with label in 2 lines (text is too long) in french without change delivered translations.

procrastinatio:
OK, tested.

Issue #1224, opened 2015-01-14
Error 500 on identify


cedricmoullet:
http://api3.geo.admin.ch/rest/services/ech/MapServer/identify?geometry=692124.830021982,220998.9882507511&geometryFormat=geojson&geometryType=esriGeometryPoint&imageDisplay=1600,1080,96&layers=all:ch.vbs.territorialregionen&mapExtent=681014.830021982,215908.9882507511,697014.830021982,226708.9882507511&tolerance=5

geometry issue ?

cedricmoullet:
One more: http://api3.geo.admin.ch/rest/services/ech/MapServer/identify?geometry=704206.7132676127,133316.31032245018&geometryFormat=geojson&geometryType=esriGeometryPoint&imageDisplay=1280,939,96&layers=all:ch.are.alpenkonvention&mapExtent=286906.7132676126,-19433.689677549875,1118906.7132676127,590916.3103224501&tolerance=5

gjn:
Seems like any identify for these layers is not working:
`ch.are.alpenkonvention`
`ch.vbs.territorialregionen`

[Link](http://map.geo.admin.ch/?X=159095.66&Y=834230.24&zoom=1&lang=en&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.vbs.territorialregionen,ch.are.alpenkonvention&layers_opacity=0.6,0.75)

loicgasser:
I am on it.

loicgasser:
It crashed in shapely. Probably a geometry issue. @ltclm Can you check that the first and the last coordinates of the Polygons of these layers end with the same coordinate and if no, change it?

loicgasser:
or 3D geoms?

ltclm:
**ch.are.alpenkonvention -> 3d, valid**
```sql
psql -U www-data -h pg-0.dev.bgdi.ch -d are -c "select distinct st_ndims(the_geom),st_isvalidreason(the_geom) FROM siedlung_landschaft.alpenkonvention;"
 st_ndims | st_isvalidreason
----------+------------------
        3 | Valid Geometry
``` 

**ch.vbs.territorialregionen -> 3d, valid**
```sql
psql -U www-data -h pg-0.dev.bgdi.ch -d vbs -c "select distinct st_ndims(the_geom),st_isvalidreason(the_geom) FROM territorialregionen;"
 st_ndims | st_isvalidreason
----------+------------------
[...]

Issue #1150, opened 2014-12-04
Vector.__read__ do not support 3D geometries


procrastinatio:
See https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/models/vector/__init__.py#L55

loicgasser:
```python
>>> from shapely.geometry import LineString
>>> from shapely.wkt import loads
>>> line_txt = '(632168.882 180514.191 3250.65,618345.024 180508.595 3247)'
>>> line_wkt = loads(line_txt)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ltgal/mf-chsdi3/buildout/eggs/Shapely-1.3.3-py2.7-linux-x86_64.egg/shapely/wkt.py", line 10, in loads
    return geos.WKTReader(geos.lgeos).read(data)
  File "/home/ltgal/mf-chsdi3/buildout/eggs/Shapely-1.3.3-py2.7-linux-x86_64.egg/shapely/geos.py", line 221, in read
    raise ReadingError("Could not create geometry because of errors "
shapely.geos.ReadingError: Could not create geometry because of errors while reading input.
>>> line_txt = 'LINESTRING (632168.882 180514.191 3250.65,618345.024 180508.595 3247)'                                                
>>> line_wkt = loads(line_txt)
>>> line_wkt.wkt
'LINESTRING Z (632168.882 180514.191 3250.65, 618345.024 180508.595 3247)'

>>> from shapely.wkb import dumps
>>> dumps(line_wkt)                                                                                                                   
'\x01\x02\x00\x00\x80\x02\x00\x00\x00\x06\x81\x95\xc3\xd1J#A\x0c\x02+\x87\x11\t\x06A\xcd\xcc\xcc\xccLe\xa9@^\xbaI\x0c\xd2\xde"A)\\\x8f\xc2\xe4\x08\x06A\x00\x00\x00\x00\x00^\xa9@'
>>> loads('(632168.882 180514.191 3250.65,618345.024 180508.595 3247)')                                                               
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ltgal/mf-chsdi3/buildout/eggs/Shapely-1.3.3-py2.7-linux-x86_64.egg/shapely/wkt.py", line 10, in loads
    return geos.WKTReader(geos.lgeos).read(data)
[...]

PR #1286, opened 2015-02-17
Add custom geoalchemy2 egg for 3D support

loicgasser:
This is to solve 
https://github.com/geoadmin/mf-chsdi3/issues/1224
and
https://github.com/geoadmin/mf-chsdi3/issues/1150

@ltclm Let me know what table should be used to use Lubis original 3D data.

FYI: vbs and are layers are really dumm 3D layers as the third dimension is always equal to 0... (??)
Hope it makes a little more sense for lubis data.

No issue in geoadmin, which is a good thing.


ltclm:
Hi Loic
There is no 3d data anymore in the lubis database on db master (pg-0.dev.bgdi.ch). 

I have changed the FME workbench and just re-imported the 3d data directly in the pgcluster.dev.bgdi.ch database (the master is still in 2d mode). 
The mf dev staging should have access to the 3d data now, you can use the default tables.

ltclm:
```sql
select ebkey,st_astext(the_geom),rotation FROM luftbilder_swisstopo_bw limit 5;
     ebkey      |          st_astext           | rotation
----------------+------------------------------+----------
 19542890031226 | POINT Z (745338 289100 4100) |        0
 19722800013381 | POINT Z (703248 280180 3500) |        0
 19609991096473 | POINT Z (574728 96866 4620)  |        0
 19662560045775 | POINT Z (689600 256000 3800) |        0
 19892360012393 | POINT Z (736826 236000 5200) |        0
(5 rows)

```

procrastinatio:
As the solution is buried down in a tar.gz file, this how it was solved (thanks to @fredj)

```python
#######################################################
from sqlalchemy import create_engine, Table, MetaData from sqlalchemy.orm import sessionmaker

from geoalchemy2.shape import to_shape

engine = create_engine('postgresql://www-data:www-data@pgcluster.dev.bgdi.ch/vbs', echo=True) table = Table('territorialregionen', MetaData(), autoload=True, autoload_with=engine)

session = sessionmaker(bind=engine)()

item = session.query(table).first()

shape = to_shape(item.the_geom)
#######################################################

And I've the error but if I change:
diff --git a/geoalchemy2/types.py b/geoalchemy2/types.py index 54563fe..24561d6 100644
--- a/geoalchemy2/types.py
+++ b/geoalchemy2/types.py
@@ -100,7 +100,7 @@ class _GISType(UserDefinedType):
         return '%s(%s,%d)' % (self.name, self.geometry_type, self.srid)
 
     def column_expression(self, col):
-        return func.ST_AsBinary(col, type_=self)
+        return func.ST_AsEWKB(col, type_=self)
 
     def result_processor(self, dialect, coltype):
         def process(value):

```



See also https://github.com/geoalchemy/geoalchemy2/pull/92

procrastinatio:
OK

[ch.are.alpenkonvention](http://mf-chsdi3.dev.bgdi.ch/gal_3d_support/rest/services/ech/MapServer/identify?geometry=704206.7132676127,133316.31032245018&geometryFormat=geojson&geometryType=esriGeometryPoint&imageDisplay=1280,939,96&layers=all:ch.are.alpenkonvention&mapExtent=286906.7132676126,-19433.689677549875,1118906.7132676127,590916.3103224501&tolerance=5)

[ch.vbs.territorialregionen](http://mf-chsdi3.dev.bgdi.ch/gal_3d_support/rest/services/ech/MapServer/identify?geometry=692124.830021982,220998.9882507511&geometryFormat=geojson&geometryType=esriGeometryPoint&imageDisplay=1600,1080,96&layers=all:ch.vbs.territorialregionen&mapExtent=681014.830021982,215908.9882507511,697014.830021982,226708.9882507511&tolerance=5)

ltclm:
Hi Loic
Im not sure if these lubis db changes are breaking some tests on dev.
Anyway i have created a database **lubis_3d**. Please use this database for lubis tests.
I will now replace the lubis db with the original one.

Marcel

procrastinatio:
@loicgasser we should event. add some systematic test for all layers and services, on one random features each, to see if something breaks.

loicgasser:
Agree

procrastinatio:
OK, LGTM

loicgasser:
@ltclm We'll take care of Lubis in a separate PR.
Same for tests.

gjn:
We should try to re-use the official geoalchemy release. The referrenced PR in geoAlchemy is still open (https://github.com/geoalchemy/geoalchemy2/blob/master/geoalchemy2/types.py), but I think we can still use current geoalchemy master and simply overrate the class with our needed functionality. 

Then there would be no need to use a custom egg.

I'm working on geoJson loading from DB where I will try the overwrite approach. If I'm successfull, I can try to do the same to revert/replace this PR.

PR #1288, opened 2015-02-18
Feedback now contains unique identifier

gjn:
As requested by @p1d1d1 and topo, the feedback message now contains a unique identifier for each message. It's included both in the text message of the mail as well as in the meta.json.

### DIFF
--- a/chsdi/views/feedback.py
+++ b/chsdi/views/feedback.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 
+import uuid
 import cgi
 import json
 import datetime
@@ -72,11 +73,13 @@ def feedback(self, request):
         mailServer.sendmail(to, to, msg.as_string())
         mailServer.close()
 
+    # unique messag id
+    uid = str(uuid.uuid1())
     ua = getParam('ua', 'no user-agent found')
     permalink = getParam('permalink', 'No permalink provided')
     feedback = getParam('feedback', 'No feedback provided')
     email = getParam('email', 'Anonymous')
-    text = u'%s just sent a feedback:\n %s. \nPermalink: %s. \n\nUser-Agent: %s'
+    text = u'%s just sent a feedback:\n\n%s\n\nPermalink: %s\n\nUser-Agent: %s\n\nMessagID: %s'
     attachement = getParam('attachement', None)
     kml = getParam('kml', None)
     now = datetime.datetime.now()
@@ -92,14 +95,15 @@ def feedback(self, request):
         'kml': kmlfilename if (kml is not None and kml is not '') else '',
         'attachement': attachfilename,
         'userAgent': ua,
-        'date': now.strftime("%Y-%m-%d %H:%M")
+        'date': now.strftime("%Y-%m-%d %H:%M"),
+        'messagId': uid
     }
 
     try:
         mail(
             defaultRecipient,
             defaultSubject,
-            text % (email, feedback, permalink, ua),
+            text % (email, feedback, permalink, ua, uid),
             attachement,
             kml,
             kmlfilename,


davidoesch:
Here I would have suggested that the kml name with jjjmmddhhss and the corresponding email is already a unique identifier. Each email has a unique id - it is in even human readable in the mime format. Furthermore it gets a uuid in the revision db of swisstopo 

Don't overcomplicate things. For the sake of simplicity and sustainably I strongly suggest that uuid are generated and used where it is needed - no "fremdschlüssel" generated by a mf-geoadmin. It is a question sustainability in operating on geoadmin. (next wish : change id migrate id) - no business logic in the application..
I vote to solve this on the topo system and revert this mechanism.

Further more : such decision discuss this in the future before implementing it 

@cedricmoillet and @gjn? Your votes?

gjn:
I've got not problem deleting this PR and to solve the issue in the 'process landscape'.

p1d1d1:
Yes the message gets an UUID in the IGIS2TOPO process: but this is not the issue.
The issue is to have something that can identify a message in the Feedback2IGIS process.
The FME process by IGIS can sometimes fail, mainly when there is no meta.json file attached. THis can happen if a Helpdesker use "ReplyTo" instead of "Forward".
So, time to times, IGIS sent emails saying: "Please can you re-forward the following messages" (see example at the end of this comment). Finding the message in the helpdesk inbox is not trivial.

KMLName cannot be used as UUID since KML it's not always there. Meta.json is always there, so couldn't we add to the json file name date+time?


By the way I could also tell IGIS that we prefer to wait a couple of weeks and see if failures continue to happen. If it's the case, then we'll try to find a solution.

Let me know your decision, so that I can inform IGIS.

***************************************************
Lieber Helpdesk

Bei der Kontrolle der weitergeleiteten Revisionsmeldungen gab es Ungereimtheiten bei den untenstehenden 2 Meldungen. Darf ich euch bitten diese nochmals an fmerevisionservice@lt.admin.ch zu senden.

Danke & Gruss,
Alain

Job#1196
{
   "fns_type": "email_publisher",
   "email_publisher_to": "revisionservice@st02517a.lt.admin.ch",
   "email_publisher_content_type{0}": "text/plain; charset=\"iso-8859-1\"",
   "email_publisher_received": "Thu Feb 05 13:00:28 CET 2015",
   "email_publisher_sent": "Thu Feb 05 13:00:28 CET 2015",
   "email_publisher_content{0}": "\r\n\r\n-----Ursprüngliche Nachricht-----\r\nVon: webgis@swisstopo.ch [mailto:webgis@swisstopo.ch] \r\nGesendet: Donnerstag, 5. Februar 2015 12:32\r\nAn: _swisstopo-webgis\r\nBetreff: Customer feedback\r\n\r\nanna.leonowicz@swisstopo.ch just sent a feedback:\r\n Kraftwerkareal in TLM lo??schen. \r\nPermalink: http://map.geo.admin.ch/?topic=swisstopo&X=249317.08&Y=738222.56&zoom=9&lang=de&bgLayer=ch.swisstopo.pixelkarte-farbe. \r\n\r\nUser-Agent: Mozilla/5.0 (Windows NT 6.1; WOW64; Trident/7.0; SLCC2; .NET CLR 2.0.50727; .NET CLR 3.5.30729; .NET CLR 3.0.30729; Media Center PC 6.0; .NET4.0C; .NET4.0E; InfoPath.3; rv:11.0) like Gecko\r\n",
   "email_publisher_attachment{0}": "E:\\FME\\MailAttachments\\revisionservice1660667647748876_Meta.json",
   "email_publisher_from": "webgis@swisstopo.ch",
   "email_publisher_subject": "WG: Customer feedback",
   "email_publisher_attachment{1}": "E:\\FME\\MailAttachments\\revisionservice1660667650432387_Drawing-20150205123136.kml"
}
******************************************************************

davidoesch:
Tackle issues at the root / identify root cause instead of new fearures (maintainability)

Proposals

A) use processes :move all revision messages related to swisstopo to a new folder in the Webgis inbox. Forward after you moved the mail.Grant read access to the inbox to topgis. They can solve it directly without kogis.




B) automate : if there is a mail with a kml  attachment send from revision widget with solely a background layer in the permalink. You can assume it concerns revision forward it automatically via an exchange server ruleset to topgis.




I'm in favor of a)

davidoesch:
Forget b) we will soon have as well possible reporting of public transportation stops --- which will go to bav. With kml. And meta.json.




So it is about to establish a stable process not tools and ruleset.

p1d1d1:
Actually I don't understand where is the problem with UUID or whatever ID.
A mean to identify a message will be helpfull in case of failures somewhere during the process.

Let's take public transportation as an example:
* you forward the messages to BAV and probably you'd also better to store them in a folder "BAV" in the Webgis account.
* for some reasons something fails on the BAV side --> BAV sends an email to the Help Desk saying "Sorry guys something went wrong with the following messages, could you please resend them?"
* when you have hundreds emails in a folder finding a message without an ID is gonna be a mess

Sorry but I don't see at what extent an ID is "tools and ruleset" or business logic. By the way, if you don't want to generate an ID for the messages, as I told in a previous message:
* what about add Date&Time to the metajson filename? This file is always present and date&time can be used as a key.

gjn:
Well, each email is already unique, when taking date of email, message and email address the user specified (can be empty).

When the problem is finding the right email in outlook or any mail client, then having somehting in meta.json does not help you, as usually, you can't search in attachements.

I don't really care in this instance, but maybe improving the process is the best solution here. If something is really annoying, people stop doing it (pressing reply instead of forward); so maybe it's best to keep it annyoing to improve the process :)

p1d1d1:
Ya i'll keep it annoying.
By the way I was suggesting something in the meta name and not in the meta file: something like Meta-20150219153000.json

davidoesch:
proposal :

A) subject field: currently we have only the word "feedback" in it. Add an human readable identifier for ALL mail so it is "feedback (20150228124545)" use the same file naming convention. It searchable in outlook 


B) in addition to a) send it as well to jira and or redmine. So it is directly in the ticketing system 




C) outlook training : epa offers outlook classes, so employees can get training on how to use forward and reply buttons in outlook. Or we externalize to a company which can handle the forwarding.







My prio1 is a) number c) is up to upper mmgt

p1d1d1:
@gjn why has this been closed?

gjn:
Because we want to try to solve the issue on the process side. If in a couple of weeks it's still desired, in 2 clicks the PR is re-created.

p1d1d1:
@gjn @davidoesch 
Cannot believe that adding some info to the name of the meta file (Meta-20150219153000.json) is really a big issue. Why for example we have these info if the name of the KML file?

gjn:
Well, if someone make 'reply' instead of 'forward', he won't send the meta.json file either, no?...so the initial problem still remains.
Let's try to solve it on process side. If in a couple of weeks people still can't 'forward' instead of 'reply', we re-discuss.

p1d1d1:
Yep.
Thanks

davidoesch:
if we change something it will be global:subject field: currently we have only the word "feedback" in it. Add an human readable identifier for ALL mail so it is "feedback (ID 20150228124545)" use the same file naming convention. It searchable in outlook

PR #1292, opened 2015-02-20
More robust test

loicgasser:
Because a test is failing at the moment. I propose to make this test a little more robust.


gjn:
Great stuff!

PR #1294, opened 2015-02-24
test hydrologie_hintergrundkarte

pfanguin:


pfanguin:
This is an aggregate (wms-bgdi_dev), which is in prod for the next deploy.

chsdi/models/vector/bafu.py
```diff
@@ -11,6 +11,19 @@
 Base = bases['bafu']
 
 
+class hintergrundkarte(Base, Vector):
```
procrastinatio: Class name begins with UPPERCASE

procrastinatio:
Minor error, but orhterwise LGTM

chsdi/models/vector/bafu.py
AFoletti: That's not consistent at all... should we go through every model and change to uppercase every class?

Issue #1280, opened 2015-02-14
No image for some photos


cedricmoullet:
http://api3.geo.admin.ch/luftbilder/viewer.html?lang=fr&width=16624&layer=ch.swisstopo.lubis-luftbilder_schwarzweiss&bildnummer=19931850013142&title=ch.swisstopo.lubis-luftbilder-dritte-kantone.ebkey&rotation=None&datenherr=swisstopo&height=16496

ltclm:
All the images with no rotation information in the database are not shown anymore.
The fullres url contains rotation=None and the image is not shown.
I have replaced NULL values which are displayed as "None" in the template with 0 in the database and deployed the tables.

To test:
* disable cache
* search for ebkey ``19931850013142`` in the application 
* or replace ``rotation=None`` with ``rotation=0`` in the url above.

PR #1295, opened 2015-02-24
More robust rotation handling for lubis tooltips/viewer

gjn:
This is to fix https://github.com/geoadmin/mf-chsdi3/issues/1280 where urls to Lubis images contain the text 'None'. The rotation handling is much more robust now in the tooltips and in the viewer.


loicgasser:
LGTM

PR #1296, opened 2015-02-24
keyword parcel,address can be anywhere in the search text

ltclm:
keyword can be anywhere in the search text, as proposed by @gjn it has to end with colon
https://github.com/geoadmin/mf-geoadmin3/issues/2164

### DIFF
--- a/chsdi/views/search.py
+++ b/chsdi/views/search.py
@@ -331,16 +331,17 @@ class Search(SearchValidation):
 
     def _detect_keywords(self):
         if len(self.searchText) > 0:
-            PARCEL_KEYWORDS = ('parzelle', 'parcelle', 'parcella', 'parcel')
-            ADDRESS_KEYWORDS = ('addresse', 'adresse', 'indirizzo', 'address')
-            firstWord = self.searchText[0].lower()
-            if firstWord in PARCEL_KEYWORDS:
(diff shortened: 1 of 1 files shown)


gjn:
I think such a change needs to be discussed with project managers :)

@davidoesch @cedricmoullet Do you think it's a good idea to change our keyword handling in the search from `parcel` to `parcel:*` ?

davidoesch:
Ouch - what is the user need for this?

gjn:
To avoid something like https://github.com/geoadmin/mf-geoadmin3/issues/2164 and it would open up the possibility to have more keywords.

But it's advanced usage and if not desired we can close this.

cedricmoullet:
@davidoesch this was your idea (inspired from gmail search).
I still think something like that is needed.

davidoesch:
Ok... The  initial idea was to handle it like gmail and operating with tags like "tag: parcel 12345"

ponceta:
Why not separate a bit the things? If I want an adress I do not care about layers or parcels and vice versa?

image

gjn:
jenkins test

gjn:
Some tests fail with this PR. See https://jenkins.ci.bgdi.ch/job/chsdi3_pullrequests/32/console

PR #1299, opened 2015-02-25
corrected a legend

AFoletti:


oterral:
@AFoletti isn't  the official max width 480px ?

AFoletti:
You are quite right. I did not think about that because I simply took the existing DE legend and copied it as EN. Could you maybe check how's working on mobile?
(as a side note: we may check all the legends)

oterral:
We have managed the overflow on mf-geoadmin side so it's not an emergency. I create an issue for that, and we take care of this in a future PR.

PR #1303, opened 2015-02-26
Assure spaces are not replaced by plus sign in created kml

gjn:
Fix for https://github.com/geoadmin/mf-geoadmin3/issues/2123


gjn:
It's same fix as for the multiprint: https://github.com/geoadmin/mf-chsdi3/pull/1197

oterral:
Nice it works thx.

Issue #1308, opened 2015-02-27
Migrate from Lubis 2D to Lubis 3D


loicgasser:
ch.swisstopo.lubis-luftbilder_schwarzweiss
ch.swisstopo.lubis-luftbilder_farbe
ch.swisstopo.lubis-luftbilder_infrarot
ch.swisstopo.lubis-bildstreifen

davidoesch:
Third dimension? Which one? Time?

loicgasser:
height

loicgasser:
This has been changed. OK now

PR #1309, opened 2015-02-27
LUBIS: Change geom def to reflect 3D props

loicgasser:
Changes in the model don't change a thing. It is still working with 2D data (all nosetests ok). From what I have seen in the code I don't think it changes anything in how geoms are handled.

I am deploying 3D data to test.
https://github.com/geoadmin/mf-chsdi3/issues/1308


gjn:
LGTM. Also works in app.

PR #1314, opened 2015-03-05
Assures that e2e tests aren't run before/during deploy

gjn:
@procrastinatio This is a PR for your testvarnish branch. It assures that those new tests won't be executed during deploy.


procrastinatio:
OK. Alternatively se also could specify two directory where tests should be run on deploy:

 `./buildout/bin/nosetests chsdi/tests/functional/  chsdi/tests/integration/`

gjn:
Thanks for the review.

PR #1321, opened 2015-03-11
clean code

alcapat:
but html-footer always not visible :-(


chsdi/templates/htmlpopup/verkehrszaehlstellen.mako
```diff
@@ -22,8 +21,7 @@ display:none !important;
 </style>
```
oterral: Move the style tag in the body element pls.

oterral:
Thx @alcapat

PR #1324, opened 2015-03-12
display copyright

alcapat:
Remove code in order to display copyright in the footer


loicgasser:
shall we enter that for today's deploy or too late?

alcapat:
If we can add it for today, it's nice

oterral:
Not too late, it's a minor change

PR #1327, opened 2015-03-13
Do ampersand escaping in mako rendering

gjn:
Result is not different, but it's conceptually better to escape ampersands in the xml renderer than in our code.

### DIFF
--- a/chsdi/views/sitemaps.py
+++ b/chsdi/views/sitemaps.py
@@ -28,7 +28,6 @@ class SiteMaps(SiteMapValidation):
 
 # Maximum number of urls allowed in multi-files
 __MAX_NUM_URLS__ = 5000
-__AMPERSAND__ = '&amp;'
 
 
 @view_config(route_name='sitemap')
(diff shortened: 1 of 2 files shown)


chsdi/views/sitemaps.py
```diff
@@ -127,10 +126,10 @@ def address_part(params):
              .limit(__MAX_NUM_URLS__))
     paths = []
     for res in query.all():
-        paths.append('?' + res.__bodId__ + '=' + res.id + __AMPERSAND__ +
-                     'X=' + str(int(res.X)) + __AMPERSAND__ +
-                     'Y=' + str(int(res.Y)) + __AMPERSAND__ +
-                     'zoom=9')
+        paths.append('?' + res.__bodId__ + '=' + res.id +
```
procrastinatio: maybe using `urllib.urlencode`?

procrastinatio:
Minor commnet. Otherwise OK.

chsdi/views/sitemaps.py
loicgasser: +1

gjn: Really url encode an url that is in an xml file? Why?

Issue #1121, opened 2014-11-11
Add new 'limit' param to search


fredj:
Instead of the hard codded 50, let the user request *less* items from the service.

loicgasser:
@fredj doesn't want more results, he wants less... more than 50 results in search request does not make much sense to me.
So 50 max and per default and user can ask for less results using a parameter `limit` as suggested.

loicgasser:
I'll take care of it next week.

PR #1353, opened 2015-03-26
Add a limit parameter to SearchServer

loicgasser:
https://github.com/geoadmin/mf-chsdi3/issues/1121

This PR allows the user to define a custom `limit` parameter for SearchServer requests.
Actual default value is kept, but it is now also a max value.

### DIFF
--- a/chsdi/lib/validation/search.py
+++ b/chsdi/lib/validation/search.py
@@ -19,6 +19,7 @@ class SearchValidation(MapNameValidation):
         self._returnGeometry = None
         self._origins = None
         self._typeInfo = None
+        self._limit = None
 
     @property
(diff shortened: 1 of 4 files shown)


gjn:
Code LGTM. Didn't test.

A question though: would it make sense to put the limit logic (and the different default limits) into the validation? (I know technically we would need the searchtype there to attribute the default value correctly).

loicgasser:
Yes I debated doing it. I can of course but I find it nice to have it on top of the class view, quick and easy to read. I can adpat if you think it's better not really sure what is more approriate on this one,

gjn:
No need to adapt if it requires to heavy changes in validation component.

PR #1356, opened 2015-03-31
Add 300k scale print

ponceta:
As https://github.com/geoadmin/mf-geoadmin3/issues/2256 wishes. 300k print

Works fine, test [here](http://mf-geoadmin3.dev.bgdi.ch/ltpoa/)


procrastinatio:
Looks correct, but your demo do not work.

ponceta:
Should work now, maybe apache or app was not up to date.

procrastinatio:
OK, checked with a ruler (!)

PR #1358, opened 2015-03-31
Update Pygments version to 2.0

kandre:
Due to version conflict, newer version seems required.

In that case, pipy might be down


kandre:
http://pypi.camptocamp.net/pypi/Pygments/

PR #1359, opened 2015-03-31
Update Babel version to 1.3

kandre:
Sphinx 1.3.1 requires 'babel>=1.3'

and http://pypi.camptocamp.net/pypi/Pygments/ is not down.

@loicgasser what do you think?


kandre:
@loicgasser forget it, my branch was corrupted

PR #1363, opened 2015-04-07
Limit the number of words in the search

gjn:
This renders the search a little more robust against long queries. It limits the number for **words** that the searchText can contain to 10.

I think this is reasonable.

@davidoesch @loicgasser @cedricmoullet Thoughts?

### DIFF
--- a/chsdi/lib/validation/search.py
+++ b/chsdi/lib/validation/search.py
@@ -5,6 +5,7 @@ from pyramid.httpexceptions import HTTPBadRequest
 from chsdi.lib.validation import MapNameValidation
 
 MAX_SPHINX_INDEX_LENGTH = 63
+MAX_SEARCH_TERMS = 10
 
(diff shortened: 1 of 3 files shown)


davidoesch:
A often observed behavior of a user : he enters his complete address (which is longer than 10 char) without looking at search bar and the proposed results - actually he does not use the auto complete function. But if he makes a typo in his long entry --- should be fixed.

As Google does it....

gjn:
It's 10 words, not 10 chars.

The issue you raise is a different one and is partly fixed with the soundex/metaphone indices.

davidoesch:
Correct --- 10 words - actually more than enough. What happens If a user starts entering word number 11?

gjn:
At the moment: nothing (no suggestions/results anymore)

What should happen (front-end is not yet adapted)

davidoesch:
Ok this makes sense.

PR #1371, opened 2015-04-14
Ready to merge: add bfe.windenergieanlagen layer

daguer:


cedricmoullet:
jenkins test

cedricmoullet:
@daguer could you please check the tests ?
@AFoletti could you please review ?
Does it need to be in today's deploy ?

chsdi/models/vector/uvek.py
```diff
@@ -630,6 +630,69 @@ class kleinwasserkraftpotentiale(Base, Vector):
 register('ch.bfe.kleinwasserkraftpotentiale', kleinwasserkraftpotentiale)
 
 
+class WindenergieanlagenFacility(Base, Vector):
+    __tablename__ = 'view_windenergieanlagen_facility'
+    __table_args__ = ({'schema': 'bfe', 'autoload': False})
+    __template__ = 'templates/htmlpopup/windenergieanlagen_facility.mako'
+    __bodId__ = 'ch.bfe.windenergieanlagen'
+    __extended_info__ = True
+    __queryable_attributes__ = ['fac_name']
```
AFoletti: What's the purpose of this line?

chsdi/models/vector/__init__.py
```diff
@@ -5,6 +5,7 @@
 import datetime
 import decimal
 from pyramid.threadlocal import get_current_registry
+from shapely import wkb
```
AFoletti: What for??

AFoletti:
I can't correctly build this branch in its current state. Can you confirm?

daguer:
@cedricmoullet Test is OK and deploy is next week (22.04).

procrastinatio:
These missing legends files are making jenkins failing on 'master'

```
FAIL: test_all_legends_images (chsdi.tests.integration.test_mapservice.TestMapServiceView)
----------------------------------------------------------------------
  File "/var/lib/jenkins/jobs/PR_chsdi3/workspace/chsdi/tests/integration/test_mapservice.py", line 417, in test_all_legends_images
    self.failUnless((layer + '_' + lang) in legendImages, layer + lang)
AssertionError: ch.bfe.windenergieanlagende
```

procrastinatio:
Setting the staging to 'prod' for this layer in the dev BOD is making test failing... So either we:
- Use prod BOD fro jenkins
- Do not set staging to 'prod' because it is not on prod.

AFoletti:
I would leave the staging to test or int, since the deploy of this layer is planned for next week

procrastinatio:
Yes but the whole 'staging' is a mess:
- If you set on 'test', your branch won't test for HtmlPopup and Legend in jenkins, so it is useless.
- If you set on 'prod', all PR will fail.

gjn:
I think @procrastinatio is right.

So we should set the per layer staging to prod only when everything is ready.

procrastinatio:
It would also be nice to run tests against mf-chsdi' master before deploying DB from pg-master to 'dev'. Because now, we have always the impression, code is breaking thing, as it is changes in the BOD, which are not obvious for 'us' developpers.

gjn:
Fully agree.
@ltclm Do you think we could establish a pre-hook for this (or at least a post-hook)? It would also be nice to be able to integrate it into jenkins.
Maybe we could have a jenkins job which checks chsdi3 master branch with bod_master db and run this before we deploy bod_master to bod_dev?
To be discussed.

daguer:
But no legend is missing...
"AssertionError: ch.bfe.windenergieanlagende"
the correct name is ch.bfe.windenergieanlagen_de !
The test is not correct?!?

gjn:
I suspect that the legends exists on your branch, but not on any other branch. So you're branch will not fail, but any other branch (including master) will fail.

procrastinatio:
Hem, the test is correct (all layers are tested like this), the the error message refers to the layer, not the missing file. So it pass on your branch, but fails everywhere else (if yout set it to staging='prod')

procrastinatio:
OK, merging.

PR #1372, opened 2015-04-14
releasenotes for deploy of 22.04.2015

ltclm:


AFoletti:
replaced by https://github.com/geoadmin/mf-chsdi3/pull/1388

PR #1379, opened 2015-04-15
Removing wmts validation warnings during test

procrastinatio:
* Updating lxml library
* Updating OGC Schema to the latest version (2012!)


gjn:
This is so much better. Thanks alot.

PR #1381, opened 2015-04-16
Geodata Staging on integration: switch to prod

gjn:
During deploy, we deploy on integration and do automated and manual tests on the integration environment. When passing from integration to prod, we change a) the back-ends (db and sphinx) and b) the staging for the bod. Both a) and b) lead to troubles in the past during deploys.

This PR adresses troubles/problems introduced by b). We are setting the staging to prod when deploying to integration. We can now test the prod staging _before_ having to deploy to prod.

Note: this only affects the root (mf-chsdi3.int.bgdi.ch)
Note: branches on integration are still using the staging 'integration'. So we are still able to handle the use case for external tests
Note: a) still remains

@procrastinatio @AFoletti @alcapat @daguer @ltclm @cedricmoullet 
I would like your inputs for this.

### DIFF
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -190,6 +190,7 @@ exclude-directories = buildout
 interpreted-options = app_version = __import__('time').strftime('%s')
                       apache_entry_path = '' if options.get('apache_base_path') == 'main' else ('/' + options.get('apache_base_path'))
                       git_branch = __import__('subprocess').check_output(['git', 'rev-parse', '--symbolic-full-name', '--abbrev-ref', 'HEAD']).rstrip()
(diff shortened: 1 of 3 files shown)


gjn:
jenkins test

ltclm:
I think it is a useful enhancement for testing/simulating the bod staging prod on the integration system. As we assume that integration is pre-production this could help to detect bod configuration problems before deploying to prod. :+1:

AFoletti:
+1 on my side too. As @ltclm said, INT was always meant to be 1:1 PROD. We are with this pr a step closer :)

procrastinatio:
OK. Maybe further enhancement:
* Maybe jenkins should also use this _int_ DB
* Remove all branch projects on _int_ (should be on _demo_, no ?)

PR #1385, opened 2015-04-20
mapproxy config updated with new showme tilesets

AFoletti:
- added a new timestamp for the 8 showme layers
- the old showme timestamps were switched from "zoomlevel_max = 26" to "zoomlevel_max = 23" (no need to compute further down due to dataset resolution

I see many other changes not directly involved with my modifications... @procrastinatio , could you maybe check that? Thanks!


procrastinatio:
Looks OK.

PR #1386, opened 2015-04-20
update pdf legends and thumbnails for pk25,pk50 and pk100

ltclm:
see redmine https://redmine.prod.bgdi.ch/issues/4437


ltclm:
@mariokeusen @davidoesch 
legend thumbnails and pdf documents have been replaced by latest version from cms

PR #1396, opened 2015-04-23
Adapt wmts tests

gjn:
This PR adapts the Varnish/wmts tests to be run outside mod_wsgi context as well.

It also makes sure that all official wmts adresses are tested (as published in the documentation on api3.geo.admin.ch).

There is some randomisation in the tiles tests now which
a) toggles the official adresses
b) toggles the referers and the expected results.
This might be controversial...

Note: those tests will be run on jenkins.ci only and only for prod (usually) as part of api3 job: https://jenkins.ci.bgdi.ch/job/api3.geo.admin.ch/ (The gettile tests alone will take 15 minutes)

@procrastinatio I would like your comments.

### DIFF
--- a/chsdi/tests/mapproxy/__init__.py
+++ b/chsdi/tests/mapproxy/__init__.py
@@ -1,0 +1,29 @@
+#-*- coding: utf-8 -*-
+
+import os
+from chsdi.tests.integration import TestsBase
+
+
+class MapProxyTestsBase(TestsBase):
+
+    def setUp(self):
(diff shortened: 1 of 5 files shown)


gjn:
jenkins test

procrastinatio:
OK, but in this case we also have to randomize the tile addresses.

gjn:
If you agree, I'll merge and we can adress open points in additional PR's?

PR #1399, opened 2015-04-23
[WIP] Use different dbhost for int/dev/ci

gjn:
This PR changes the dbhost used for int/dev/ci environments. It uses the backup slave who contains exactly the same db's as the production slaves.
The productions slaves will only be used by production environment now.

We have alot of `OperationalError: (OperationalError) FATAL:  remaining connection slots are reserved for non-replication superuser connections` in your Jenkins build. I assume this is because _everything_ uses the same slaves and we are reaching some limits with our nosetests.

Please review/merge quickly. Then we should rebase our PR's.

### DIFF
--- a/buildout_dev.cfg
+++ b/buildout_dev.cfg
@@ -10,6 +10,8 @@ api_url = //mf-chsdi3.dev.bgdi.ch
 host = mf-chsdi3.dev.bgdi.ch
 # geomadmin
 geoadminhost = mf-geoadmin3.dev.bgdi.ch
+# db host (a slave)
+dbhost = pg-backup.bgdi.ch
 # database staging
 db_staging = dev
 # sphinx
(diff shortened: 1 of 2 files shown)


gjn:
@ltclm Is the address correct?

ltclm:
yes, the address is pg-backup.bgdi.ch .
I think we should discuss this again. This is only a m1.medium instance and:
* it takes longer to update this instance because the limitied network bandwith is slowing down the replication / recovery
* it can start/stop anytime, actually there is a backup script running each night at 03:40 which will stop postgres and take the snapshot.
* furthermore this instance will be used do pg_dump _prod and _master databases and push them to our internal storage

For nosetests (=ci?) yes but for dev and especially for int i propose to use the same infra as production. We should probably try to find a better instance type or add an instance to the pg.bgdi.ch cluster.

gjn:
There error I mentioned above worries me a little bit. Nosetests _do_ add charge to the infra. So I'm not sure it's a good idea to have everything on the same back-end.

Let's discuss it tomorrow.

gjn:
We discussed it. The idea is to have an additional slave, at least for ci, but mabe for dev/int as well. This slave will have it's own DNS and ELB.

PR #1400, opened 2015-04-23
Optimized legend images

ponceta:
Images Optimization (199 of 1942) were not optimized which is only 10% but a few kbs are spared with this. Same process as https://github.com/geoadmin/mf-geoadmin3/pull/2294

https://mf-geoadmin3.dev.bgdi.ch/ltpoa/ (test)


cedricmoullet:
@AFoletti could you please review/merge ?

PR #1404, opened 2015-04-27
Fix typeahead version

oterral:
Fix the search example fixing typeahead version


oterral:
jenkins test

ponceta:
@oterral 
Why don't we use cdnjs on geoadmin?
Why don't we push our api loader on cdnjs?
Would it be a good or a bad idea?
https://github.com/cdnjs/cdnjs

oterral:
I think,  we don't want to be dependent on an external service for geoadmin.

ponceta:
:+1:

PR #1410, opened 2015-05-01
Print include traceback in debug logs

loicgasser:
This is a result of my investigations concering the print yesterday.

This PR adds traceback info in multi print when exceptions are found.

Also linted test files for the print.


gjn:
Nice.

PR #1414, opened 2015-05-05
Remove some unused dependencies

Jenselme:
As we (Bruno Friedmann and myself) discussed with @cedricmoullet this morning, we made some modifications in the API that can interest you. I have split our modifications to ease review and discussion. More PR coming soon.

While studying the dependencies of the API, we have identified some that are in fact not required any more. Here are my commits to delete them.


Jenselme:
Forgot to mention: to bootstrap the project, there is no need to use the package from pypi.camptocamp.net and to force the use of distribute (setuptools works fine) any more.

So, for instance in deploydev.sh
```
python bootstrap.py --version 1.5.2 --distribute --download-base http://pypi.camptocamp.net/distribute-0.6.22_fix-issue-227/ --setup-source http://pypi.camptocamp.net/distribute-0.6.22_fix-issue-227/distribute_setup.py
```
can become
```
python bootstrap.py --version 1.5.2
```

We used our own scripts to deploy so it doesn't appear in this PR.

loicgasser:
You're probably right, I'll give it a try tomorrow.

loicgasser:
Thank you for the cleanup

PR #1415, opened 2015-05-05
Use the `unaccent` function provided by postgresql to remove accents

Jenselme:
I think that you are using a custom postgresql function. As far as I know the unaccent function which comes with postgresql (can be in a postgresql-contrib package on some linux distributions) does the same thing.

For more information: http://www.postgresql.org/docs/current/static/unaccent.html

### DIFF
--- a/chsdi/lib/sqlalchemy_customs.py
+++ b/chsdi/lib/sqlalchemy_customs.py
@@ -10,4 +10,4 @@ class remove_accents(FunctionElement):
 
 @compiles(remove_accents)
 def compile(element, compiler, **kw):
-    return "remove_accents(%s)" % compiler.process(element.clauses)
+    return " unaccent(%s)" % compiler.process(element.clauses)


loicgasser:
This is a custom sql function. I wouldn't remove it as I need to assess the difference between those 2.
@ltclm any idea?

ltclm:
The differences are minimal:
```sql
$ psql -c "SELECT column1 as input, unaccent(column1),remove_accents(column1) FROM (VALUES ('äöü'), ('éèà'), ('CamelCase')) a;" -d bod_master
   input   | unaccent  | remove_accents
-----------+-----------+----------------
 äöü       | aou       | aeoeue
 éèà       | eea       | eea
 CamelCase | CamelCase | camelcase
(3 rows)
```
The ``unaccess`` Extension is already configured on our system.
Im not sure about the impact of this change, concerning sphinx index, i think we can easily replace the custom function with the contrib function.

loicgasser:
It will impact our search accuracy since we won't be able to distinguish ö from o anymore. Not sure it really fits our current needs. It would require some more adaptation in the code as well.
https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/lib/helpers.py#L102

And everywhere we use this function.

loicgasser:
@Jenselme for your information we use sphinxsearch https://github.com/geoadmin/service-sphinxsearch
This is a direct dependency in chsdi.

Jenselme:
@loicgasser I see your problem. Since I didn't know what your custom function did, I thought it was equivalent to unaccent.

I think that possible solutions are:
- To provide your custom SQL function in this repo and document its use
- Use the python `remove_accents` function instead. I would also like to see a better documentation of this function (maybe just precise it is for sphinx and SQL request)

> we use sphinxsearch https://github.com/geoadmin/service-sphinxsearch

I guess this is the service you use for all the search on map.geo.admin.ch (I haven't had the time yet to study search).

cedricmoullet:
@ltclm could you put the code of this function in this repo ?

ltclm:
https://github.com/geoadmin/mf-chsdi3/pull/1422

Jenselme:
The documentation proposed in #1422 is good for me. Thanks!

PR #1416, opened 2015-05-05
Create a template file for configuration

Jenselme:
Based on your README, each developer has his/hers own configuration file tracked by git. We think that this pollutes the repository with lots of unnecessary files. Moreover, it makes the addition of variables cumbersome.

Instead, we decide to create a template configuration file which contains all the possible variables. This file, `buildout_config.cfg.dist` is tracked by git. Each developer must then just copy this file without the extension and replace all needed variables by their proper value. Our script then just have to use this file to deploy for development or production.

I don't know if it makes sense in your workflow.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ buildout_branch.cfg
 deploy/deploy-branch.cfg
 deploy/conf/00-branch.conf
 print/WEB-INF/web.xml
+buildout_config.cfg
--- a/buildout_config.cfg.dist
+++ b/buildout_config.cfg.dist
@@ -1,0 +1,57 @@
+[buildout]
+extends = buildout.cfg
+
+[vars]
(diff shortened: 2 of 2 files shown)


loicgasser:
I don't really see the point of it in our current workflow. @procrastinatio some thoughts?

gjn:
I don't see a need for this.

If you want to add a new variable, you need to add it to `buildout.cfg`, and all other configurations will inherit it from there. Only if you need a different value, you need to adapt it in your personal configuration or in the enviornment specifc configurations.

I agree that personal configurations pollute the root of the project (and we should maybe move them to a dedicated directory), but it's also very convenient to have the personal/environment specific configurations under source control.

procrastinatio:
Dito. 

Whos is @iwibot  ?

PR #1418, opened 2015-05-05
Allow the use of pserve for developpment

Jenselme:
When developing a pyramid application, I find the use of the `pserve` command to be very useful since it can automatically reload after a file change with the command:
```
./buildout/bin/pserve development.ini --reload
```

However, to use with the frontend, you must allow cross domain. This adds a variable to easily make pyramid adds headers when needed. I don't guarantee that all endpoints are covered. Moreover, even if you intend to let this variable empty, it must be present to avoid buildout_filetemplate to crash. I hope I will be able to correct this soon.

The  `pyramid.debug_header` variable must be added to all the buildout_*.cfg used for development.

### DIFF
--- a/chsdi/lib/helpers.py
+++ b/chsdi/lib/helpers.py
@@ -235,3 +235,8 @@ def get_debug_headers(config):
         return {name.strip(): value.strip() for name, value in headers_list}
     else:
         return {}
+
+
+def get_debug_headerlist(config):
+    headers = get_debug_headers(config)
+    return [(name, value) for name, value in headers.items()]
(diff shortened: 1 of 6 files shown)


loicgasser:
We make heavy use of pserve but we use it only with localhost.
We each have a dedicated port.
See:
https://github.com/geoadmin/mf-chsdi3/blob/master/buildout_ltgal.cfg#L9

Using ssh tunnel and remote port forwarding allows us all to enter debug mode whenever we need. I don't think that we should write code in our services for that. but thank you for trying.

Jenselme:
OK. I saw the variables about the server port but I didn't understand how you used it. I haven't thought of ssh tunnels. It may be a better solution than this one.

PR #1369, opened 2015-04-14
Add warning message in pdf 

oterral:
Display third data party copyright in red in the pdf and add a warning message

PR #1420, opened 2015-05-06
Add warning message if third party data are displayed

oterral:
Fix #1369

https://mf-chsdi3.dev.bgdi.ch/teo_kml_warn/

Note from @procrastinatio:

Test with  geoadmin/mf-geoadmin3#2282 !!!

### DIFF
--- a/print/config.yaml
+++ b/print/config.yaml
@@ -95,35 +95,26 @@ layouts:
           width: 300
           items:
             - !text
-              condition: langde
               fontSize: 7
               align: left
               text: |
                 ${dataOwner}
+
+#---------------------------------------
+# Third party data copyright
+#---------------------------------------
+        - !columns
+          condition: thirdPartyDataOwner
+          absoluteY: 60
+          absoluteX: 25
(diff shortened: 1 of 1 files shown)


gjn:
I might be stupid, but I don't see the warning anyhwere. @procrastinatio Could you test this in all formats?

procrastinatio:
This PR is only modify the print template, adding a condition *thirdPartyDataOwner* which is going to display a warning in red. If this variable is not set, you won't see any modification.

This PR has to be tested against the https://github.com/geoadmin/mf-geoadmin3/pull/2282, which really set this  *thirdPartyDataOwner* variable in the print spec sent to the server.

gjn:
I _was_ being stupid.

procrastinatio:
With the very minimalistic description of most PR and issues, it is sometime really difficult to figure out what is wrong and what a PR is supposed to correct. And then, when you have multiple and dependant PR in two projects, with a various changes in data and BOD, good luke

Issue #1424, opened 2015-05-08
Profile broken on FF


procrastinatio:
via helpdesk
```
 Leider funktioniert die Profilerstellung seit gestern nicht mehr!. 
Permalink: http://map.geo.admin.ch/?X=212405.32&Y=651414.54&zoom=5&lang=de&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&catalogNodes=532,533&layers=ch.swisstopo.geologie-geocover&layers_opacity=0.75. 

User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:31.0) Gecko/20100101 Firefox/31.0

```

loicgasser:
?? That's weird, what's new here?

AFoletti:
More to this: on Chrome the profile works in the application but you can't correctly print it

PR #1425, opened 2015-05-08
Enabling CORS for FF

procrastinatio:
Fixing https://github.com/geoadmin/mf-chsdi3/issues/1424


loicgasser:
ok good fix I tested, have to fix some css in geoadmin, weird behaviour in FF.
Saw the same fix on internet for FF and CORS.

PR #1432, opened 2015-05-11
Comment out buggy test

loicgasser:
This is test is super unstable for 2 reasons:

1. http://www.w3.org/2001/xml.xsd is called and is often to reachable (and other links as well)
2. On dev using subprocess in a loop is buggy because of memory issues

We should find another way to validate our getcap for now.


loicgasser:
jenkins test

gjn:
1. I think this should be stable. Alot of people depend on it. It's bad that it wasn't recently, but it seems it is working again (on jenkins api3.geo.admin.ch job is working again)
2. Those tests are running on ci instance...it it might be problematic, I agree. What can we do to improve this?

I consider these tests as quite important and we should enable them again as top priority.

PR #1435, opened 2015-05-12
tooltip for nagra layers

pfanguin:


pfanguin:
This layer is for the next deploy

gjn:
Thanks a bunch.

PR #1437, opened 2015-05-12
Removing file storage nose tests (jenkins)

procrastinatio:
as long as we do not have a clever boto credentials management (similar to shortener)


gjn:
ok. I've also adapted jenkins job definitions accordingly.

Issue #1433, opened 2015-05-12
Enable wmts get cap tests again


gjn:
https://github.com/geoadmin/mf-chsdi3/pull/1432

We should enable those tests again.

@loicgasser Could you take a look and what could be improved?

PR #1440, opened 2015-05-12
Rewrite schema wmts validation

loicgasser:
This a rewrit of our schema validation test for wmts.

We give up on subprocesses and temp files. It's significantly faster.

https://github.com/geoadmin/mf-chsdi3/issues/1433


gjn:
That's awesome. Two things though:

a) This test somehow introduces a side-effect that makes other tests of the same class fail. I couldn't track it down. Do you have any idea?
b) We should also add `21781` to the list of EPSG to be verified in this test.

procrastinatio:
@gjn 

a/ It is not could practice to make loops within a test. Use generator to test each case individually (as in  [test_wmtsgettile.py](https://github.com/geoadmin/mf-chsdi3/blob/gal_schema_validate_getcap/chsdi/tests/mapproxy/test_wmtsgettile.py) )
b/ Yes, *21781* is missing

loicgasser:
@procrastinatio @gjn adapted according to your comments

gjn:
This is good for me.

I think we have to live with the external xsd. Usually, they should have high availability, as I assume other tools/apps are depending on it also.

PR #1443, opened 2015-05-13
Remove loader caching mechanism

gjn:
As discussed between @cedricmoullet and @procrastinatio , this removes the caching mechanism of our loader, which results in ca. 10ms slower response times of the loader but will hopefully get rid of the 5xx errors we sometimes observe.


procrastinatio:
Looks OK to me.

PR #1446, opened 2015-05-15
Only one war to provide print per instance

gjn:
Based on idea by @procrastinatio:

This PR assures that:

1) Only one WAR file is created in `/srv/tomcat/tomcat1/webapps/` to reduce print servers that tomcat needs to handle
2) All configurations (user specific, branch specific) are using this single WAR file
3) If you need to work on print, there is an easy way to work on your own specific WAR (see readme file)

Once this is adopted, we can remove all user specific wars from `/srv/tomcat/tomcat1/webapps/`

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,21 @@ If buildout failed previously, you might need to bootstrap the project again, us
 
     python bootstrap.py --version 1.5.2 --distribute --download-base http://pypi.camptocamp.net/distribute-0.6.22_fix-issue-227/ --setup-source http://pypi.camptocamp.net/distribute-0.6.22_fix-issue-227/distribute_setup.py
 
+# Printing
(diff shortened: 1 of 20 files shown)


procrastinatio:
Very cool. Work as expected (except these days, the mf-chsdi3 project is vanishing too often in _main_).

Should we also introduce an _PRINT_URL_ in the mf-geoadmin3 project ?

procrastinatio:
Note @gjn When you rebuild you project/branche, the _war_ is automagically removed

gjn:
I think we don't need any changes in mf-geoadmin3. It's ok to point print to mf-chsdi3.dev.bgdi.ch/ltjeg/print, as this will simply use the main war already in place.

I'm also hopeful that we keep our `main` a little more stable this way. `main` is updated nightly to the master branch (by jenkins).

PR #1455, opened 2015-05-21
Use JS api in geojson example and adapt loaderjs

loicgasser:
This PR changes the behaviour of loaderjs and adapts the example to use the latest version of our JS API

Before we would always re-perform a request to get the layersconfig and inject it directly in the `loaderjs` template. As a result, we were never using the layers config files generated during the buildout stored in `chsdi/static/js/` .


gjn:
I think there was a reason that we didn't use the static js files. If I remember correctly, during deploy to prod we update the static files from the prod address. But in that moment, the deploy is not yet fully finished; and not all prod instances have the same code. This could theoretically result in a) not fully correct config files and b) not all prod instances having the same static files. The risk is low, but there is a risk.
@procrastinatio What's your take?

loicgasser:
Ok I understand what could be the problem.
Then serverconfig part should not call external addresses but rather use the application in which it is contained.

loicgasser:
Alternatively we could make those files static.

loicgasser:
Something like https://github.com/geoadmin/mf-chsdi3/pull/1044/files but only for layersconfig.

gjn:
Well, we need to assure that it's always up to date. The approach above would be manual process before each deploy, right?

loicgasser:
Yes
Incoming requests on layersconfig service in prod could also simply return those static files. if we use static configurations for topics and layersconfig service map.geo.admin.ch would probably load a little faster as well.

loicgasser:
But let's do it step by step.

gjn:
Yes, if and when we put the bod in code (not decided yet) then it'll be alot easier.

loicgasser:
I'll use api_url from settings instead for now. But let's discuss that next week.

loicgasser:
ok I replaced it with https://github.com/geoadmin/mf-chsdi3/commit/e94724f20a396f61515686f59ae19d899f770cd5 ready to merge

gjn:
Thanks alot.

PR #1444, opened 2015-05-13
[WIP] Api example: Geojson integration

kandre:
[Testlink](http://mf-chsdi3.dev.bgdi.ch/kan_geojson_example/examples/geoadmin_geojson.html)

This PR should be merged as soon modifications in our API are merged (and hence this code adapted accordingly and squashed).

PR #1456, opened 2015-05-21
API geojson example: Fix Styles urls as host dependant

kandre:


kandre:
@loicgasser I just ve seen your message in https://github.com/geoadmin/mf-chsdi3/pull/1444.

This is my solution for fixing those static style url. It need a rebase, does it worth or we use directly your solution?

kandre:
This PR can be closed, fix came with #1455

PR #1459, opened 2015-05-26
Removing boto credentials for dynamodb

gjn:
I seems like those are not needed. At least, it works wihtout those and we don't have them in the shortener service to access dynamodb. Maybe this is what causes prod to fail on the files service.

This is for the deploy branch only for now.


gjn:
This is working fine on int. I'll deploy this branch to prod.

PR #1462, opened 2015-05-26
Boto credentials for files dynamodb access

gjn:
This PR removes the need for boto credentials to access dynamodb table; as it's also done in for the shortener [1]. It applied the same patch as https://github.com/geoadmin/mf-chsdi3/pull/1459, which was deployed to prod for testing. It seems to work (verify again tomorrow morning).

I've also added a little error handling.

@procrastinatio Please review and test.

[1] https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/views/shortener.py#L57

PR #1463, opened 2015-05-27
Apply patch for files service

gjn:
Patch as in https://github.com/geoadmin/mf-chsdi3/pull/1462/files for production. This enables boto debugging!


gjn:
This will be deployed to prod now.

PR #1464, opened 2015-05-27
tooltip for sgpk maechtigkeit lockergesteine

pfanguin:


procrastinatio:
OK, tooltips and legend are i.O

PR #1471, opened 2015-05-27
fix vector style and legends png

pfanguin:


loicgasser:
LGTM @gjin can we merge to have it in prod today?

PR #1472, opened 2015-05-27
Fixing files service on production env

gjn:
This reflects the status that is currently deployed to production from the rel_150520 branch.

If production turns out to be stable until, we can simply merge this and deploy it tomorrow.

Meanwhile, @loicgasser works on overal boto improvements (shared connections, etc) which I think are important but probably need heavier testing still.


gjn:
Ok, we merge this for tomorrows deploy. Service seems to be stable on prod with this.

PR #1473, opened 2015-05-27
Better error handling, centralize dyanomdb conns and tables

loicgasser:
With this PR everyone should have a .boto file to serve the application in dev mode.
Connections are created once and reused over time (best practice).
Only boto layer2 interface is used now.

### DIFF
--- a/chsdi/models/clientdata_dynamodb.py
+++ b/chsdi/models/clientdata_dynamodb.py
@@ -2,7 +2,9 @@
 
 import pyramid.httpexceptions as exc
 
-from boto.dynamodb import connect_to_region
+from boto import connect_s3
+from boto.dynamodb2 import connect_to_region
+from boto.dynamodb2.table import Table
 
 '''
 CREATE a table
@@ -42,11 +44,39 @@ table.delete()
 # http://boto.readthedocs.org/en/latest/boto_config_tut.html
 
 
-def get_table(table_name='shorturl'):
-
-    # url_short is the pkey
+def _get_dynamodb_conn(region='eu-west-1'):
     try:
-        conn = connect_to_region(region_name='eu-west-1')
-        return conn.get_table(table_name)
+        conn = connect_to_region(region)
+    except Exception as e:
+        raise exc.HTTPBadRequest('DynamoDB: Error during connection init %s' % e)
+    return conn
+
+
+def _get_s3_conn(profile_name='geoadmin_filestorage'):
+    # TODO use profile instead when correctly installed
+    try:
+        conn = connect_s3(profile_name=profile_name)
+    except Exception as e:
+        raise exc.HTTPBadRequest('S3: Error during connection %s' % e)
+    return conn
+
+
+conn_dynamodb = _get_dynamodb_conn()
+conn_s3 = _get_s3_conn()
+
+
+def get_dynamodb_table(table_name='shorturl'):
+    try:
+        table = Table(table_name, connection=conn_dynamodb)
+    except Exception as e:
+        raise exc.HTTPBadRequest('Error during connection to the table %s' % e)
+    return table
+
+
+def get_bucket(request):
+    bucket_name = request.registry.settings['geoadmin_file_storage_bucket']
+    try:
+        bucket = conn_s3.get_bucket(bucket_name)
     except Exception as e:
         raise exc.HTTPBadRequest('Error during connection %s' % e)
+    return bucket
--- a/chsdi/views/files.py
+++ b/chsdi/views/files.py
@@ -5,12 +5,12 @@ import base64
 import time
 import zipfile
 import StringIO
-import datetime
 
 from boto.dynamodb2.exceptions import ItemNotFound
 
 from boto.exception import S3ResponseError
 from boto.s3.key import Key
+from boto.utils import parse_ts
 
 from pyramid.view import view_config, view_defaults
 import pyramid.httpexceptions as exc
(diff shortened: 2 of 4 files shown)


gjn:
Great simplification. I'll review in detail tomorrow.

This needs a rebase because of https://github.com/geoadmin/mf-chsdi3/pull/1472 that was just merged. We'll test this PR in detail after tomorrows deploy.

loicgasser:
ok, rebased

loicgasser:
boto updated and parse_ts function replaced as discussed.

cedricmoullet:
@gjn @loicgasser @procrastinatio can we deploy this tomorrow ?

loicgasser:
Apparently for next one

loicgasser:
@procrastinatio could you have a look and merge if ok with you?

procrastinatio:
I have already pretty much tested, but until it is deployed, we won't know for sure.

gjn:
How do we load test this a little bit?

We're moving from a per-request connection object to a global connection object...

loicgasser:
boto is wrapper around Amzon REST API, so it s fine.
http://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html

gjn:
i was more thinking if the boto connection object could handle concurrency (same object used in multiple threads in parallel). as this might happen under load. usually, you need to design your classes/objects specifically for. i did not look in documentation nor in source code.

chsdi/models/clientdata_dynamodb.py
```diff
@@ -42,11 +44,39 @@
...
+    # TODO use profile instead when correctly installed
+    try:
+        conn = connect_s3(profile_name=profile_name)
+    except Exception as e:
+        raise exc.HTTPBadRequest('S3: Error during connection %s' % e)
+    return conn
+
+
+conn_dynamodb = _get_dynamodb_conn()
+conn_s3 = _get_s3_conn()
```
gjn: this is called at webapp start, correctly? if yes, raising httpbadrequest in the functuons does not really make sense. also, will it stip startup of application?

gjn: 'functions' and 'stop', sorry.

chsdi/views/files.py
```diff
@@ -165,8 +107,11 @@ def __init__(self, request):
                 self.file_id = id
             try:
                 key = self.bucket.get_key(self.file_id)
-            except:
-                raise exc.HTTPInternalServerError('Cannot access file with id=%s' % self.file_id)
+            except S3ResponseError as e:
+                raise exc.HTTPInternalServerError('Cannot access file with id=%s: %s' % (self.file_id, e))
+            except Exception as e:
+                raise exc.HTTPInternalServerError('Cannot access file with id=%s: %s' % (self.file_id, e))
```
gjn: both raised exceptions are exactly the same, so why having both?

gjn:
according to http://ls.pwd.io/2013/06/parallel-s3-uploads-using-boto-and-threads-in-python/, boto connections are not thread safe (and using new connection for every request is very fast). the official doc doesn't specifically mention thread safety for connection. Also, on SO, they say that boto connections are not thread safe: http://stackoverflow.com/questions/19523680/is-boto-library-thread-safe

Unless I get other information, I consider this PR critical and it shouldn't have been merged.

procrastinatio:
Hem, is there any difference between a dynamodb and a s3 connection ? we only did use the same approach for s3 as for dynamodb (which obviously work).

gjn:
we have introduced 2 global objects that are not thread-safe that could be used in 2 different threads on 2 concurrent requests. the behaviour then is undefined. in the lucky case, the request fails, in the unlucky case, the application crashes. in any case, it's a bug.

gjn:
workers in tileworkers are processes, right?

loicgasser:
connect has to be called once. then we create new buckets coming from the same connection. it works like sqlalchemy... http://stackoverflow.com/questions/11622548/boto-connect-xxx-method-and-connection-pools

loicgasser:
I can find other links that will tell you the same.

loicgasser:
If we want to be on the safe side, I can of course change that and create a new conn for each request as it was before. no problem and takes no time.

gjn:
read the first answer from your link. it says exactly what I'm saying.

an alternative is to use connectionpool class, where the doc explicitely assures thread safety.

currently, it's a bug and we should correct it.

loicgasser:
connection object uses a pool such that each connection is used only by a single thread at a time.

But let s not be dramatic and change that for a good sleep.

gjn:
could you tell me where you've read that, because i didn't find anything.

loicgasser:
same link last answer. To be sure I should dig in the code and see if connect_s3 inherits/uses ConnectionPool class, it did nt use to be threadsafe so you re probably right to worry. Anyway will revert this part.

loicgasser:
so connect_s3 is wrapping S3Connection which extends AWSConnection which uses ConnectionPool class. so not a bug and we re safe.

gjn:
ouf, great. thanks alot for verifying and sorry for steering the pot.

since which version is it thread-safe!

gjn:
we can discuss the other in-code comments on tuesday.

chsdi/views/files.py
loicgasser: I thought to distinguish errors coming from S3 and otther erros.
S3ResponseError (meaning a request has been sent)
and
Exception (any other error)
text could be different.

loicgasser:
you were right to make me check it out. I just trusted what I red but for this kind of things it's better to double check :D

PR #1477, opened 2015-05-28
Remove reframe redirection configuration

gjn:
As we are using [the service now directly](https://github.com/geoadmin/mf-geoadmin3/pull/2375), we can remove the apache rewrite for the reframe service. This gets rid of another mod_proxy usage.

To be merged _after_ todays deploy.


loicgasser:
LGTM

gjn:
@procrastinatio @cedricmoullet 87% of our 504 errors from backend over the last month has been from this reframe redirection (either because geodesy service timed out or mod_proxy was bad). So removing this will reduce 504 heavily.

Of the remaining 504 errors, almost all are from the SearchServer service.

procrastinatio:
Nice.

PR #1479, opened 2015-05-28
Assure that files service works with IE9

gjn:
IE9 does not send any content-type header. There's no way to change that. This means that storing kml files coming from IE9 simply doesnt work.

This PR changes the files service so it will accept requests coming from IE9 without checking the content-type header.


gjn:
Nosetests fail. I'm on it.

gjn:
Should be good now.

loicgasser:
you rule

gjn:
:)
IE9 rules. Another 2.5 hours taken from my working time.

procrastinatio:
OK, I have a look. If good, merging and redeplyoing...

PR #1480, opened 2015-05-28
Silence warnings on the fixrights task (as in ga3)

gjn:
This PR silences the fixrights command (chgrp and chmod), whitout changing functionality. This way, we get rid of useless messages in console, both when deploying by hand and in jenkins logs.

If something is wrong with rights, we'll know anyway.


procrastinatio:
OK.

PR #1484, opened 2015-06-01
Support geometry in body of profile request

gjn:
This PR supports the transfer of the geometry (`geom` parameter) in the body of a POST request instead as part of the URL. Both versions are still supported. Having the geometry in the body assures that we don't run into varnish troubles because of too long URLS.

I PR in mf-geoadmin3 will follow that will show how it's to be used in the front-end.

We could emergency deploy this quite fast and switch map.geo and veloland to use this method for their profile calculations.


gjn:
This is actually not needed if the requests are formulated correctly.

Issue #1478, opened 2015-05-28
MapProxy: WMTSCapabilities.EPSG.2056.xml not in all langs


loicgasser:
https://mf-chsdi3.int.bgdi.ch/1.0.0/WMTSCapabilities.EPSG.2056.xml not available in all langs

procrastinatio:
So [FR](https://wmts10.geo.admin.ch/1.0.0/WMTSCapabilities.EPSG.2056.xml?lang=fr) and [DE](https://wmts10.geo.admin.ch/1.0.0/WMTSCapabilities.EPSG.2056.xml?lang=de)

PR #1490, opened 2015-06-04
Considering the lang param in mapproxy requests

procrastinatio:
Even though *en* and *it* are usually not translated

See https://github.com/geoadmin/mf-chsdi3/issues/1478


loicgasser:
QSA for query string append. LGTM

loicgasser:
We are potentially testing the outside world, but how else could we test apache.

PR #1491, opened 2015-06-04
create a link to the pdf only if the file exist

alcapat:


procrastinatio:
Is there no other way to do this, because it may break test.

loicgasser:
Agree external links are always subject to instability. But I think we already those things in some other templates. My question is how often are those pdfs subject to changes? Because we could put this info in the DB otherwise.

alcapat:
jenkins test

Issue #1488, opened 2015-06-04
Store kml files gzipped


gjn:
According to [1], it should be possible to store and serve gzipped files (in our case kml) to S3. In order to do that, we need to gzip the file and put the content-encoding header accordingly when storing the file to S3.

The downside is that clients that don't support the deflate option (Accept-Encoding: deflate, gzip,...). Most browser do support this options, but I didn't find a list of clients _not_ supporting it. Does anyone know?

[1] http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/ServingCompressedFiles.html#CompressedS3

procrastinatio:
Yes, it works (I have done tests), but for now requests on _public.geo.admin.ch_ only count for a few [ppm](http://en.wikipedia.org/wiki/Parts-per_notation). So if we really want to reduce the bandwith, we would gain much more improving S3 caching or further optimizing PNG

The main issue is not [storing files in S3](https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/views/files.py#L207), it is managing metadata when backuping and restoring.

Maybe a duplicate of https://github.com/geoadmin/mf-geoadmin3/issues/2362

loicgasser:
We need to have only one type of files per bucket. (so no kmz)
I propose that we gzip all the kml files we already have.

Then we would know that Content-Encoding is always 'gzip' and Content-Type is always 'application/vnd.google-earth.kml+xml' and Content-Length is set automatically on S3.

@gjn delflate option is transparent is all browsers nowadays. http://webmasters.stackexchange.com/questions/22217/which-browsers-handle-content-encoding-gzip-and-which-of-them-has-any-special

procrastinatio:
So please merge https://github.com/geoadmin/mf-chsdi3/pull/1473 before as a lot as been refactored, efore I'll dot it.

gjn:
References misses mobile browsers. And maybe exotic clients that _could_ be used to download those kml's.

procrastinatio:
@gjn Varnish will gunzip the file for the client not sending *Accept-Encoding: gzip*. So, no souci.

gjn:
This is working fine. Thanks alot.

PR #1493, opened 2015-06-05
Gziping all ingooing kml

procrastinatio:
Fixing https://github.com/geoadmin/mf-chsdi3/issues/1488

Ingooing KML are stored as gzipped in S3 (with *Content-Encoding: gzip*) because S3 cannot preprocess files _on-the-fly_
Consequence, client has to support *Content-Accept: gzip*.

### DIFF
--- a/chsdi/views/files.py
+++ b/chsdi/views/files.py
@@ -3,8 +3,9 @@
 import uuid
 import base64
 import time
-import zipfile
+import gzip
 import StringIO
+
 
 from boto.dynamodb2.exceptions import ItemNotFound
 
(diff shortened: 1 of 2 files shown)


cedricmoullet:
Nice.
Do I assume correctly that old files will stay as they are ? This will work and will be applied only for the new files ?
Do you know cases where Content-Accept: gzip is not supported ?
Is there any consequence for the download function ? User should receive a pure KML and not a gzipped KML.

procrastinatio:
No, we have to merge and deploy this PR. The new files are going to be gziped. I will then download them, gziped and upload them again file. Otherwise, we'll have to backup the metadata or guess it when restoring the backup...

The compression is only at the HTTP level, it is fully transparent for the application (except the time to dowlnoad and to decompress). 

In a web context, as long as web browser support HTTP/1.1 and the request is not modified some way (plugin!) , it should not be a problem.

But now, HAproxy/Varnish is dezipping the file, effectivelly defaiting the aim of the whole thing.

Not through HAproxy/Varnish --> gzipped
http://public.dev.bgdi.ch.s3.amazonaws.com/NlPlsAFjQVO7GKrJ9zr2vQ

Through HAproxy/varnish --> gunzipped
http://public.dev.bgdi.ch/NlPlsAFjQVO7GKrJ9zr2vQ

chsdi/views/files.py
```diff
@@ -128,25 +129,40 @@ def options_files_collection(self):
...
-            if ziped_data is not None:
+            if content_encoding == 'gzip' and ziped_data is not None:
                 data = ziped_data
             try:
-                k = Key(self.bucket)
+                k = Key(bucket=self.bucket)
                 k.key = self.file_id
                 k.set_metadata('Content-Type', mime)
+                k.content_type = mime
+                k.content_encoding = content_encoding
```
gjn: what happens if content_encoding is None?

For consistency, why not also do `self.key.set_metadata('Content-Encoding', content_encoding)` as in update?

procrastinatio: 1/ If a _metadata_ is `None`it is not set (won't be sent to the client)
2/ Because it is not necessary

gjn:
jenkins test

gjn:
ok. thanks alot. I'll merge once jenkins is happy.

PR #1498, opened 2015-06-09
Adapt fixrights to use different recipe (bash support)

gjn:
This adresses a bug introduced in our build by an adaption to the fixrights part.

We switch the recipe used and use the bash shell for the command. I think this should fix the troubles building using our scripts.


gjn:
jenkins test

PR #1500, opened 2015-06-09
Make sure the connection is created if failing at app start

loicgasser:
As discussed this is very unlikely but with this approach we make sure the connection will be created if it fails for whatever reason during application start.


gjn:
Thanks alot for this!

Issue #1496, opened 2015-06-09
Error 500


cedricmoullet:
http://api3.geo.admin.ch/rest/services/api/MapServer/identify?geometryType=esriGeometryEnvelope&geometry=-Infinity,-Infinity,Infinity,Infinity&imageDisplay=0,0,0&mapExtent=0,0,0,0&tolerance=0&layers=all:ch.swisstopo.swissboundaries3d-gemeinde-flaeche.fill,ch.swisstopo.swissboundaries3d-land-flaeche.fill&returnGeometry=false&lang=fr

gjn:
geometry parameter is wrong. Was this request created by map.geo.admin.ch?

cedricmoullet:
no

procrastinatio:
So now, we have a 400 'Bad Request'.

PR #1510, opened 2015-06-12
Do not break on identify bad requests

procrastinatio:
See https://github.com/geoadmin/mf-chsdi3/issues/1496

In this PR, it gives valuables information on why the SQLAlchemy did fail. Or would it rather better to silently return no feature(s) ?


gjn:
I like it alot! It reflects the error better (in the #1496, it's really a bad request by the client).

Expected impact on performance?

procrastinatio:
Also added a test.

I don't think it has any singificant impact on performance (I do not test parameters)

gjn:
Thanks alot. Great stuff.

Issue #1522, opened 2015-06-18
OWSChecker internal error


procrastinatio:
The OWSChecker returns a not well specified “Internal Server Error” with the following WMS:

http://ows.sigeom.ch/ows/public

PS: the service, anyway, does not validate in OGC TEAM Engine

via heldpesk

PR #1527, opened 2015-06-23
OWSChecker: do not attempts to check WMS-10 if WMS-09 fails

procrastinatio:
The support of `GetFeatureInfo`is not mandatory in WMS (but is in _egeo0056_). So if a service has no support for `GetFeatureInfo`(WMS-09), do not try to get which mie-type are supported (WMS-10).

See https://github.com/geoadmin/mf-chsdi3/issues/1522


procrastinatio:
jenkins test

PR #1534, opened 2015-06-29
ready to merge for next deploy: update web-dav

daguer:


procrastinatio:
Unfortunately `www.bfe-gis.admin.ch`doe not support `https`. So will end with mixed http/https content in the extended tooltip.

PR #1536, opened 2015-06-30
ready to merge for next deploy: update bfe web-dav url

daguer:


gjn:
@daguer for tomorrow?

cedricmoullet:
@daguer Is https an option ? Or could we use the proxy ?

daguer:
@gjn  Yes, it's for today...
@cedricmoullet We can use the proxy, I adapt my PR

gjn:
Not a fan of proxy. If it works that way, we should let it as it is.

cedricmoullet:
I would reconsider the useage of the proxy. Of course, the problem should be solved at the source. But in this case, our application is considered unsecure...

PR #1541, opened 2015-07-01
LUBIS tooltips: using unicode in mako

procrastinatio:
Do not force to `str`if content may be unicode...


gjn:
I wonder why we didn't had this issue earlier. @ltclm are there new lubis datasets? New import mechanisms.

procrastinatio:
We have to check all mako, because I a pretty sure it not the only occurence.

PR #1545, opened 2015-07-03
[WIP] Merge all the ol targets

### LINKED ISSUE #1544: ol3-serverconfig and ol3-update fail with fresh checkout
Probably build system changed with update to 3.6 and those 2 targets were not adapted in the buildout.cfg files.

This means we can't deploy branches anymore and fresh checkouts fail as well.

@oterral Could you have a look?

oterral:
Update the ga* files with:

`./buildout/bin/buildout install ol `

Fix #1544

[Test](http://mf-chsdi3.dev.bgdi.ch/teo_ol_conf/)

### DIFF
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -11,8 +11,6 @@ parts = eggs
         doc
         node-modules
         lessc
-        ol3-install
-        ol3-serverconfig
 develop = .
 index = http://pypi.camptocamp.net/pypi/
 allow-hosts = pypi.camptocamp.net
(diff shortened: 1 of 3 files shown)


buildout.cfg
```diff
@@ -255,49 +253,24 @@ on_update = true
...
-      rm -f ${buildout:directory}/chsdi/static/js/serverconfig*.js &&
-      API_URL=${vars:api_url} ${buildout:bin-directory}/python build-ga.py build/serverconfig &&
-      cp ${buildout:directory}/chsdi/static/js/ol3/build/*config*.js  ${buildout:directory}/chsdi/static/js/ &&
-      cd ${buildout:directory}
-    fi
-
-[ol3-update]
-recipe = collective.recipe.cmd
-on_install = true
+on_install = false
```
gjn: I thinks we need to put this to `true`, otherwise buildout will fail the first time we do `buildout/bin/buildout install ol`.

gjn:
With a fresh checkout of this branch, the loader is failing.

gjn:
I think we need to keep the ol3-serverconfig part as part of all builds. It generates the layer list the loader will be using based on current db (as time of build). Maybe discuss with @procrastinatio because this setup is not ideal.

gjn:
This is high prio for me, because a clean checkout won't work.

buildout.cfg
oterral: If you set it to true `ol` target is always triggered when you do a  `buildout install`   and it's totally useless.

oterral:
I've look at it with @procrastinatio  server-config seems to be useless. But I will verify.

oterral:
We have fixed it, the problem was the variable `api_url` which didn't exist. It's ok the now the fresh build works

buildout.cfg
gjn: It's described as such (install ol) in the PR :)

oterral: Yes you 're right it breaks the first time. I've retry and it's ok with `on_install = true` .

oterral:
`install ol` fixed

buildout.cfg
oterral: and `ol`is not called on the basi build

gjn:
LGTM. Thanks alot.

PR #1514, opened 2015-06-15
ch.swisstopo.swissnames3d

ltclm:

PR #1546, opened 2015-07-06
releasenotes for deploy 20150708

ltclm:


ltclm:
this jenkins test wont work until https://github.com/geoadmin/mf-chsdi3/pull/1514 is merged
ch.swisstopo.swissnames3d is already on staging prod but the corresponding model is missing in the code.

PR #1547, opened 2015-07-06
do not test htmlpopup for aggregated layer

ltclm:
only parent layers are tested, analogous to other tests

### DIFF
--- a/chsdi/tests/integration/test_mapservice.py
+++ b/chsdi/tests/integration/test_mapservice.py
@@ -430,7 +430,8 @@ class TestMapServiceView(TestsBase):
         from sqlalchemy.orm import scoped_session, sessionmaker
         val = True
         DBSession = scoped_session(sessionmaker())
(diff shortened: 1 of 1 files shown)


ltclm:
will solve buildout problem like this one: https://github.com/geoadmin/mf-chsdi3/pull/1514

ltclm:
@loicgasser thanks for support with this buildout issue

loicgasser:
jenkins test

loicgasser:
same error occurs did it work when running nosetests locally? Could you try to do all tests in TestMapServiceView?

ltclm:
yes, there is a prod layer without model because the pr https://github.com/geoadmin/mf-chsdi3/pull/1514 is pending.
I have set the layers staging to int and deployed the bod.
Should work now.

Probably the nosetests staging should be definable by parameter instead of using always using 'prod'.

ltclm:
jenkins test

PR #1554, opened 2015-07-09
deploy/hooks/post-restore

gjn:
Same problematic as https://github.com/geoadmin/mf-geoadmin3/pull/2486

### DIFF
--- a/deploy/hooks/post-restore
+++ b/deploy/hooks/post-restore
@@ -1,0 +1,11 @@
+#!/bin/sh
+#
+# variables set here:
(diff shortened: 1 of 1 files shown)


cedricmoullet:
Do we really need that (since it was not done previously) ?

gjn:
The default hook of deploy is run when we don't define it. And the default only runs a start.

See `cat /etc/deploy/hooks/post-restore`

cedricmoullet:
Ok, clear

PR #1556, opened 2015-07-13
Remove simplejson and ordereddict from dependency list

Jenselme:
Use json from stdlib instead


loicgasser:
Thanks for the cleanup. I tested examples and the loader, everything seems to work as expected.

PR #1557, opened 2015-07-13
Use requests instead of httplib2

Jenselme:


gjn:
What's the benefit of using request instead of httplib?

Jenselme:
- It avoids one dependency
- I find it more coherent: this project requires a library to do http requests. Currently you are using requests in some places (in most of the tests) and httplib2 in other. Moreover, requests seems to be the best way to do http requests in python.

procrastinatio:
Yes, I do agree `requests`has also nice features like `retries`, uws.

chsdi/views/ogcproxy.py
```diff
@@ -71,9 +71,9 @@ def ogcproxy(self):
         else:
             raise HTTPNotAcceptable()
 
-        if content.find('encoding=') > 0:
-            m = re.search("encoding=\"(.*?)\\\"", content)
-            doc_encoding = m.group(1)
+        if resp.encoding:
+            doc_encoding = resp.encoding
+            print(doc_encoding)
```
gjn: Is this print statement needed? Or debugging artefact?

Jenselme: Not at all, it is a debugging artefact. PR updated.

gjn:
jenkins test

gjn:
The ogc proxy does not work anymore (returns 406). Sample request: /ogcproxy?url=http%3A%2F%2Fmapserver1.gr.ch%2Fwms%2Fadmineinteilung%3FSERVICE%3DWMS%26REQUEST%3DGetCapabilities%26VERSION%3D1.3.0

Jenselme:
Strange, I don't have this problem: ogcproxy works fine for me.

gjn:
@procrastinatio Could you verify if it works or not?

loicgasser:
what about urllib2? We probably have a ton of deps in makos. I would double check everything first.

procrastinatio:
`urllib2`is a standard, no ?
Usually the ogcproxy fails because the requested url, in this case a WMS service, fails

loicgasser:
It is, just saying if we want only one way to make http request then it would make sense to change it everywhere. If it s just to remove a dependency then it's just fine.

gjn:
When I do it, ogcproxy failed with multiple requests. I'll retest (not for tomorrow anyway).

Jenselme:
> It is, just saying if we want only one way to make http request then it would make sense to change it everywhere.

Indeed. I missed urllib2 because it is in the standard library and is used in ows_checker which I haven't modified yet. Should I update this PR to use request there too?

procrastinatio:
jenkins test

procrastinatio:
All three tests pass for me.

loicgasser:
jenkins test

loicgasser:
I wouldn't spend too much time on removing urlib2. It s fine like this, but we need to figure out why tests fail with jenkins.

cedricmoullet:
Will be fixed in another PR

PR #1562, opened 2015-07-15
Adding ch.swisstopo.zeitreiehen to WMTS

procrastinatio:
...but using png instead of pngjpeg (as in map.geo.admin.ch), as this custom image format is not well supported amongs desktop GIS.

See in Redmine  [BGDI Datenhaltung - Datenintegration #4835]

[GetCapabilities](http://mf-chsdi3.dev.bgdi.ch/mom_wmts_zeitreiehen/1.0.0/WMTSCapabilities.xml)

[Beispiel](http://mf-chsdi3.dev.bgdi.ch/mom_wmts_zeitreiehen/examples/ol3_lv95_all.html?layer=ch.swisstopo.zeitreihen&timestamp=19051231&format=png&base_url=http://mf-chsdi3.dev.bgdi.ch/mom_wmts_zeitreiehen)

ArcGis (10.2.2) use and display only png source, while QGis (2.8.1) display both sources, but actually only `.png`is used.

### DIFF
--- a/chsdi/static/doc/examples/ol3_lv95_all.js
+++ b/chsdi/static/doc/examples/ol3_lv95_all.js
@@ -59,6 +59,7 @@ $.ajax({
     var timestamps_nb = 0;
     var config = data[layer];
     data['ch.kantone.cadastralwebmap-farbe'] = cadastralCfg;
+    data['ch.swisstopo.zeitreihen'].format = 'png';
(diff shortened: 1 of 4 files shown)


procrastinatio:
jenkins test

procrastinatio:
According to @beat-github it is OK on QGis. Also good on ArcGis from my point of view.

PR #1563, opened 2015-07-15
Use Search.LOCATION_LIMIT instead of hardcoded value when parsing location results

Jenselme:
The maximum limit is set to 50 in the class but when parsing the result, a hardcoded value of 20 was used. This commit fixes this.

### DIFF
--- a/chsdi/views/search.py
+++ b/chsdi/views/search.py
@@ -15,7 +15,7 @@ from chsdi.lib import mortonspacekey as msk
 
 class Search(SearchValidation):
 
-    LIMIT = 50
+    LOCATION_LIMIT = 50
     LAYER_LIMIT = 30
     FEATURE_LIMIT = 20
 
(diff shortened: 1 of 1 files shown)


cedricmoullet:
@loicgasser could you please review and merge if ok ?

loicgasser:
I think this is a BFS limitation if I recall correctly. We can t return more than 20 addresses per request. I personally think it s a shame. So ok from a technical point of view, but what about political @davidoesch ?

Jenselme:
What is BFS? What I find strange with the current implementation is that you can query up to 50 addresses in sphinx but only format the fist 20.

davidoesch:
50 restriction is imposed by the cantons as owner of the addresses. Is a legal based constraint we on our side can not change.see the various discussion in the api forum

gjn:
@davidoesch The question is not 50, but 20. Is the BFS limit on addresses 20 or 50?

If it's 50, we can merge this.

davidoesch:
The cantonal restriction is 20

gjn:
So we can't merge this, I'm afraid.

gjn:
Good news. We can increase this limit to 50. So I'll merge this.

PR #1565, opened 2015-07-17
update ch.bfe.energieforschung

daguer:


chsdi/models/vector/uvek.py
```diff
@@ -488,64 +488,42 @@ class ENERGIESTAEDTE2000AUFDEMWEG(Base, Vector):
 
 
 class ENERGIEFORSCHUNG(Base, Vector):
-    __tablename__ = 'energieforschung'
+    __tablename__ = 'energieforschung_new'
```
gjn: Is this a new layer? Or replacing existing layer?

Does that mean that old links with the old technical name won't work anymore?

gjn:
In future, can someone from dataintegration do a review as well?

chsdi/locale/fi/LC_MESSAGES/chsdi.po
```diff
@@ -2217,64 +2217,13 @@ msgid "ch.bafu.laerm-strassenlaerm_tag"
...
 
 msgid "ch.bafu.landesforstinventar-baumarten"
-msgstr "Baumarten (LFI)"
+msgstr "Spezias da bostgs (IFN)"
 
 msgid "ch.bafu.landesforstinventar-totholz"
-msgstr "Totholz (LFI)"
+msgstr "Laina morta (IFN)"
 
 msgid "ch.bafu.landesforstinventar-waldanteil"
```
gjn: Is it correct to remove all those translations below? And why only for fi and not the other languages?

gjn:
When doing translations, look at the diff..

chsdi/models/vector/uvek.py
gjn: Stupid me, technical layer name didn't, so ignore my comment.

chsdi/locale/fi/LC_MESSAGES/chsdi.po
daguer: I have only change one layer (energieforschung).
May be a problem with the rebase of my branch...
I do it again on Monday...

loicgasser: Translations have been doubled. Translations are there so it is correcting a bad rebase. LGTM to me.

loicgasser:
jenkins test

loicgasser:
Jenkins error is unrelated see: https://github.com/geoadmin/mf-chsdi3/pull/1568

PR #1568, opened 2015-07-20
Fix address limit

loicgasser:
This fixes the new limit introcuced in https://github.com/geoadmin/mf-chsdi3/pull/1563


loicgasser:
trivial

PR #1577, opened 2015-07-27
Update translations

alcapat:
New translations for the deploy of 5th august


loicgasser:
This is redundant with other PR like https://github.com/geoadmin/mf-chsdi3/pull/1580/files so I'll close.

PR #1586, opened 2015-07-30
extended model has to use the same pk as base model

ltclm:


gjn:
Error in Jenkins is unrelated. I think this fixed it. Thanks alot.

PR #1587, opened 2015-07-30
Update ol3 api

### LINKED ISSUE #1585: Avoid potential variable clash in global scope
Which steps can we do to avoid that our ga.js compilation conflicts with other libararies global definitions? It happened twice in the last month that we had conflicts with 2 different external libraries.

See message in api group:

```
ga.js seems to overwrite the global variable Jx ( on Line 690: Jx=M(Vw.....).

As consequence, Jx-lib no longer works when included before the GeoAdmin-API.  I changed the order of including of the js-Files: Now Jx works as used to but I have no idea if there will be some bug in the API...

As Jx is used by many Applications I recommend to use an other name for this variable or use it only in local context.
```

oterral:
Fix compiled build file

 Fix #1585


gjn:
Why was the version updated?

oterral:
You mean the version number?

gjn:
yes

PR #1589, opened 2015-07-30
Increase verbosity when sitemap tests fail

gjn:
Adds a little more info when sitemap tests fail. Right now, we don't know if it's because of OSError thrown or because they are indeed not validating.

OSError is probably because of weak instance type of Jenkins.


gjn:
Build failures related to OSError (CPU/Memory related) now look like `AssertionError: 0 != 5342`. Just as this PR's build now: https://jenkins.ci.bgdi.ch/job/PR_chsdi3/654/console

Issue #1594, opened 2015-08-05
wmtsCapabilities.4326 in ArcMap


rfrancel:
WMTS tiles from this URL are not visible in ArcMap (but are in QGis):
https://mf-chsdi3.int.bgdi.ch/1.0.0/WMTSCapabilities.EPSG.4326.xml

PR #1597, opened 2015-08-07
Using the GLOBAL_GEODETIC tilematrixset scheme

procrastinatio:
In order to use mapproxy WMTS in 3D, we have to offer the tiles with the GLOBAL_GEODETIC scheme, wich start at -180, 90.

*  fixing axis order
*  using the the global GLOBAL_GEODETIC scheme
* custom LOD for WGS1984 (not the one defined in ech0056!)
* all topics present in map.geo.admin.ch

This also fix https://github.com/geoadmin/mf-chsdi3/issues/1594


cedricmoullet:
@procrastinatio please squash

PR #1607, opened 2015-08-18
Use assertion in non-TestCase unit tests

gjn:
failUnless does not exist on this class because it's not based on TestCase. We don't want to base this class on TestCase because we want to use generators. So we need assert instead.

This PR also uses the new puppet managed http_proxy.

@procrastinatio Please review.

### DIFF
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -126,7 +126,7 @@ mapproxy_wsgi_options=processes=4 threads=32
 #zadara
 zadara_dir=/var/local/cartoweb/downloads/
 #http proxy
-http_proxy=http://ec2-54-93-109-29.eu-central-1.compute.amazonaws.com:80
+http_proxy=http://ec2-52-28-118-239.eu-central-1.compute.amazonaws.com:80
 # shortener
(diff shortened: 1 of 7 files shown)


gjn:
The `test_wmtsgettile.py` tests run up the memory until the end of the tests, consuming **more than 4GB** of RAM at the end. I think that's bad. I tried to mitigate this by using the yield in the get_tiles function, but that didn't help.

@procrastinatio Do you have an idea on how we could limit memory consumption during those tests?

loicgasser:
LGTM.
MapProxy is probably the one to blame and which can explain the RAM consumption. We're performing more than 1'000 requests on MapProxy in a very short time window...

loicgasser:
From doc:
http://www.python-requests.org/en/latest/user/advanced/#keep-alive

gjn:
I was able to find a solution. Memory consumption is low now, it's faster and in addition tests can now be called the same way as all other tests (they will be found by nose without specifiying path/testname).

@loicgasser Please review.

loicgasser:
Thanks a lot!

Issue #1601, opened 2015-08-12
SearchServer 500 error


cedricmoullet:
http://api3.geo.admin.ch/rest/services/api/SearchServer?lang=de&type=locations&searchText=Bussigny-pr%E8s-Lausanne%20jonction&

produces a 500:

[Wed Aug 12 11:23:40 2015] [error] [client 10.220.5.136] mod_wsgi (pid=30722): Exception occurred processing WSGI script '/var/www/vhosts/mf-chsdi3/private/chsdi/buildout/parts/modwsgi/wsgi'.
[Wed Aug 12 11:23:40 2015] [error] [client 10.220.5.136] Traceback (most recent call last):
[Wed Aug 12 11:23:40 2015] [error] [client 10.220.5.136]   File "/var/www/vhosts/mf-chsdi3/private/chsdi/buildout/eggs/pyramid-1.5.2-py2.7.egg/pyramid/router.py", line 242, in __call__
[Wed Aug 12 11:23:40 2015] [error] [client 10.220.5.136]     response = self.invoke_subrequest(request, use_tweens=True)
[Wed Aug 12 11:23:40 2015] [error] [client 10.220.5.136]   File "/var/www/vhosts/mf-chsdi3/private/chsdi/buildout/eggs/pyramid-1.5.2-py2.7.egg/pyramid/router.py", line 217, in invoke_subrequest
[Wed Aug 12 11:23:40 2015] [error] [client 10.220.5.136]     response = handle_request(request)
[Wed Aug 12 11:23:40 2015] [error] [client 10.220.5.136]   File "/var/www/vhosts/mf-chsdi3/private/chsdi/buildout/eggs/pyramid-1.5.2-py2.7.egg/pyramid/tweens.py", line 21, in excview_tween
[Wed Aug 12 11:23:40 2015] [error] [client 10.220.5.136]     response = handler(request)
[Wed Aug 12 11:23:40 2015] [error] [client 10.220.5.136]   File "/var/www/vhosts/mf-chsdi3/private/chsdi/buildout/eggs/pyramid_tm-0.12-py2.7.egg/pyramid_tm/__init__.py", line 99, in tm_tween
[Wed Aug 12 11:23:40 2015] [error] [client 10.220.5.136]     reraise(*exc_info)
[...]

Issue #1481, opened 2015-05-28
Error 500: SearchServer


loicgasser:
http://api3.geo.admin.ch/rest/services/api/SearchServer?lang=de&type=locations&searchText=Le%20Ch%E2b&

procrastinatio:
subscriber...

gjn:
Duplicate of https://github.com/geoadmin/mf-chsdi3/issues/1601

Issue #1053, opened 2014-10-08
SearchServer should deal with bogus request


procrastinatio:
Many requests on the `SearchServer`originating from _http://portal.ejpd.admin.ch/etcmobile_ ending in HTTP 500

curl 'http://mf-chsdi3.dev.bgdi.ch/rest/services/api/SearchServer?lang=de&type=locations&searchText=tr%E8fl&'

```
[Wed Oct 08 11:18:26 2014] [error] [client 10.220.4.40]     return codecs.utf_8_decode(input, errors, True)
[Wed Oct 08 11:18:26 2014] [error] [client 10.220.4.40] UnicodeDecodeError: 'utf8' codec can't decode  byte 0xe8 in position 2: invalid continuation byte
```

gjn:
Funny fact:

`nice_subscriber` approach works in mf-chsdi3 using `buildout/bin/pserve production.ini`. It does NOT work using `buildout/bin/pserve production.ini` and it does NOT work as mod_wsgi behind apache.

As indicated [here](https://groups.google.com/forum/#!topic/pylons-discuss/qLglgrgpoho), it might be related to the debug_toolbar - but I can't see how it gets in the way in mod_wsgi.

gjn:
See #1609

Issue #990, opened 2014-09-04
Error 500 for SearchServer


cedricmoullet:
http://api3.geo.admin.ch/rest/services/api/SearchServer?lang=de&type=locations&searchText=Br%FCttis&

[Thu Sep 04 21:53:34 2014] [error] [client 10.220.4.40]   File "/home/ltmoc/mf-chsdi3/chsdi/subscribers.py", line 10, in add_renderer_globals
[Thu Sep 04 21:53:34 2014] [error] [client 10.220.4.40]     event['_'] = request.translate
[Thu Sep 04 21:53:34 2014] [error] [client 10.220.4.40] AttributeError: 'Request' object has no attribute 'translate'

PR #1609, opened 2015-08-20
Improve handling of non-utf8 URLs and Query Strings

gjn:
This adresses https://github.com/geoadmin/mf-chsdi3/issues/1601, https://github.com/geoadmin/mf-chsdi3/issues/1481, https://github.com/geoadmin/mf-chsdi3/issues/990 and https://github.com/geoadmin/mf-chsdi3/issues/1053

Alot of groundworks has been done by @procrastinatio in #1053 - which contained the solution by having a 'nice' subscriber that catched the decoding problem and throws an Bad Request Error (error Code 400) instead of failing silently and returning a 500.

The problem was that this fix didn't work when the debugtoolbar was active. But the fix works as expected by either [commenting out the debugtoolbar in the development.ini](https://github.com/geoadmin/mf-chsdi3/blob/master/development.ini.in#L15) file or by using the production.ini file (which is used under main).

I've also updated the debugtoolbar version in this PR; even though the newer version still has this problem.


gjn:
In last 30 days, we had 2000 of such 500 errors on the SearchServer. Not that this fix also fixes _any_ request which doesn't contain utf-8 encoded characters.

loicgasser:
debugtoolbar had a lot of issues with utf-8 encoding and it was really not handled properly.
See https://github.com/geoadmin/mf-chsdi3/pull/1209
I believe a lot of fixes have been made so and hope it's now safe to upgrade. Thanks for your investigations as I know it is time consuming and can be really frustrating.

Issue #1611, opened 2015-08-20
Wrong htmlPopup request should return correct error code.


gjn:
Prio1 as it was found in production error logs.

`/rest/services/api/MapServer/ch.bafu.bundesinventare-bln/htmlPopup` currently returns `500` (internal error, see error under [1], but it should return a `404` (not found) error as the request is wrong (it doesn't contain the featureID).

[1]
```
...
[Wed Aug 19 17:15:29 2015] [error] [client 10.220.5.136]     cursor.execute(statement, parameters)
[Wed Aug 19 17:15:29 2015] [error] [client 10.220.5.136] DataError: (DataError) invalid input syntax for integer: "htmlPopup"
[Wed Aug 19 17:15:29 2015] [error] [client 10.220.5.136] LIGNE 3 : WHERE bundinv.bln.gid = 'htmlPopup'
[Wed Aug 19 17:15:29 2015] [error] [client 10.220.5.136]                                   ^
[Wed Aug 19 17:15:29 2015] [error] [client 10.220.5.136]  'SELECT bundinv.bln.gid AS bundinv_bln_gid, ST_AsEWKB(bundinv.bln.the_geom) AS bundinv_bln_the_geom, bundinv.bln.bln_name AS bundinv_bln_bln_name, bundinv.bln.bln_obj AS bundinv_bln_bln_obj, bundinv.bln.bln_fl AS bundinv_bln_bln_fl \\nFROM bundinv.bln \\nWHERE bundinv.bln.gid = %(gid_1)s' {'gid_1': u'htmlPopup'}
```

PR #1612, opened 2015-08-20
Handle exception when query has wrong type for id

gjn:
This is a fix for https://github.com/geoadmin/mf-chsdi3/issues/1611

We handle the `DataError` exception now that can be thrown when there is a type mismatch between the given ID and the type of the ID field in the db.

### DIFF
--- a/chsdi/tests/integration/test_mapservice.py
+++ b/chsdi/tests/integration/test_mapservice.py
@@ -286,6 +286,9 @@ class TestMapServiceView(TestsBase):
         resp = self.testapp.get('/rest/services/ech/MapServer/ch.bafu.bundesinventare-bln/0', status=404)
         resp.mustcontain('No feature with id')
 
+        resp = self.testapp.get('/rest/services/api/MapServer/ch.bafu.bundesinventare-bln/htmlPopup', status=404)
+        resp.mustcontain('No feature with id')
+
     def test_feature_valid(self):
         resp = self.testapp.get('/rest/services/ech/MapServer/ch.bafu.bundesinventare-bln/362', status=200)
         self.failUnless(resp.content_type == 'application/json')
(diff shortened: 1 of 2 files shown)


loicgasser:
I have one as well

loicgasser:
https://github.com/geoadmin/mf-chsdi3/pull/1614

The difference is that I stop iterating through the tables as I assume all IDs types are identical for a same layer.
We could also validate the param before hand.

loicgasser:
I think we could use: 

https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/views/features.py#L463
maybe rename the function and add a few tests.

gjn:
Ok, I make my case for 404. :)

End user doesn't care about type of id's. We have a mixture in that some layers use text, others use integers, etc, and client can't determine which type is used for which layers.

So all he is passing is text and the feature either exists or it doesn't. So 404 seems more correct to me.

gjn:
Or, here's no reason why `https://api3.geo.admin.ch/rest/services/ech/MapServer/ch.bfs.gebaeude_wohnungs_register/3175980_toto` is 404 and /rest/services/api/MapServer/ch.bafu.bundesinventare-bln/toto` should be 400.

loicgasser:
Yes ok for 404 makes sense, what about validating the id before sending it to the DB?
It's not in my PR but it would be nice. what do you think?

gjn:
Not sure it's worth it. It would need to be done for every id separately (we can have mix of valid and invalid id's). Hitting the db with wrong type might have performance (and other consequences), but it's happening quite rarely (only when doing it by hand).

I have now qualms when you want to do it though.

loicgasser:
we could use validators in the vector class.
http://docs.sqlalchemy.org/en/rel_1_0/orm/mapped_attributes.html#simple-validators

loicgasser:
at least for ids.

loicgasser:
I'll open an issue with low prio as I think it's probably a good thing as a ton of people are sending a bunch of nonsense requests to our servers.

gjn:
Ok.

Thanks for the review.

PR #1614, opened 2015-08-20
Raise bad request on type mismatch

loicgasser:


gjn:
Fixes same as https://github.com/geoadmin/mf-chsdi3/pull/1612

PR #1617, opened 2015-08-21
Pythonic way for html popup tests

gjn:
This move the html popup tests (and related layers test) in a separate test file and tests all layers (htmlpopup, legend and legendimages) in a more pythonic way by using generator functions.

This has the advantage that each layer test is visible separately in the output and the output is not blocked for 60 seconds.

New: we now also test the extendedHtmlPopup when the model says it has one.

It's also a first step in the direction to test more than just one feature for a given layer and a given htmlpopup.


gjn:
jenkins test

Issue #1613, opened 2015-08-20
500 error on htmlPopup


cedricmoullet:
http://api3.geo.admin.ch/1439987023/rest/services/gewiss/MapServer/ch.bafu.typisierung-fliessgewaesser/53502/htmlPopup?coord=710036.3326100261,197316.0993199869&imageDisplay=1920,1075,96&lang=de&mapExtent=708120.3326100261,196302.0993199869,711960.3326100261,198452.0993199869

loicgasser:
solved with https://github.com/geoadmin/mf-chsdi3/pull/1612

loicgasser:
It now return 404

gjn:
This is a different kind of error which is not solved with #1612. Looks like buggy mako to me:

```
Traceback (most recent call last):
  File "/home/ltjeg/mf-chsdi3/buildout/eggs/pyramid_mako-1.0.2-py2.7.egg/pyramid_mako/__init__.py", line 148, in __call__
    result = template.render_unicode(**system)
  File "/home/ltjeg/mf-chsdi3/buildout/eggs/Mako-1.0.0-py2.7.egg/mako/template.py", line 452, in render_unicode
    as_unicode=True)
  File "/home/ltjeg/mf-chsdi3/buildout/eggs/Mako-1.0.0-py2.7.egg/mako/runtime.py", line 803, in _render
    **_kwargs_for_callable(callable_, data))
  File "/home/ltjeg/mf-chsdi3/buildout/eggs/Mako-1.0.0-py2.7.egg/mako/runtime.py", line 835, in _render_context
    _exec_template(inherit, lclcontext, args=args, kwargs=kwargs)
  File "/home/ltjeg/mf-chsdi3/buildout/eggs/Mako-1.0.0-py2.7.egg/mako/runtime.py", line 860, in _exec_template
    callable_(context, *args, **kwargs)
  File "/home/ltjeg/mf-chsdi3/chsdi/templates/htmlpopup/base.mako", line 68, in render_body
    ${self.table_body(c, lang)}
  File "/home/ltjeg/mf-chsdi3/chsdi/templates/htmlpopup/fliessgewaesser_typ.mako", line 13, in render_table_body
    <%
TypeError: cannot concatenate 'str' and 'NoneType' objects
```

loicgasser:
Mmm bad copy paste.

gjn:
[...]

PR #1619, opened 2015-08-21
Testing popups for specific layers

gjn:
This PR is the result after the discussion in https://github.com/geoadmin/mf-chsdi3/issues/1613

It depends on https://github.com/geoadmin/mf-chsdi3/pull/1617

Currently, our nosests (run on jenkins) test our makos with exactly 1 feature, which is always the same for each layer that we are testing. This doesn't test all border cases.

This PR adds a script (`./html_popup.sh`) which allows to test a single layer with more features. Specifically, it allows to test
- a specific layer only
- specify the staging of the layer as it is in the connected bod (so 'dev' layers can be tested)
- randomize the feature(s) that are used for testing
- choose the number of feature(s) you want to use for testing, including all features available (nuclear option).

See the script itself for details about the configuration.

Run this script by hand when integrating a new layer or adapting an existing layer and you want to test with more features than 1.

This is a 'brute' force attempt and there's no systematic way to test for border cases (null, None, etc)


ltclm:
As discussed today:

Given a model definition like the one for layer ``ch.swisstopo.swissboundaries3d-gemeinde-flaeche.fill`` : 
https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/models/vector/stopo.py#L82

The following query can be used to get a list of feature ids with all possible attribute states:
```sql
select max(id)
	, CASE pow(0,length(gemname::text)) WHEN 1 THEN 'empty string' WHEN 0 THEN 'contains value' ELSE 'NULL' END 		-- Text
	, CASE pow(0,length(gemflaeche::text)) WHEN 1 THEN 'empty string' WHEN 0 THEN 'contains value' ELSE 'NULL' END		-- Numeric
	, CASE pow(0,length(perimeter::text)) WHEN 1 THEN 'empty string' WHEN 0 THEN 'contains value' ELSE 'NULL' END		-- Numeric
	, CASE pow(0,length(kanton::text)) WHEN 1 THEN 'empty string' WHEN 0 THEN 'contains value' ELSE 'NULL' END		-- Text
from tlm.swissboundaries_gemeinden
group by
	CASE pow(0,length(gemname::text)) WHEN 1 THEN 'empty string' WHEN 0 THEN 'contains value' ELSE 'NULL' END 		-- Text
	, CASE pow(0,length(gemflaeche::text)) WHEN 1 THEN 'empty string' WHEN 0 THEN 'contains value' ELSE 'NULL' END		-- Numeric
	, CASE pow(0,length(perimeter::text)) WHEN 1 THEN 'empty string' WHEN 0 THEN 'contains value' ELSE 'NULL' END		-- Numeric
	, CASE pow(0,length(kanton::text)) WHEN 1 THEN 'empty string' WHEN 0 THEN 'contains value' ELSE 'NULL' END		-- Text
```

The query differentiates the following attribute states:
```
NULL: pow(0,length(attribut::text)) == NULL
empty string: pow(0,length(attribut::text)) == 1
string: pow(0,length(attribut::text)) == 0
```

You have to cast all the attributes from the model to text ``::text``, otherwise the ``length`` function is not working.

This query can be used to analyze the variety of all the atributes from the model. If the make is using only a subset of these attriubte, the test query can simulate the worst case scenario (if mako changes).

chsdi/tests/integration/test_layers.py
```diff
@@ -0,0 +1,108 @@
...
+from sqlalchemy.sql.expression import func
+
+from chsdi.models.bod import LayersConfig
+from chsdi.models import models_from_name
+
+
+class LayersChecker(object):
+
+    def __enter__(self):
+        def num(s):
```
loicgasser: s.isdigit()

loicgasser:
```
if s.isdigit()
 return long(s)
else:
 return s
```

chsdi/tests/integration/test_layers.py
```diff
@@ -0,0 +1,108 @@
...
+                return long(s)
+            except ValueError:
+                return s
+
+        self.testapp = TestApp(get_app('development.ini'))
+        self.session = scoped_session(sessionmaker())
+
+        # configuration via environment. Default is None for all
+        self.staging = os.environ.get('CHSDI_STAGING') if os.environ.get('CHSDI_STAGING') is not None else 'prod'
+        self.onlylayer = os.environ.get('CHSDI_ONLY_LAYER')
```
loicgasser: Minor:
I would transform that to a list per default and split on  a separator so that we can test several layers using one variable.

Just change the test to q[0] not in self.onlylayers etc..

loicgasser:
2 minor comments, address if you feel like it otherwise feel free to merge. Thanks for the improvement.

chsdi/tests/integration/test_layers.py
gjn: Adapted. Much better.

chsdi/tests/integration/test_layers.py
gjn: well, if need grows, will do it. Script is thought to be used by dataintegrators to test a layers.

PR #1620, opened 2015-08-21
Add 'tooltip' and 'queryableAttributes' to layersConfig

gjn:
This is the result of the discussion in https://github.com/geoadmin/mf-chsdi3/pull/1096

Long term goal is to get rid of `queryable` and `selectbyrectangle` attribute in LayersConfig and in the bod. In the future, we will have

`tooltip`: layer has a tooltip. Thus can be used in front-end for left mousclick or selection by rectangle
`queryableAttributes`: if the layer can be used in the query tool, it will have an array of attributes that can be queried. If not, it will not contain this property
`searchable`: [existing] layers can be search view search tool (and will have a sphinx index)

This is the first step by adding `tooltip` and `queryableAttributes` where appropriate. Client side code will be adapted as well to reflect these changes.

### DIFF
--- a/chsdi/models/bod.py
+++ b/chsdi/models/bod.py
@@ -4,7 +4,7 @@ from sqlalchemy import Column, Text, Integer, Boolean
 from sqlalchemy.dialects import postgresql
 
 from chsdi.lib.helpers import make_agnostic
-from chsdi.models import bases, models_from_name
+from chsdi.models import bases, models_from_name, get_models_attributes_keys
 
 Base = bases['bod']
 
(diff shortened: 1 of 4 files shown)


chsdi/models/bod.py
```diff
@@ -125,6 +126,18 @@ def layerConfig(self, params):
...
 
+        # adding __queryable_attributes__ if they have them
+        models = models_from_name(self.layerBodId)
+        if models is not None:
+            queryable_attributes = []
+            for model in models:
+                if hasattr(model, '__queryable_attributes__'):
+                    queryable_attributes.extend(model.get_queryable_attributes_keys(params.lang))
+
+            queryable_attributes = list(set(queryable_attributes))
```
loicgasser: Kind of the same than https://github.com/geoadmin/mf-chsdi3/blob/gjn_attr/chsdi/views/layers.py#L112
Is is worthed to factorize a little?

chsdi/tests/integration/test_mapservice.py
```diff
@@ -495,6 +495,15 @@ def test_layersconfig_with_callback(self):
     def test_layersconfig_wrong_map(self):
         self.testapp.get('/rest/services/foo/MapServer/layersConfig', status=400)
 
+    def test_layersconfig_queryable_attributes(self):
+        resp = self.testapp.get('/rest/services/all/MapServer/layersConfig', status=200)
+        self.failUnless(resp.content_type == 'application/json')
```
loicgasser: assertTrue

ltclm:
obsolete attributes selectbyrectangle and queryable removed from bod, they are still in the view.
https://github.com/geoadmin/db/commit/4e60b13649fd39d706159ef85a688f524ddb365b

PR #1623, opened 2015-08-25
Use assertTrue instead of deprecated failUnless

loicgasser:


loicgasser:
Trivial sef merged all tests green

PR #1624, opened 2015-08-25
[WIP] Don't return a geometry for features that are too big

gjn:
This is an alternative/addition to https://github.com/geoadmin/mf-chsdi3/pull/1621

In #1621, we return a `413` (requested feature is too big) whenever we do a request that contains one result that contains a feature which wbk string is longer than 1M characters. This protects our backends against very long requests.

The approach in #1621 has the downside that for some requests, the identify will also ignore features that are well withing the limit, or features from other layers that are also within the limits (We can combine serveral layers for one single `identify` request).

This PR tries to address the downside of the above by, in all cases, return a valid result (`200`) but to remove (and not parse) the geometry of features hitting the specified limits. 

There's a downside to this as well: the clients need to adapt their code to handle features with no geometry. This has not been verified in mf-geoadmin3 yet.

### DIFF
--- a/chsdi/views/features.py
+++ b/chsdi/views/features.py
@@ -496,9 +496,15 @@ def _format_search_text(columnType, searchText):
         raise exc.HTTPBadRequest('Find operations cannot be performed on geometry columns')
 
 
+def has_long_geometry(feature):
+    if len(getattr(feature, feature.geometry_column_to_return().name).data) > 1000000:
+        return True
+    return False
+
+
 def _process_feature(feature, params):
     # TODO find a way to use translate directly in the model
-    if params.returnGeometry:
+    if params.returnGeometry and not has_long_geometry(feature):
         f = feature.__geo_interface__
         # Filter out this layer individually, disregarding of the global returnGeometry
         # setting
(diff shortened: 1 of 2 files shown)


procrastinatio:
Great, I think the WSGI gateway is really a bottleneck, so if we may avoid transfer too much data, it's better.

chsdi/views/features.py
```diff
@@ -496,9 +496,15 @@ def _format_search_text(columnType, searchText):
         raise exc.HTTPBadRequest('Find operations cannot be performed on geometry columns')
 
 
+def has_long_geometry(feature):
+    if len(getattr(feature, feature.geometry_column_to_return().name).data) > 1000000:
```
procrastinatio: Maybe use a constant

procrastinatio:
OK, looks good. Tested with:
`rest/services/all/MapServer/ch.swisstopo.geologie-geocover/956757?format=geojson"`

But in dataset `ch.swisstopo.geologie-geocover`we have only 10 geometries bigger than 1'000'000 out of 77'427.

Integrators should be aware of this _limitation_.

gjn:
why was this merged? it was WIP because it wasn't tested with frontend cide, which could break.

or was it tested?

cedricmoullet:
We discussed it andMarc tested it. @procrastinatio could you confirm ?

gjn:
This does not break the application, but there are console errors when hitting a feature which is too big. This should be handled correctly in front-end.

e.g. https://mf-geoadmin3.dev.bgdi.ch/src/?ch.swisstopo.geologie-geocover=910652

Issue #1630, opened 2015-08-27
Identify with wrong parameters issues wrong error


gjn:
http://api3.geo.admin.ch/rest/services/ech/MapServer/identify?geometry=NaN,NaN&geometryFormat=geojson&geometryType=esriGeometryPoint&imageDisplay=1920,979,96&layers=all:ch.bfs.gebaeude_wohnungs_register&mapExtent=NaN,NaN,NaN,NaN&tolerance=5

It returns 500. Instead, it should return 400.

PR #1632, opened 2015-08-27
Assure that NaN parameters are parsed correctly

gjn:
This fixes https://github.com/geoadmin/mf-chsdi3/issues/1630

It also adresses other cases where a client transferes `NaN` instead of valid numbers. In case of complex geometries, this even avoids segfaulting the application.

We treat any `NaN` in the request that should be parsed to a float as invalid.


loicgasser:
Good one, I didn't know NaN was considered a float. LGTM

Issue #1628, opened 2015-08-26
Zeitreihen print contains data that is not officially released


gjn:
When printing 1:25'000 with https://map.geo.admin.ch/?X=288058.07&Y=682938.66&zoom=7&topic=ech&lang=de&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen&time=2013&layers_timestamp=20131231 and do a multipage, the multipage pdf is containg the last lk25'000 of 2006. But in the application, we have the last lk25'000 of 2013.

The print request json is in [1].

My assumption is that the releases service used ([2] in this case) is delivering the wrong timestamps. In fact, as can be seen by [2], which was triggered by [1], it doesn't contain 2006 but not 2013. It seems like our metadata and the data we should in the application don't match.

@ltclm @daguer Could it be that on the last updates, we forgot to update the meta data? Or what could be wrong?

[1] http://api3.geo.admin.ch/rest/services/ech/MapServer/ch.swisstopo.zeitreihen/releases?imageDisplay=1%2C1%2C96&mapExtent=682969.245278%2C287928.662778%2C682978.064722%2C287937.482222
[2] 
```
[...]

PR #1635, opened 2015-09-02
Multiprint: Assure passed parameters are normalized to correct imageD…

gjn:
This is a fix for https://github.com/geoadmin/mf-chsdi3/issues/1628

Before, we assumed that the passed parameters `display` and `bbox` correspond to 96dpi. In fact, this was not the case. With this PR, we calculate the real dpi based on the passed parameters and normalize to 96dpi as 96dpi is need for the `releases` service to select correct model for given resolution.

### DIFF
--- a/chsdi/views/printmulti.py
+++ b/chsdi/views/printmulti.py
@@ -56,19 +56,6 @@ def _zeitreihen(d, api_url):
     return timestamps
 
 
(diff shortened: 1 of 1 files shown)


cedricmoullet:
@procrastinatio could you please review

PR #1636, opened 2015-09-02
new layer ch.swisstopo.lubis-luftbilder_schraegaufnahmen

ltclm:


loicgasser:
@ltclm
mmmhh, could you use a Date format for flightdate and avoid code duplication in mako, by importing the related helper functions instead?

loicgasser:
you can copy what has been done in Zeitreihen and the other lubis.

ltclm:
you're right, should be changed within https://github.com/geoadmin/mf-chsdi3/issues/1703

PR #1647, opened 2015-09-10
CRS:84 to WMS config for 3D

procrastinatio:
WMS 1.3.0 honor the axis order, so EPSG:4326 and CRS:84 have switch order, for client supporting or not this feature.

Test link: [GetCapabilities](http://mf-chsdi3.dev.bgdi.ch/mom_mapproxy_crs84/mapproxy/service?REQUEST=GetCapabilities)


oterral:
Thx it works

PR #1648, opened 2015-09-10
Use wms-bod translated (DE) for test because wms-bod.map never used

alcapat:
Add the parameter lang=de in url link because actually we test the wms-bod not translated and this wms is never used. So it's better to test the translated version (DE by default).


AFoletti:
oh yes!

PR #1653, opened 2015-09-16
[WIP] Set config nose for CI

alcapat:


nose_run.sh
```diff
@@ -13,6 +13,9 @@ while getopts "ipa" opt; do
   case "$opt" in
   i) conf_to_use="buildout_nose_int.cfg"
      ;;
+  c) conf_to_use="buildout_nose_ci.cfg"
+     nosetests_options="-I test_links.py -I mapproxy/test_wmtscapabilities.py -I test_wmtsgettile.py -I test_varnish.py"
+     ;;
```
gjn: we should use -e parameter with `-e *e2e*` (not sure if it's correct, needs to be validated)

PR #1662, opened 2015-09-18
Queryable attributes fixes

AFoletti:
This PR fixes several problems with queryable attributes:
- some missing translations are added
- some missing queryable attributes are added
- makos and translations are adapted in oder to use the automatically generated msg_id
- a nasty regexp error, resulting in a 400 in several layers, is fixed


gjn:
Great stuff.

Issue #1638, opened 2015-09-04
Apache crashes


gjn:
In our logs [1], I could see a couple of apache crashes due to out-of-memory exceptions. These could be observed only on our prod instances that server mapproxy and profile. (vpc-mf0-prod5 and vpc-mf0-prod6)

No analysis done yet.

[1] https://logs.bgdi.ch/kibana/#/dashboard/elasticsearch/apache%20crashes

procrastinatio:
The following 4 instances vpc-sphinxsearch-prod-2, vpc-sphinxsearch-prod-1, vpc-wms-prod-5 and vpc-wms-prod-4 are responsible for 2/3 of the crashes...

procrastinatio:
This is a general issue on *all* prod servers. They all crash one day or another:_puppetmaster, wms-prod, shpinx sphinxsearch-prod, mf0-prod and mf0-stateless_.

As Apache act as a proxy for many services (mapserver, wsgi, tomcat) it should be configured to protect the backends.
The cluster _mf0-prod_ with its three memory hogs services is clearly underpowered.

Issue #1599, opened 2015-08-12
Profile.json issue


cedricmoullet:
503 error code on 2015-08-12T09:29:56.000Z for http://api3.geo.admin.ch/rest/services/profile.json

The following error can be seen in Apache Error log:

[Wed Aug 12 11:29:56 2015] [error] [client 10.220.6.61] (70008)Partial results are valid but processing is incomplete: mod_wsgi (pid=16403): Unable to get bucket brigade for request., referer: https://map.geo.admin.ch
[Wed Aug 12 11:29:56 2015] [error] [client 10.220.6.61] mod_wsgi (pid=7172): Exception occurred processing WSGI script '/var/www/vhosts/mf-chsdi3/private/chsdi/buildout/parts/modwsgi/wsgi'.
[Wed Aug 12 11:29:56 2015] [error] [client 10.220.6.61] Traceback (most recent call last):
[Wed Aug 12 11:29:56 2015] [error] [client 10.220.6.61]   File "/var/www/vhosts/mf-chsdi3/private/chsdi/buildout/eggs/pyramid-1.5.2-py2.7.egg/pyramid/router.py", line 242, in __call__
[Wed Aug 12 11:29:56 2015] [error] [client 10.220.6.61]     response = self.invoke_subrequest(request, use_tweens=True)
[Wed Aug 12 11:29:56 2015] [error] [client 10.220.6.61]   File "/var/www/vhosts/mf-chsdi3/private/chsdi/buildout/eggs/pyramid-1.5.2-py2.7.egg/pyramid/router.py", line 217, in invoke_subrequest
[Wed Aug 12 11:29:56 2015] [error] [client 10.220.6.61]     response = handle_request(request)
[Wed Aug 12 11:29:56 2015] [error] [client 10.220.6.61]   File "/var/www/vhosts/mf-chsdi3/private/chsdi/buildout/eggs/pyramid-1.5.2-py2.7.egg/pyramid/tweens.py", line 21, in excview_tween
[Wed Aug 12 11:29:56 2015] [error] [client 10.220.6.61]     response = handler(request)
[Wed Aug 12 11:29:56 2015] [error] [client 10.220.6.61]   File "/var/www/vhosts/mf-chsdi3/private/chsdi/buildout/eggs/pyramid_tm-0.12-py2.7.egg/pyramid_tm/__init__.py", line 99, in tm_tween
[...]

PR #1663, opened 2015-09-18
Deactivating not needed services on clusters

procrastinatio:
We have currently two mf-chsdi3 clusters:
* For API3, services
* For Mapproxy, profile/height and print

The goal is to deactivate services that do not need to run on a specific cluster (trying solve, or at least isolate #1599 and #1638)

Two new targets are defined `prod_api`and `prod_varia`, while `prod`is retained.

### DIFF
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -275,3 +275,12 @@ interpreted-options = app_version = __import__('time').strftime('%s')
                       apache_entry_path = '' if options.get('apache_base_path') == 'main' else ('/' + options.get('apache_base_path'))
 extends = vars
 
+
+[removeconfig]
+recipe = collective.recipe.cmd
+on_install = true
+on_update = true
+cmds =
+    if [ -f "${buildout:directory}/apache/tomcat-print.conf" ]; then rm "${buildout:directory}/apache/tomcat-print.conf"; fi &&
+    if [ -f "${buildout:directory}/apache/mapproxy.conf" ]; then rm "${buildout:directory}/apache/mapproxy.conf"; fi
+
--- a/buildout_prod.cfg
+++ b/buildout_prod.cfg
@@ -1,6 +1,7 @@
 [buildout]
 extends = buildout.cfg
 
+
 [vars]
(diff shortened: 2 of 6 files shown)


gjn:
I think now print runs on api cluster, not 'varia' cluster (varnish config)

gjn:
buildout_prod is used on jenkins (and elsewhere?). 
I would keep it and base the new ones on it.

procrastinatio:
@gjn varnish will redirect all print traffic on `prod_varia` (PR by @danduk82), but in the long term, we should use a special address, something like `print.geo.admin.ch`.

gjn:
ok, so we'll wait until it's active on varnish.

cedricmoullet:
@procrastinatio is varnish ready ? Still a WIP ?

procrastinatio:
All _print_ traffic should be redirected on the _mf0_prod_ cluster...

So now, we should be able to deactivate profile/height and print on _prod_stateless_ and carefully monitor mf0_prod_

gjn:
@procrastinatio Can you adapt to use `buildout_prod.cfg` as base for `buildout_api.cfg` and `buildout_varia.cfg` to avoid duplication of stuff (like variables, etc)?

gjn:
Thanks alot for the adaptions.

This is strange. When I build with _prod, then with prod_api, the apache configurations for print and mapproxy are not removed. Am I doing something wrong?

Alternatively, we could do it in the deploy hook...but with buildout, it would be better.

gjn:
ping @procrastinatio

procrastinatio:
Don't know. if I add `parts-=mapproxy print-war` to my config, both `parts`are effectively removed.

gjn:
Yes, but the apache configurations are still there...seems like the removeconfigs is not done?

gjn:
will be done with https://github.com/geoadmin/mf-chsdi3/pull/1823 and @procrastinatio creates rule in varnish to continue to support this address (chmobil, ign).

PR #1664, opened 2015-09-19
added date format d.m.Y

cype:


AFoletti:
Nothing against it. More so if I'm right about the source of this PR  ;)
But since the source is external to the team I'll let a developer merge it

cype:
:+1: Thanks Ambrogio & Cedric. More PRs to come from my side!

cedricmoullet:
@AFoletti please test the PR when you review it. This one generated issues so that I had to revert it.

cype:
Ups sorry, that's my fault. Since I don't have a development environment, I'm not able to test my code. Therefore you should be careful with my PR.

AFoletti:
I just checked the data and the problem is that the "end" attribute may contain null values.
@cype : I'll use your code and implement a not-null check, then post a new PR

cype:
Thanks Ambrogio!

PR #1667, opened 2015-09-21
Revert "added date format d.m.Y"

cedricmoullet:
Reverts geoadmin/mf-chsdi3#1664


cedricmoullet:
I revert this PR since it produces the following error:
AttributeError: 'NoneType' object has no attribute 'strip'

PR #1668, opened 2015-09-21
change date format in a BFE tooltip

AFoletti:
fixes #1664


procrastinatio:
If would rather use a `try...except...` block as everything may go wrong.

Why not using a  Mako_filter_ or _decorator_ (and store it in a central localtion) ?

AFoletti:
changed as suggested by @procrastinatio

PR #1682, opened 2015-09-25
Remove custom geoAlchmey egg

gjn:
This is functionally the same as the geoAlchemy fix we did in https://github.com/geoadmin/mf-chsdi3/pull/1286

It removes the local egg, updates GeoAlchmey from 0.2.4 to 0.2.5.

The fix is now applied by overwriting the changed function in our custom Geometry class.


loicgasser:
Cool thanks

Issue #1492, opened 2015-06-04
service public transportation Stationboard 


davidoesch:
Situation
with ch.bav.haltestellen-oev we will have public transportation points displayed on the map. The tooltip should contain (NTH) at least some information regarding the connections. We can't use SBB  service (since there is no official API and roting is not TU independent)  or transport.opendata.ch (not official)

Task
- [x] Arrival-/Departure times at stops, will be provided on data.geo.admin.ch by third party 
- [x] data will be delivered to data.geo.admin.ch/ch.bav.haltestellen-oev/ 
- [x] add delivered data to BGDI DB
- [ ] build a stationboard service with *same* parameter as in http://transport.opendata.ch/#stationboard more http://transport.opendata.ch/ 
- [x] add in the tooltip template of ch.bav.haltestellen-oev as in http://www.timeforcoffee.ch/ the next three connections

UI
- [x] Filter by
-- line number and a box around as in timeforcoffee (MUST v2)
-- direction 
-- departure time ( countdown !)

- [x] if departure info is missing: create a msgid display: no data available
- [ ] replace title "next departure "Your next departures from + "stationname"
- [x] countdown in minutes (not hours) and bold font as in timeforcoffee.ch (MUST v2)
- [ ] countdown if smaller than 5 min:bold font and red
- [ ] countdown if now :red symbols as in http://www.fahrplanfelder.ch/en/explanations/explanation-of-symbols.html#c160
- [x] content tooltip: reduce it to the next 4, add as well the station name as in the DI formular, 
- [x] add a link at the bottom to expert tooltip containing all the tooltip info according to the spec and the next 25 departures

Must for v1:
- [x]  Code review backend: gjin 
- [x] Code review backend: oteral
- [x] Dokumentation in api3.geo.admin.ch as "Service Stationboard" , based on http://transport.opendata.ch/#stationboard , incl a migration help as well
- [x] 	Ersatz Haltestellen TLM durch Haltestellen BAV in swisssearch -> mit Marcel

Must for v2:( update 16 december)
- [x] tooltip: change "Next departures from STATION" to " STATION :next departures to "
- [x] expert tooltip: put link to object into footer as in https://api3.geo.admin.ch/rest/services/swisstopo/MapServer/ch.swisstopo.lubis-luftbilder_schwarzweiss/19652280023398/extendedHtmlPopup?lang=en
[...]

PR #1684, opened 2015-09-25
Service Oev config

pfanguin:


pfanguin:
jenkins test

pfanguin:
Test link : https://mf-chsdi3.dev.bgdi.ch/fap_oev_service/transports/stops/8571395

chsdi/views/transports.py
```diff
@@ -0,0 +1,72 @@
...
+    @view_config(request_method='GET')
+    def get_departures(self):
+        current_date = datetime.datetime.now()
+        #next_thirty_minutes = current_date + datetime.timedelta(minutes=30)
+        query = self.request.db.query(Oev_departures).filter(and_(Oev_departures.stop == self.id, Oev_departures.time > current_date)).order_by(Oev_departures.time).limit(10)
+
+        def serialize(time):
+            return time.strftime('%Y/%m/%d %H:%M:%S')
+
+        def type_transports(type):
```
loicgasser: Simply use dictionary here I think.

chsdi/views/transports.py
```diff
@@ -0,0 +1,72 @@
...
+class TransportView(object):
+
+    def __init__(self, request):
+        self.request = request
+        if request.matched_route.name == 'transports':
+            self.id = request.matchdict['id']
+
+    @view_config(request_method='GET')
+    def get_departures(self):
+        current_date = datetime.datetime.now()
```
loicgasser: You have to take into account time zones. Here is a way how to do it:
http://www.saltycrane.com/blog/2009/05/converting-time-zones-datetime-objects-python/

Otherwise we're 2 hours behind..
>>> strftime("%a, %d %b %Y %H:%M:%S +0000", gmtime())
'Tue, 20 Oct 2015 07:45:55 +0000'

chsdi/views/transports.py
```diff
@@ -0,0 +1,72 @@
...
+from chsdi.models.vector.uvek import Oev_departures
+import datetime
+
+
+@view_defaults(renderer='jsonp', route_name='transports')
+class TransportView(object):
+
+    def __init__(self, request):
+        self.request = request
+        if request.matched_route.name == 'transports':
```
loicgasser: This condition is redundant with the route_name in the decorator above... So this will always be true.

chsdi/views/transports.py
```diff
@@ -0,0 +1,72 @@
...
+            self.id = request.matchdict['id']
+
+    @view_config(request_method='GET')
+    def get_departures(self):
+        current_date = datetime.datetime.now()
+        #next_thirty_minutes = current_date + datetime.timedelta(minutes=30)
+        query = self.request.db.query(Oev_departures).filter(and_(Oev_departures.stop == self.id, Oev_departures.time > current_date)).order_by(Oev_departures.time).limit(10)
```
loicgasser: It would be nice to have the limit returned results as a query parameter of the request. Same for transport types.
You can define default and upper values for these parameters.

chsdi/models/vector/uvek.py
```diff
@@ -10,6 +10,37 @@
 Base = bases['uvek']
 
 
+class Oev_departures (Base):
```
loicgasser: I would prefer OevDepartures for a class name

chsdi/views/transports.py
```diff
@@ -0,0 +1,72 @@
...
+            elif type == 14:
+                return '14'
+            elif type == 15:
+                return '15'
+            elif type == 16:
+                return '16'
+            elif type == 17:
+                return '17'
+
+        results = [{
```
loicgasser: Some error handling would be nice.

chsdi/models/vector/uvek.py
```diff
@@ -10,6 +10,37 @@
...
+    time = Column('time', DateTime)
+    label = Column('label', Text)
+    destination = Column('destination', Text)
+    type = Column('type', Integer)
+    via = Column('via', Text)
+
+register('ch.bav.departures-oev', Oev_departures)
+
+
+class Oev_haltestellen (Base, Vector):
```
loicgasser: OevHaltestellen

pfanguin:
@gjn and @loicgasser and @procrastinatio could you please do a review ? 
Thanks

chsdi/templates/htmlpopup/oev_haltestellen.mako
```diff
@@ -0,0 +1,151 @@
+<%inherit file="base.mako"/>
+
+<%def name="table_body(c, lang)">
+    <%!
+        from pyramid.url import route_url
```
gjn: Is this really needed? (route_url is method of request object)

chsdi/templates/htmlpopup/oev_haltestellen.mako
```diff
@@ -0,0 +1,151 @@
...
+    %>
+<style>
+  tr:last-child {
+    display:none;
+  }
+</style>
+<script>
+
+var timesRun = 0;
+var dest;
```
gjn: Try to avoid global variables.

chsdi/templates/htmlpopup/oev_haltestellen.mako
```diff
@@ -0,0 +1,151 @@
...
+      $("#timeDiff").html(timeDiff);
+    }); 
+}
+
+function select(val) {
+  if (window.refresh) {
+    clearInterval(window.refresh);
+  }
+  getInfos(val); 
+  window.refresh = setInterval(getInfos, 60000);
```
gjn: window.refresh could be a local variable here, then pass this variable to the getInfos function.

```
function select(val) {
   var refresh;
   ...
   ...
   refresh = setInterval(getInfos(refresh), 60000);
}
```
should be possilbe. Of course, adapt the getInfos function.

gjn: Also, you can pass in `val` to getInfos everytime here, so you don't need the gloable `dest` variable.

chsdi/templates/htmlpopup/oev_haltestellen.mako
```diff
@@ -0,0 +1,151 @@
...
+}
+
+$( document ).ready(function() {
+  $.getJSON( "https://mf-chsdi3.dev.bgdi.ch/fap_oev_service/stationboard/stops/${c['attributes']['nummer']}/destinations", function(result){
```
gjn: Hardcoded path is not good.

chsdi/views/stationboard.py
```diff
@@ -0,0 +1,76 @@
...
+        else:
+            if self.destination == 'all':
+                query2 = self.request.db.query(OevDepartures).filter(and_(OevDepartures.stop == self.id, OevDepartures.time > current_date)).order_by(OevDepartures.time).limit(5)
+            else:
+                query2 = self.request.db.query(OevDepartures).filter(and_(OevDepartures.stop == self.id, OevDepartures.destination == self.destination, OevDepartures.time > current_date)).order_by(OevDepartures.time).limit(5)
+
+            def serialize(time):
+                return time.strftime('%H:%M')
+
+            results2 = [{
```
gjn: You can use `results` here as well.

Alternatively, you can directly return the array instead of assigning it first to a variable.

chsdi/views/stationboard.py
```diff
@@ -0,0 +1,76 @@
...
+    def __init__(self, request):
+        self.request = request
+        if request.matched_route.name == 'stationboard_destination':
+            self.id = request.matchdict['id']
+
+    @view_config(request_method='GET')
+    def get_destination(self):
+        query = self.request.db.query(OevDepartures).filter(OevDepartures.stop == self.id).distinct(OevDepartures.destination).count()
+
+        if query == 0:
```
gjn: @loicgasser query == 0 good way to check for no results?

gjn:
I've reviewed and added a couple of comments. Some general points:

a) This needs rebase/squash
b) Please add integration tests for the views (100% coverage if possible)
c) Do we need official documentation for this service (ping @davidoesch )

chsdi/views/stationboard.py
loicgasser: instead use:
```python
from sqlalchemy.orm.exc import NoResultFound, OperationalError

query = self.request.db.query(OevDepartures).filter(
    OevDepartures.stop == self.id
).distinct(OevDepartures.destination)
try:
    results = [{
        'id': q.stop,
        'destination': q.destination
     } for q in query]
except NoResultFound:
    results = [{ 'destination': 'nodata' }]
except OperationalError, e:
    # Catches missing database and tables, columns etc...
    raise HTTPInternalServerError(e)


````

File: chsdi/views/stationboard.py [hunk 14]
```diff
@@ -0,0 +1,76 @@
...
+        self.request = request
+        if request.matched_route.name == 'stationboard':
+            self.id = request.matchdict['id']
+            self.destination = request.params.get('destination')
+
+    @view_config(request_method='GET')
+    def get_departures(self):
+        current_date = datetime.datetime.now()
+        if self.destination == 'all':
+            query = self.request.db.query(OevDepartures).filter(and_(OevDepartures.stop == self.id, OevDepartures.time > current_date)).order_by(OevDepartures.time).limit(5).count()
```
loicgasser: please replace this part with the solution I gave you hereunder.

chsdi/views/stationboard.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def __init__(self, request):
+        self.request = request
+        if request.matched_route.name == 'stationboard':
+            self.id = request.matchdict['id']
+            self.destination = request.params.get('destination')
+
+    @view_config(request_method='GET')
+    def get_departures(self):
+        current_date = datetime.datetime.now()
```
loicgasser: Please take into account time zone like I told you in a previous remark.

chsdi/views/stationboard.py
```diff
@@ -0,0 +1,76 @@
...
+        if query == 0:
+            results = [{
+                'destination': 'nodata'
+            }]
+            return results
+        else:
+            if self.destination == 'all':
+                query2 = self.request.db.query(OevDepartures).filter(and_(OevDepartures.stop == self.id, OevDepartures.time > current_date)).order_by(OevDepartures.time).limit(5)
+            else:
+                query2 = self.request.db.query(OevDepartures).filter(and_(OevDepartures.stop == self.id, OevDepartures.destination == self.destination, OevDepartures.time > current_date)).order_by(OevDepartures.time).limit(5)
```
loicgasser: Using less code for filters, you could do something like (untested):
```python
filters = [OevDepartures.stop == self.id, OevDepartures.time > current_date]
if self.destination == 'all':
    filters.append(OevDepartures.destination == self.destination)
query = self.request.db.query(OevDepartures).filter(and_(*filters))
````
And then same code as hereunder.

File: chsdi/views/stationboard.py [hunk 17]
```diff
@@ -0,0 +1,76 @@
...
+
+
+@view_defaults(renderer='jsonp', route_name='stationboard')
+class TransportView(object):
+
+    def __init__(self, request):
+        self.request = request
+        if request.matched_route.name == 'stationboard':
+            self.id = request.matchdict['id']
+            self.destination = request.params.get('destination')
```
loicgasser: Some parameter validation is required here.

chsdi/templates/htmlpopup/oev_haltestellen.mako
```diff
@@ -0,0 +1,151 @@
...
+    display:none;
+  }
+</style>
+<script>
+
+var timesRun = 0;
+var dest;
+function getInfos(val){
+    if (val) {
+      dest = val;
```
loicgasser: You need to handle the case where we have several tooltips opened at the same time.

chsdi/templates/htmlpopup/oev_haltestellen.mako
loicgasser: yes we have a variable `api_url` you can use from the settings.

chsdi/templates/htmlpopup/oev_haltestellen.mako
loicgasser: Also I would really rather have the JS part in geadmin. @gjn Too late for that now I guess, but I would really much rather do this properly and avoid polluting the global scope, load this script x times and create another piece of code we cannot test automatically. Enhancement issue?

gjn: Of course, the globals has to go.

Not sure if to put it into geoadmin is good as it creates stong coupling. the htmlpopup call might be used and inserted elsewhere (other sites) as well and in case of the extendedhtmlpopup needs to be standalone anyway.

That being said, I share your concern about testability; but there might be other ways to address that. I envision more feature rich popups in the future, and we need to find a good solution.

loicgasser: I was thinking that this would not be an html popup but rather a new component, I don't think we have JS in regular htmlpopups at the moment. Regular htmlpopup would be without JS with a link to the extendedpopup, but in geoadmin we'd handle things diffrently for this layer.
Probably involves some more thinking, but it would be a good start for front-end templating.
(In layersConfig maybe)

davidoesch:
Dokumentation in api3.geo.admin.ch as "Service Stationboard" , based on http://transport.opendata.ch/#stationboard , incl a migration help as well 
-> see as well to do list in https://github.com/geoadmin/mf-chsdi3/issues/1492

chsdi/templates/htmlpopup/oev_haltestellen.mako
pfanguin: yes, because I need to add the link for the extended tooltip and I add this manually, because I must change the style of the mako.

pfanguin:
I made several changes, @loicgasser, @gjn and @procrastinatio could you please make review ? Thanks

pfanguin:
Test link : https://mf-geoadmin3.dev.bgdi.ch/fap_oev/?topic=ech&lang=fr&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.bav.haltestellen-oev&layers_opacity=0.75

davidoesch:
wow I like: one further improvement: make the icons for zoom=0 and 1 and 2 even smaller. zoom 3 is perfect

loicgasser:
Code is good.
Use 'str'.isdigit() to test for the validity of the id instead of try except. you can return a message instead of the error. (please provide a valid id for instance)

TODO: doc, tests

Possible improvement:
Add a limit parameter to let the user choose the set of results he wants to retrieve. (Defaults to 5 and max 50 I would say)
I agree with david (partially), points still a little too big for those zoom levels. I think generalization would be better though.

loicgasser:
Also timediff could/should be calculated server side. It's nice to have it directly in the service and you can remove some code in the mako.

AFoletti:
that's a point layer. It should have 100% opacity per default, not 0.75 :)

gjn:
Todo: use the [brand new](https://github.com/geoadmin/mf-geoadmin3/pull/2863) `showTooltip=true` parameter in the link to object link. You probably need to rebase your mf-geoadmin3 branch for it to work correctly.

pfanguin:
See index sphinx : https://github.com/geoadmin/service-sphinxsearch/pull/243

chsdi/templates/htmlpopup/oev_haltestellen.mako
```diff
@@ -0,0 +1,174 @@
...
+  var dest;
+  var getInfos = function(val){
+    timesRun += 1;
+    if (val) {
+      dest = val;
+    }
+    if(timesRun === 20){
+      clearInterval(refresh);
+    }
+    console.log(timesRun);
```
loicgasser: oups

AFoletti:
A question about the legends: why are there "double" entries (like "tram,bus")? Should not a point only have one transport type?

chsdi/views/stationboard.py
```diff
@@ -0,0 +1,85 @@
...
+            try:
+                self.id = int(id)
+            except ValueError, e:
+                raise HTTPBadRequest(e)
+
+            self.destination = request.params.get('destination')
+
+    @view_config(request_method='GET')
+    def get_departures(self):
+        current_date = datetime.datetime.now(timezone('Europe/Berlin'))
```
procrastinatio: Why do we need to set the timezone ? No timezone is set at the DB level ?

loicgasser: This is the time zone of the server not (time at which the request was performed)

pfanguin:
@loicgasser : if you want the time difference in server side, I must remove the time zone of the current date, otherwise it does not work.

pfanguin:
Test link api doc : https://mf-chsdi3.dev.bgdi.ch/fap_oev_service/services/sdiservices.html#stationboard

davidoesch:
ilike !

chsdi/static/doc/source/services/sdiservices.rst
```diff
@@ -947,3 +947,90 @@ Example
...
+
+
+.. _owschecker_description:
+
+----------
+
+Stationboard
+------------
+
+This service provide the next connections from a location to a specific destination
```
loicgasser: This service `provides`

chsdi/static/doc/source/services/sdiservices.rst
```diff
@@ -947,3 +947,90 @@ Example
...
+Examples
+********
+
+- Get the fives next connection from a location to all destinations: `https://api3.geo.admin.ch/stationboard/stops/8501120?destination=all <../../../stationboard/stops/8501120?destination=all>`_
+- Get the fives next connection from a location to one destination: `https://api3.geo.admin.ch/stationboard/stops/8501120?destination=Genève <../../../stationboard/stops/8501120?destination=Gen%C3%A8ve>`_
+
+Response syntax
+***************
+
+Here is an example of response.
```
loicgasser: of `a`response or simply remove the whole sentence.

chsdi/static/doc/source/services/sdiservices.rst
```diff
@@ -947,3 +947,90 @@ Example
...
++-----------------------------------+-------------------------------------------------------------------------------------------+
+| Parameters                        | Description                                                                               |
++===================================+===========================================================================================+
+| **destination (required)**        | Specifies the arrival location of the connection                                          |
++-----------------------------------+-------------------------------------------------------------------------------------------+
+
+Examples
+********
+
+- Get the fives next connection from a location to all destinations: `https://api3.geo.admin.ch/stationboard/stops/8501120?destination=all <../../../stationboard/stops/8501120?destination=all>`_
```
loicgasser: `five` not `fives`

chsdi/static/doc/source/services/sdiservices.rst
```diff
@@ -947,3 +947,90 @@ Example
...
 - Check WMTS with Swiss ech-0056 profile (html): `https://api3.geo.admin.ch/owschecker/form?base_url=http%3A%2F%2Ftile1-sitn.ne.ch%2Fmapproxy%2Fservice&service=WMTS <../../../owschecker/form?base_url=http%3A%2F%2Ftile1-sitn.ne.ch%2Fmapproxy%2Fservice&service=WMTS>`_
+
+
+
+
+.. _owschecker_description:
+
+----------
+
+Stationboard
```
loicgasser: What about the other service?

chsdi/static/doc/source/services/sdiservices.rst
loicgasser: Maybe tell that it is based on timetables and not real time information. Also specify that it is used for public transportation.

chsdi/static/doc/source/services/sdiservices.rst
```diff
@@ -947,3 +947,90 @@ Example
...
+Response syntax
+***************
+
+Here is an example of response.
+
+.. code-block:: javascript
+
+    {
+        "via": "Lausanne; Renens VD; Morges; Allaman; Rolle; Gland; Nyon; Coppet",
+        "d": "20/11/2015 11:21",
```
loicgasser: Hard to understand what `d` means here. Can't we use a better name for this property?

chsdi/static/doc/source/services/sdiservices.rst
```diff
@@ -947,3 +947,90 @@ Example
...
+
+  https://api3.geo.admin.ch/stationboard/stops/{id}
+
+Input Parameters
+****************
+
++-----------------------------------+-------------------------------------------------------------------------------------------+
+| Parameters                        | Description                                                                               |
++===================================+===========================================================================================+
+| **destination (required)**        | Specifies the arrival location of the connection                                          |
```
loicgasser: Additionnaly there is a callback parameter if I recall correctly. You can also use a lang param, please add those.

loicgasser:
> if you want the time difference in server side, I must remove the time zone of the current date, otherwise it does not work.

It's probably ok for now (meaning no need to change it). If you want it to work you simply need to use timezone at the DB level as well.

chsdi/views/stationboard.py
```diff
@@ -0,0 +1,89 @@
...
+        ).limit(5)
+        try:
+            results = [{
+                'id': q.stop,
+                'time': serialize(q.time),
+                'label': q.label,
+                'destination': q.destination,
+                'via': q.via,
+                'currentDate': serializeD(current_date),
+                'd': serializeD(q.time)
```
loicgasser: sorry but I really don't like this `d` which is hard to understand.

chsdi/static/doc/source/services/sdiservices.rst
```diff
@@ -947,3 +947,93 @@ Example
...
+  https://api3.geo.admin.ch/stationboard/stops/{id} OR
+  https://api3.geo.admin.ch/stationboard/stops/{id}/destinations
+
+Input Parameters
+****************
+
++-----------------------------------+-------------------------------------------------------------------------------------------+
+| Parameters                        | Description                                                                               |
++===================================+===========================================================================================+
+| **destination (required)**        | Specifies the arrival location of the connection                                          |
```
loicgasser: Specifies the `destination` of the connection. Defaults to `all`

chsdi/static/doc/source/services/sdiservices.rst
```diff
@@ -947,3 +947,93 @@ Example
...
++-----------------------------------+-------------------------------------------------------------------------------------------+
+| **callback (optional)**           | The name of the callback function.                                                        |
++-----------------------------------+-------------------------------------------------------------------------------------------+
+
+Examples
+********
+
+- Get the five next connection from a location to all destinations: `https://api3.geo.admin.ch/stationboard/stops/8501120?destination=all <../../../stationboard/stops/8501120?destination=all>`_
```
loicgasser: `connections`

chsdi/static/doc/source/services/sdiservices.rst
```diff
@@ -947,3 +947,93 @@ Example
...
+- Get the five next connection from a location to one destination: `https://api3.geo.admin.ch/stationboard/stops/8501120?destination=Genève  <../../../stationboard/stops/8501120?destination=Gen%C3%A8ve>`_
```
loicgasser: `connections`

chsdi/static/doc/source/services/sdiservices.rst
```diff
@@ -947,3 +947,93 @@ Example
...
+- Get all destinations of a given location: `https://api3.geo.admin.ch/stationboard/stops/8507000/destinations <../../../stationboard/stops/8507000/destinations>`_
```
loicgasser: `from` a given connection.

loicgasser: sounds better to me.

chsdi/views/stationboard.py
```diff
@@ -0,0 +1,89 @@
...
+    def __init__(self, request):
+        self.request = request
+        if request.matched_route.name == 'stationboard':
+            id = request.matchdict['id']
+            if id.isdigit() is False:
+                raise HTTPBadRequest('Please provide a valid id for stationboard')
+            else:
+                self.id = id
+
+            self.destination = request.params.get('destination')
```
loicgasser: `self.destination = request.params.get('destination', 'all')`

chsdi/static/doc/source/services/sdiservices.rst
loicgasser: mmhh, I just noticed now, why do we use the id once and a string the next time?
`destination` parameter should be an id as well.... Can you adapt? You realize you're asking for an exact match string match?

loicgasser: We could support both though...

chsdi/models/vector/uvek.py
```diff
@@ -11,6 +11,43 @@
...
+class OevHaltestellen (Base, Vector):
+    __tablename__ = 'oev_haltestellen'
+    __table_args__ = ({'schema': 'bav', 'autoload': False})
+    __template__ = 'templates/htmlpopup/oev_haltestellen.mako'
+    __bodId__ = 'ch.bav.haltestellen-oev'
+    __label__ = 'name'
+    __extended_info__ = True
+    __queryable_attributes__ = ['nummer', 'name']
+    id = Column('bgdi_id', Text, primary_key=True)
+    nummer = Column('nummer', Integer)
```
pfanguin: nummer should be id

pfanguin:
jenkins test

loicgasser:
@pfanguin I added the tests, used a relationship (between both tables), fixed some doc and  squashed the whole thing. I think it's good to go.

PR #1687, opened 2015-09-29
test_profile, test_mapservice & test_search

mariavkourouni:


procrastinatio:
So now we have 92% coverage for profile and 93% for search.

I think you may squash all commits into one.but otherwise looks good to me.

loicgasser:
LGTM thanks a lot

PR #1695, opened 2015-10-01
[Admin KML] Fix empty files

pfanguin:
* Removing empty files
* Using _requests_ instead of _httlip2_

[Demo](https://mf-chsdi3.dev.bgdi.ch/fap_adminkml_fix/admin/kml)


loicgasser:
could you use request module instead? We trying to use only one module for that now. Sorry for not telling you earlier.

PR #1696, opened 2015-10-02
Improve test_helpers

mariavkourouni:


chsdi/tests/functional/test_helpers.py
```diff
@@ -4,13 +4,33 @@
...
+    def test_versioned(self):
+        import os.path
+        request = testing.DummyRequest()
+        request.host = 'api3.geo.admin.ch'
+        request.scheme = 'http'
+        request.registry.settings = {}
+        request.registry.settings['apache_base_path'] = 'main'
+
+        request.registry.settings['app_version'] = None
+        request.registry.settings['entry_path'] = 'https://api3.geo.admin.ch'
```
gjn: Entry path is something that comes after host, and not a fully qualified url.

chsdi/tests/functional/test_helpers.py
```diff
@@ -4,13 +4,33 @@
...
+        test_path = request.registry.settings['entry_path']
```
gjn: hardcode test_path here. I see 2 use cases:
1. simple root url (http://something.com)
2.url with path (https://something.com/entry_path/more/to)
3) the above without http:

loicgasser:
test_versioned updated.

PR #1697, opened 2015-10-05
Improve coverage test_qrcode

mariavkourouni:


loicgasser:
Ok
chsdi.views.qrcode_generator          35      4    89%
to
chsdi.views.qrcode_generator          35      2    94%

Shortener remains untested.

@mariavkourouni  Could you provide some tests for `chsdi.views.shortener`?
Otherwise LGTM.

PR #1699, opened 2015-10-07
Workaround to avoid dupplicate error between apache and wsgi

procrastinatio:
Mapproxy 1.8 is setting it's own CORS headers. Normally we would expected headers set by WSGI application to be available to apache, but it looks like it is broken in old mod_wsgi module.

So we have to found a way not to set the _Access-control-allow-origin_ for requests handled by Mapproxy.

Testing is not easy as about 30 vhosts are setting carefully this header.

### DIFF
--- a/apache/wsgi.conf.in
+++ b/apache/wsgi.conf.in
@@ -13,9 +13,14 @@ RewriteEngine on
 ExpiresActive on
 
 # Enabling CORS
-Header set Access-Control-Allow-Origin "*"
 Header always set Access-Control-Allow-Methods "POST, GET, OPTIONS"
 Header always set Access-Control-Allow-Headers "x-requested-with, Content-Type, origin, authorization, accept, client-security-token"
+
+# Mapproxy is adding its own CORS headers. 'Set' should be enought, but is not.
+<LocationMatch "^${apache_entry_path}/(?!1\.0\.0)">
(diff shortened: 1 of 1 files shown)


gjn:
What's the effect of the error now? We have the same header twice?

procrastinatio:
Accordent to Loïc, no tiles in 3D,  it probably matters
Le 8 oct. 2015 8:11 AM, "Gilbert Jeiziner" <notifications@github.com> a
écrit :

oterral:
yes @gjn, no WMTS tiles on 3d. Webgl doesn't like duplicate CORS headers

gjn:
That's awfully strange.

Can't we use `Header merge Access-Control-Allow-Origin "*"` on line 16 instead of new location match?

procrastinatio:
Not with merge, you end with "*,*". Normaly, *set* should do the trick, adding if not already set. Biut the problem, is that Apache, do not get the header list set by the WSGI application.

So the only solution, are:
* Do not set CORS for path under Mapproxy (with a look ahead regex)
* Only set CORS for specific part of chsdi (/rest, /static), but not the root path
* Deactivate wsgi.conf on the prod server.

gjn:
Ok, let's merge this then. Quite ugly, in fact. Maybe update of apache/mod_wsgi will help here as well?

I propose an emergency deploy for this.

PR #1700, opened 2015-10-08
Alternative way to set CORS

procrastinatio:


gjn:
Yes, seems better. Thanks alot.

Are you doing the emergency deploy?

PR #1701, opened 2015-10-12
Date-format dd.mm.yyyy

cype:
Same change as in Layer ch.bfe.abgeltung-wasserkraftnutzung: Date-format dd.mm.yyyy
I'm not able to test the code because I don't have a development environment.


loicgasser:
Ideally I'd rather have these columns formatted as dates and not text. Thank you for pointing that out. We'll change the type in DB and I'll open issue.

PR #1713, opened 2015-10-19
landesschwerenetz

alcapat:


AFoletti:
please store legends with transparent background
Otherwise LGTM

PR #1724, opened 2015-10-27
Data hydrologie messstationen gefahren

alcapat:


AFoletti:
I'll wait to put this one as "next deploy" ;). Abnahmeprotokoll is still pending

loicgasser:
For my knowledge what's the difference between ch.bafu.hydroweb-messstationen_gefahren and ch.bafu.hydrologie-messstationen_gefahren ?? Same layer?

AFoletti:
that's a question for @mariokeusen. Apparently BAFU requested the layer (see the [redmine ticket] (https://redmine.prod.bgdi.ch/issues/5921) ) but the info is exactly the same as in the hydroweb layer.

loicgasser:
Then we don't really need to create a nee symbology in CHSDI, just use the same and register the model with a new id as well, it will work as an alias. @alcapat is that why you closed this PR?

PR #1732, opened 2015-11-02
Wrong config for a tooltip visible limitation

alcapat:
This tooltip must be available only for level 11, 12 and 13


AFoletti:
that's OK for @mariokeusen ?

alcapat:
Yes, you can see it in the redmine ticket : https://redmine.prod.bgdi.ch/issues/5892

Issue #1721, opened 2015-10-27
Corrupt MuliPDF


cedricmoullet:
URL: https://map.geo.admin.ch/?topic=ech&lang=fr&bgLayer=voidLayer&layers=ch.swisstopo.zeitreihen&layers_timestamp=18641231&X=240463.40&Y=638094.97&zoom=7&time=1864
Print settings:
![image](https://cloud.githubusercontent.com/assets/319676/10750426/84055d22-7c75-11e5-8178-35bd0ed44262.png)

This configuration generates a corrupted PDF

cedricmoullet:
Same issue with https://map.geo.admin.ch/?topic=ech&lang=fr&bgLayer=voidLayer&layers=ch.swisstopo.zeitreihen&layers_timestamp=18611231&X=240710.00&Y=636970.00&zoom=6&time=1861
and 
![image](https://cloud.githubusercontent.com/assets/319676/10750513/8b42e6a8-7c76-11e5-81e1-3fc9f35d5e6c.png)

oterral:
The 2 links works by me. May be it was a problem during downloading.

gjn:
I've had this sometimes as well when using 'show' instead of 'download' in IE11. Somehow, IE11 attempts to show the file before it's done downloading.

Anyhow, I was able to reproduce it in chrome. The error occurse during the download (it finishes downloading the file but reports an error during download). The file on the infrastructure exists.

Still investigating.

procrastinatio:
No, the PDF is fine in all case. It is the way IE is handling the PDF. To reproduce:
1. Close IE
2. Print a big Multi-PDF and choose 'open'. Adobe complains the PDF is corrupt.

However, if you first choose 'Save PDF', it works fine. After this, even 'Open' works... It also always works if you succeed opening any PDF. Only trying to open a big PDF from start geht nicht.

procrastinatio:
Very hard to reproduce on a test plateform:
* On dev, never happed
* On int (cluster of two), also never happed
* Only big PDF are affected (>100MB)
* Generated PDF itself is perfectly fine.
* From time to time on prod. I a not even 100% convinced it is a server related issue, as if you manage to open one time a PDF, it never happed again...

PR #1735, opened 2015-11-04
Pdf as attachment for IE greater than 10

procrastinatio:
Adding IE 11 (and greater) to the IE list that should handling PDF as PDF attachment.

Should fix https://github.com/geoadmin/mf-chsdi3/issues/1721

### DIFF
--- a/buildout_ltmom.cfg
+++ b/buildout_ltmom.cfg
@@ -1,5 +1,6 @@
 [buildout]
 extends = buildout_user_base.cfg
+#parts += print-war
 
 [vars]
 apache_base_path = ltmom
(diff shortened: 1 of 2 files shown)


gjn:
That's a nice find!

Please merge once mom configuration gets rid of testing war.

buildout_ltmom.cfg
```diff
@@ -1,5 +1,6 @@
 [buildout]
 extends = buildout_user_base.cfg
+parts += print-war
```
gjn: @procrastinatio Please remove this.

buildout_ltmom.cfg
```diff
@@ -1,5 +1,6 @@
 [buildout]
 extends = buildout_user_base.cfg
+#parts += print-war
```
loicgasser: Please adapt

procrastinatio:
Why ? It's *my* config.

gjn:
we wanted to avoid user specific print instances. but if you want one in your env., I'm fine with it. Please merge.

PR #1741, opened 2015-11-06
GetCapabilities open for everybody. Adapt test

gjn:
We adapted the varnish configuratoin to free up getCapabilites document for everybody.

Therefore, we need to adapt our varnish tests. This PR does it.

Self-merge.


gjn:
See https://github.com/geoadmin/mf-geoadmin3/issues/2772

PR #1748, opened 2015-11-09
update date format for energiestaedte

rfrancel:


AFoletti:
LGTM, once the tests pass. But it's not this branch's fault in mz opinion ;)

pfanguin:
jenkins test

loicgasser:
Can we have just a function for that please?
Did you change the column type?
You shouldn't have to cast it to a date. Taht s the advantage of using the correct column type.

AFoletti:
the column type change is initialized, but it will not be done in 5 minutes

loicgasser:
Can you use one function from helpers instead?
Please stop duplicating code everywhere.

PR #1749, opened 2015-11-09
Increase coverage chsdi.views.height

mariavkourouni:


mariavkourouni:
jenkins test

mariavkourouni:
jenkins test

pfanguin:
jenkins test

loicgasser:
That's a goos test, but I am not sure the height service should return simply a 200 code and None if the requested point is out of Switzerland.
We should change that.

loicgasser:
@mariavkourouni can you make a proposal?

mariavkourouni:
Do you mean that in this case the height service should return a bad request?

loicgasser:
Yes either bad request, either a JSON like this:
```JSON
{
  "success": false,
  "error": {
    "code": 400,
    "message": "Requested coordinate out of bounds"
  }
}
```
@gjn @procrastinatio what do yo prefer?

gjn:
I think we should return a 400 and put `Requested coordinate out of bounds` as reason. We should not return a 200.

loicgasser:
Agree, @mariavkourouni could you change that in the service and adapt the test accordingly?

mariavkourouni:
adapted!

loicgasser:
Thanks a lot

PR #1751, opened 2015-11-10
update mapproxy configuration

ltclm:


procrastinatio:
I will update again tomorrow and merging, just before deploying.

PR #1754, opened 2015-11-12
fix url format to chmobil application

ltclm:
The url to the chmobil application had to be changed.
This change has to be deployed together with uvek_master deploy.


gjn:
Emergency deploy?

ltclm:
as discussed with @mariokeusen it is sufficient to publish this change with the next planned deploy.

loicgasser:
LGMT

PR #1759, opened 2015-11-12
Factorize mako and use ogcproxy to load image under https

loicgasser:


AFoletti:
I changed the remaining makos. Thanks for your inputs

loicgasser:
Top merci!

PR #1771, opened 2015-11-23
Increase coverage chsdi.views.feedback

mariavkourouni:


loicgasser:
This test may send/spam a lot of emails. We should indeed test it, but maybe not like this. Break down the service, test functions...

gjn:
It's a no-go to send testing emails automatically to the official support address.

PR #1781, opened 2015-11-25
[WIP] Add 3config support in layers configuration service

loicgasser:
@gjn @ltclm (I am using a tem DB and table)
Explicit behaviour:
A 3d config parameter is defined -> use the config for the imagery layer as is (don't extend 2D config)
(Issue with subLayers where we would need to use 2 layers proxy for one layer in WGS84)

Implicit behaviour (nothing in layersconfig):
For WMTS: use MapProxy
For WMS: use CRS=EPSG:4326

Note that I am using the final names and timestamps getting rid of the `.3d` in the `serverLayerName`.

As discussed `opaque` and `has3dMetadata` will remain hardcoded in the mf-geoadmin3 as it will only affect a set of predefined layers. We'll get rid of `extent`.

So you should 4 new layers (which are only proxies similarly to what we have with `_wmts` and `_wms`:
- ch.swisstopo.pixelkarte-farbe_3d (serverLayerName: ch.swisstopo.pixelkarte-farbe)
- ch.swisstopo.pixelkarte-grau_3d (serverLayerName: ch.swisstopo.pixelkarte-grau)
- ch.swisstopo.zeitreihen_3d (serverLayerName: ch.swisstopo.zeitreihen)
- ch.swisstopo.swissimage-product_3d ...

I will copy all the image tiles to the correct address once this has been validated and adapt the code in mf-geoadmin3.

Changes in DB:
In tileset we have now new `tile_matrix_set_id` (e.g. 4326) Note that I did not add all the timestamps for zeitreiheihen. (for testing purposes, in general we want to be able to have a set of different timestamps per `CRS`.

In layers config table: I added 2 new columns
`config3d`
`srid` -> in order to group the layers in tileset using the projection.

Of course I needed to adapt the layers config view where all the magic happens.
Add new entries in tileset, dataset and layers_js.

[Test Link](http://mf-chsdi3.dev.bgdi.ch/gal_3d_layersconf/rest/services/ech/MapServer/layersConfig)

### DIFF
--- a/chsdi/models/bod.py
+++ b/chsdi/models/bod.py
@@ -86,29 +86,29 @@ class LayersConfig(Base):
     geojsonUrlit = Column('geojson_url_it', Text)
     geojsonUrlrm = Column('geojson_url_rm', Text)
     geojsonUrlen = Column('geojson_url_en', Text)
+    config3d = Column('fk_config3d', Boolean)
+    srid = Column('srid', Text)
 
     def layerConfig(self, params):
         config = {}
         translate = params.translate
-        excludedAttrs = ('geojsonUrlde', 'geojsonUrlfr', 'geojsonUrlit', 'geojsonUrlrm', 'geojsonUrlen')
         wmsHost = params.request.registry.settings['wmshost']
         for k in self.__dict__.keys():
-            if not k.startswith("_") and k not in excludedAttrs and \
-                self.__dict__[k] is not None and \
-                    k not in ('staging'):
+            val = self.__dict__[k]
+            if not k.startswith("_") and not k.startswith('geojsonUrl') and \
+                    val is not None and k not in ('staging', 'srid'):
                 if k == 'maps':
-                    config['topics'] = self.__dict__[k]
+                    config['topics'] = val
                 elif k == 'layerBodId':
-                    config['label'] = translate(self.__dict__[k])
+                    config['label'] = translate(val)
                 elif k == 'attribution':
-                    config[k] = translate(self.__dict__[k])
-                elif k == 'matrixSet':
-                    if self.__dict__[k] != '21781_26':
-                        config['resolutions'] = self._getResolutionsFromMatrixSet(
-                            self.__dict__[k]
-                        )
+                    config[k] = translate(val)
+                elif k == 'matrixSet' and self.__dict__['srid'] != '4326':
+                    config['resolutions'] = self._getResolutionsFromMatrixSet(
+                        val
+                    )
                 else:
-                    config[k] = self.__dict__[k]
+                    config[k] = val
 
         layerStaging = self.__dict__['staging']
         if config['type'] == 'wmts':
--- a/chsdi/static/doc/source/api/faq/index.rst
+++ b/chsdi/static/doc/source/api/faq/index.rst
@@ -81,6 +81,7 @@ Here is a list of all the freely accessible layers:
    <script type="text/javascript">
 
    function init() {
+       var suffix3d = '_3d';
        $.getJSON( "../../rest/services/api/MapServer/layersConfig", function( data ) {
           var myInnerHtml_queryable = "<br><table border=\"0\">";
           var myInnerHtml_searchable =  "<br><table border=\"0\">";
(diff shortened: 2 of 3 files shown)


ltclm:
thank you @loicgasser for the work. we can have a look at it this friday,

just some small remarks

* ch.swisstopo.swisstlm3d-karte-grau.3d
* ch.swisstopo.swisstlm3d-karte-farbe.3d

will be integrated and registered in the bod, so the ``.3d`` will stay in the layername and server layername. they will be the 3d version of ch.swisstopo.pixelkarte-farbe|grau
so it should be like that i guess:
* ch.swisstopo.pixelkarte-farbe_3d (serverLayerName: ch.swisstopo.swisstlm3d-karte-farbe.3d)
* ch.swisstopo.pixelkarte-grau_3d (serverLayerName: ch.swisstopo.swisstlm3d-karte-grau.3d)


> I will copy all the image tiles to the correct address once this has been validated and adapt the code in mf-geoadmin3.

what do you mean with this?

>  srid -> in order to group the layers in tileset using the projection.

Not necessary i think, this information is already in the tileset table which should contain the information about all native tiles for each layer.

We will have to be sure that the following views and the chsdi services which are using these views are supporting multiple matrix_ids, srid's, format, and timestamps for wmts layers:
* re3.view_layers_js
* re3.view_bod_wmts_getcapabilities_**

loicgasser:
> will be integrated and registered in the bod, so the .3d will stay in the layername and server layername.

I don't think the name should change. At the end of the day we're only changing the srid. I personnally think it's better as we can use this layer in another context than 3d. I think that's what @gjin wanted as well if I am not mistaken.

> what do you mean with this?

I mean that in order to link the JS configuration to the tileset table, we need to know the srid. Otherwise we'll mix timestamps based on 4326 and 21781. Absolutely necessary, hard to explain without the DB schema here. ;)

>We will have to be sure that the following views and the chsdi services which are using these views are supporting multiple matrix_ids, srid's, format, and timestamps for wmts layers

Absolutely

loicgasser:
We'll mix timestamps unless we change the name of layer... Again changing the srid and not the layername in the bucket path is better.

loicgasser:
ch.swisstopo.swisstlm3d-karte-grau.3d

So maybe we can say that the .3d above means that the layer is not exactly the same (becaue of terrain shading in the layer).
But that it's the idea.

For ch.swisstopo.swissimage-product the layer doesn't change so the name shouldn t change. Same for zeitreihen.

loicgasser:
So I agree with you in the end.

.3d means that a layer has been generated for 3d purposes only (basically a new layer)

_3d means that we are dealing with a counter part of a 2d layer (not a realy layer)

srid changes -> not a new layer, just a new projection and nothing else changes.

gjn:
+1 from me.

as you said, the tlm 3d data is different from the 3d one as we integrated the shading).

Principle: if it's same data, name should be the same (and only attributes as srid, etc should change).

gjn:
@loicgasser @ltclm would it be easy to create a diff between bod_master and the bod_ltgal_master.

ltclm:
we could dump data only and ddl only and make a diff of both versions.
I propose to make a ddl diff and apply the structural changes made in the ltgal version to the master database.

gjn:
could pg_dumpall help?

ltclm:
there is no difference between pg_dumpall and pg_dump, the first one is dumping the whole cluster with all db's.  We have to compare the new structure of the ltgal version with the master and manually merge the changes.

gjn:
Could we try to do this together tomorrow?

loicgasser:
DB changes are commited in dev.
To remember change staging from test to prod before deploy.

gjn:
jenkins test

gjn:
Is there a summaray about the changes that were made in the bod?

gjn:
This test fails: https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/tests/integration/test_wmtscapabilities.py#L30

loicgasser:
Yes, we just changed the DB and I squashed those commits it should be good to merge after Jekins Gjinkins agree.

gjn:
Good work. Thanks alot.

Let's merge this and fix forward upcoming issues. We're in this now.

PR #1787, opened 2015-12-01
Allow layer search test with 0 features

gjn:
Currently, when a layer has the `searchable` attribute set to `true` in the db, there has to be a sphinx index prepared for this layers.

But we can have such layers [1], thus leading in a failure of our automated tests. This PR adresses this issue by explicitely checking if a given searchable layer has features. If it has, the test remains as before. If it doesn't have any features, we test for 404 failure and explicitely ask if the service gave back the correct error.

[1] There are layers that need to be deployed without any actual data ready when deploying. Data will be added later on with automated data imports. Case in point are sachplan layers that become data-filled only after decisions in political contexts.

@AFoletti I think this is what we want
@loicgasser quick review/merge appreciated, as it's needed for tomorrows deploy

### DIFF
--- a/chsdi/tests/integration/test_layers.py
+++ b/chsdi/tests/integration/test_layers.py
@@ -86,10 +86,21 @@ class LayersChecker(object):
             assert ((layer + '_' + lang) in legendImages), layer + '_' + lang
 
     def checkSearch(self, layer):
+        models = models_from_name(layer)
+        assert (models is not None and len(models) > 0), layer
+        model = models[0]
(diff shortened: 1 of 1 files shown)


AFoletti:
great! many thanks!

AFoletti:
Just as a reminder: I'm setting the "staging" for a layer with a sphinx index and empty tables to PROD now. This will result in an error if this PR is not merged

loicgasser:
LGTM

PR #1790, opened 2015-12-02
Treat distributed indices in sphinx search tests

gjn:
Latests failures on jenkins jobs are cause by a layer which consists of multiple models. This case is covered by a distributed index in sphinxsearch.

There's a difference in how sphinxsearch treats indices with no content (no data). For single indices, the index is not created at all (even when defined in the config) and completely ignored. For distributed indices, the sub-indices are always created, even though neither of the contains any data. From the point of view of the service this means that a) single indices with no data will return 404 (as no sphinx index is found) but b) distributed indices with no data will return 200, even though there's no data.

This PR tries to address this by checking for 404 only for single indices.

It's probably not the optimal solution, but I think follows the KISS rule for now.

### DIFF
--- a/chsdi/tests/integration/test_layers.py
+++ b/chsdi/tests/integration/test_layers.py
@@ -89,11 +89,13 @@ class LayersChecker(object):
         models = models_from_name(layer)
         assert (models is not None and len(models) > 0), layer
         model = models[0]
-        query = self.session.query(model.primary_key_column())
         expectedStatus = 200
-        # we expect 404 errors for searchable layers without any data
-        if query.first() is None:
-            expectedStatus = 404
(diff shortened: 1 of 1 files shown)


gjn:
jenkins test

gjn:
@loicgasser quick review (so we can fix failing jenkins tests) would be appreciated.

loicgasser:
Yes, so we really only test the non-distributed indices, as distributed indices always return 200.
I think that in the DB depoy/sync script we check for the physical existence of sphinx indices on sphinx servers and throw a warning in case they are missing.
@ltclm Am I right?

I agree with you, let's test what we can at the moment.

PR #1789, opened 2015-12-02
Update link to geoportal FR

procrastinatio:
Update link accoridng to email from P.Terrettaz. Should fix https://github.com/geoadmin/mf-chsdi3/issues/1730

[Démo](https://mf-chsdi3.dev.bgdi.ch/mom_fr_url/rest/services/ech/MapServer/ch.kantone.cadastralwebmap-farbe/17/htmlPopup?coord=578763,183672&imageDisplay=1920,514,96&lang=fr&mapExtent=577745.6408718082,183460.49689954793,579665.6408718082,183974.49689954793)

PR #1791, opened 2015-12-03
Replace httplib with requests in link tests

gjn:
Because https://github.com/geoadmin/mf-chsdi3/pull/1789/files doesn't work with urllib2, so we replace it with requests, and suddenly it works.

@procrastinatio Please review/merge asap as it fixes broken test in master.

### DIFF
--- a/chsdi/tests/e2e/test_links.py
+++ b/chsdi/tests/e2e/test_links.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-import httplib2
+import requests
 
 from chsdi.tests.integration import TestsBase
 
@@ -8,12 +8,14 @@ from chsdi.tests.integration import TestsBase
 class TestLinks(TestsBase):
 
     def test_external_links(self):
-        h = httplib2.Http(timeout=10)
+        headers = {
+            'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/37.0.2062.68 Safari/537.36'
+        }
         for i in range(23):
             response = self.testapp.get('/rest/services/ech/MapServer/ch.kantone.cadastralwebmap-farbe/%d/htmlPopup' % i, status=200)
 
             soup = response.html
             for a in soup.findAll('a'):
                 link = a.get('href')
-                resp, content = h.request(link, headers={'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/37.0.2062.68 Safari/537.36'})
-                self.assertTrue(resp.status == 200, link)
+                r = requests.get(link, timeout=10, headers=headers)
+                self.assertTrue(r.status_code == 200, link)


fredj:
"so we replace it with requests, and suddenly it works."  :+1:

loicgasser:
ok... very comforting to know that httplib2 is that reliable. While you're at it, don't you think a head request is enough?

gjn:
It'd like to simulate a click on the link, which is a GET request.

loicgasser:
An other error at 10:02 we had today is a urllib3 connection pool error coming from `test_wms.py`

```requests.packages.urllib3.connectionpool: INFO: Starting new HTTPS connection (1): wms.geo.admin.ch
requests.packages.urllib3.connectionpool: DEBUG: "GET /?lang=de&EXCEPTION=application%2Fvnd.ogc.se_xm&SERVICE=WMS&FORMAT=image%2Fpng&REQUEST=GetMap&HEIGHT=1000&WIDTH=1000&VERSION=1.1.1&BBOX=600000%2C200000%2C610000%2C210000&LAYERS=ch.bav.sachplan-infrastruktur-schiene_ausgangslage&SRS=EPSG%3A21781 HTTP/1.1" 200 784```

File: chsdi/tests/e2e/test_links.py [hunk 1]
```diff
@@ -8,12 +8,14 @@
...
         for i in range(23):
             response = self.testapp.get('/rest/services/ech/MapServer/ch.kantone.cadastralwebmap-farbe/%d/htmlPopup' % i, status=200)
 
             soup = response.html
             for a in soup.findAll('a'):
                 link = a.get('href')
-                resp, content = h.request(link, headers={'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/37.0.2062.68 Safari/537.36'})
-                self.assertTrue(resp.status == 200, link)
+                r = requests.get(link, timeout=10, headers=headers)
+                self.assertTrue(r.status_code == 200, link)
```
loicgasser: simply `self.assertTrue(r.status_code == 200)`

gjn: The `link` part will output the link that failed on failure. It's for readability in case of failure.

loicgasser: okay

loicgasser:
@gjn Reading the doc I see that https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertTrue
```This method should also be avoided when more specific methods are available (e.g. assertEqual(a, b) instead of assertTrue(a == b)), because they provide a better error message in case of failure.```

gjn:
Yes, assertEqual should be used in such cases.

loicgasser:
I'll finish the cleanup urllib3/httplib2 cleanup and update that at the same time.

gjn:
Very funny, no it doesn't work again...using this code.

loicgasser:
Yes unstable page on their side.

procrastinatio:
Why do we have to check external links ? And only those of the layer 'cadastralwebmap' ? Our tests may fail because of something out oof our control. Scary.

If you use `request`, also please use `retry`, you may not expect a server answer the first time.

gjn:
Valid question. It's convenience test. That's why it's in separate jenkins job altogether. It's not executed on all other jobs.

I think, we are responsible that links visible on our geoportal to be correct and that they don't point on dead-ends. But this can certainly be discussed.

Anyhow, the reason it sill fails because jenkins was using old testlinks code under integration directory, but the test is now under e2e directory. We didn't clean out the workspace, that's why it was still working. (maybe traling test_links.pyc file on jenkins).

I reconfigured jenkins test and it should be good now (note: this PR was still needed).

PR #1792, opened 2015-12-03
Use general requests interface

loicgasser:
A reason why our `test_wms.py` fails is that we don't take into account the context.
http://docs.python-requests.org/en/v1.2.3/api/

### DIFF
--- a/chsdi/tests/e2e/test_wms.py
+++ b/chsdi/tests/e2e/test_wms.py
@@ -14,7 +14,11 @@ try:
     api_url = "http:" + app.registry.settings['api_url']
     base_url = "http://" + app.registry.settings['wmshost'] + '/'
     geodata_staging = app.registry.settings['geodata_staging']
+    max_retry = 1
+    if geodata_staging == 'dev':
+        max_retry = 3
 except KeyError as e:
+    geodata_staging = 'prod'
     base_url = 'http://wms.geo.admin.ch'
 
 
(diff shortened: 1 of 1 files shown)


loicgasser:
@gjn @procrastinatio Quick review merge?

gjn:
Tests failed because of wrong info in the bod (at least recently).

I don't fully understand what this PR adds.

loicgasser:
It's essentially to use the retry option.

procrastinatio:
Nice, for me. We already used `retry`for other tests. Dic you tests also against self-signed certificat we have ?

gjn:
Doesn't having to re-try already say that something is fishy?

loicgasser:
A lot of things can go wrong when performing HTTP requests, networking issue, WMS server being restarted, DB being restarted, Puppet, timeouts ... you got the picture. I don't think all the tests should fail if a request failed one time on a server. This is time consuming to analayse redeploy etcc... So a retry of 2 or 3 should be ok.

loicgasser:
For external resources only of course.

gjn:
I agree for external resources (not managed by us). I also agree on dev resources.

But for our integration and production resources, we should always be up. We should not accept downtimes there, even if they are mini downtimes.

And here we are testing wms production, no?

loicgasser:
these tests are env dependent. dev tests WMS dev etcc.

gjn:
So I would propose to not use retry for int/prod and to use retry=3 for dev.

loicgasser:
ok done

gjn:
Thanks alot!

PR #1804, opened 2015-12-09
Adapt mapproxy tests to new autoscale backends

gjn:
This adapts the test urls for mapproxy related tests. It also assures that the wmts getcapabilities contain the correct urls for the re-projected tiles (using the wmts20.[dev|int].bgdi.ch addresses for dev and int and wmts20.geo.admin.ch for prod.

### DIFF
--- a/chsdi/tests/e2e/test_wmtsgettile.py
+++ b/chsdi/tests/e2e/test_wmtsgettile.py
@@ -21,6 +21,11 @@ MAPPROXY_URLS = [
     'wmts12.geo.admin.ch',
     'wmts13.geo.admin.ch',
     'wmts14.geo.admin.ch',
+    'wmts20.geo.admin.ch',
+    'wmts21.geo.admin.ch',
+    'wmts22.geo.admin.ch',
+    'wmts23.geo.admin.ch',
+    'wmts24.geo.admin.ch',
 ]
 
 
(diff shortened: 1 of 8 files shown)


gjn:
@procrastinatio as we just discussed. Would be great to have this for tomorrow.

chsdi/tests/e2e/test_wmtsgettile.py
```diff
@@ -16,11 +16,11 @@
...
-    'wmts10.geo.admin.ch',
-    'wmts11.geo.admin.ch',
-    'wmts12.geo.admin.ch',
-    'wmts13.geo.admin.ch',
-    'wmts14.geo.admin.ch',
+    'wmts20.geo.admin.ch',
+    'wmts21.geo.admin.ch',
+    'wmts22.geo.admin.ch',
+    'wmts23.geo.admin.ch',
+    'wmts24.geo.admin.ch',
```
gjn: I will also add 10-14 again.

procrastinatio:
OK, I thinks it's good, but how do we reconcilie discrepancies between the `GetCapabilities`on `dev` using the BOD and `mapproxy` autoscaling normally `layersConfig`from `prod`. We have to make sure that both are in sync.

gjn:
Yes, that's something we need to take care of. Let's discuss it Friday.

PR #1816, opened 2015-12-11
Add test for config3d and better error msg in case of failure

loicgasser:
Add test for config3d and better error msg in case of failure
This PR also remove the property singleTile from geojson and aggregate layers as they are not use for those.

### DIFF
--- a/chsdi/tests/integration/test_catalog.py
+++ b/chsdi/tests/integration/test_catalog.py
@@ -15,7 +15,7 @@ class TestCatalogService(TestsBase):
 
     def test_catalog_with_callback(self):
         resp = self.testapp.get('/rest/services/blw/CatalogServer', params={'callback': 'cb'}, status=200)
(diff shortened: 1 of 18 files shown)


chsdi/tests/integration/test_catalog.py
```diff
@@ -33,7 +33,7 @@ def test_catalog_languages(self):
         for lang in ('de', 'fr', 'it', 'rm', 'en'):
             link = '/rest/services/ech/CatalogServer?lang=' + lang
             resp = self.testapp.get(link)
-            self.assertTrue(resp.status_int == 200, link)
+            self.assertEqual(resp.status_int, 200)
```
gjn: Please re-add the link parameter (more info in case of failure)

loicgasser: ok done

gjn:
Great. Thanks alot.

PR #1822, opened 2015-12-17
Increase connection and request timeout for e2e tests

procrastinatio:
`cadastralwepmap`has the following path , so be patient.
WMS Infogrips (Quelle, Interlis)  --->  WMS bgdi   --->  Mapproxy


gjn:
Thanks alot for this.

Issue #1338, opened 2015-03-20
Get rid of buildout and use Makefile


loicgasser:
Like in c2cgeoportal I think they made the right decision.
It is older but way more stable.

gjn:
See https://github.com/camptocamp/c2cgeoportal

loicgasser:
I would put a higher prio to this issue. to be discussed.

gjn:
Ding ding.

loicgasser:
probably as happy your are.

PR #1823, opened 2015-12-17
Replace Buildout with Makefile

loicgasser:
I started from https://github.com/ioda-net/mf-chsdi3/tree/Makefile Thanks @Jenselme!
Related issue: https://github.com/geoadmin/mf-chsdi3/issues/1338
This PR also removes MapProxy

You can already use it in dev mode:

```
cp rc_example rc_user_xxx
```

Edit `rc_user_xxx`

```
source rc_user_xxx
make all
make serve
```

Open `localhost:${USER_PORT}`

TODO:

- [x] Test WSGI app
- [x] API build
- [x] Less build
- [x] Add test target
- [x] Print config
- [x] Adapt doc
- [x] Translation
- [x] Adapt all scripts
- [x] Finish cleanup
- [x] Add no tests option to deploysnapshot
- [x] ~~Update Varnish config for IGN MapProxy~~
- [x] ~~Add missing package on int and prod (no longer a requirement)~~

Before merge:
- [ ] Adapt Jenkins
- [ ] Re-enable prod target
- [ ] Remove deploydev temp fix in Makefile
- [ ] Triple check everything


loicgasser:
@oterral @procrastinatio @gjn a pre-review would be good.
[Test Link](http://mf-chsdi3.dev.bgdi.ch/gal_makefile/) (Deployed with `make deploybranch`)

Once your config file has been created you can simply do `make user`
or `source rc_user_xxx` and `make user`

Makefile
```diff
@@ -0,0 +1,368 @@
...
+	git remote add upstream https://github.com/openlayers/ol3 && \
+	git fetch upstream && \
+	npm install && \
+	API_URL=$(API_URL) make -f Makefile-ga build-ga && \
+	make -f Makefile-ga build-ga $(CURRENT_DIRECTORY)/chsdi/static/css/ && \
+	cp build/ga.css $(CURRENT_DIRECTORY)/chsdi/static/css/ && \
+	cp build/ga*.js $(CURRENT_DIRECTORY)/chsdi/static/js/ && \
+	cp resources/EPSG* $(CURRENT_DIRECTORY)/chsdi/static/js/
+
+.PHONY: deploybanch
```
oterral: deploybranch

Makefile
```diff
@@ -0,0 +1,368 @@
...
+	jar uf $(CURRENT_DIRECTORY)/temp_$(APP_VERSION)/$(BASEWAR) $(PRINT_INPUT)
+	@echo "${GREEN}Print war update successful creation has been successful."
+	@echo "Copying jar file from $(CURRENT_DIRECTORY)/temp_$(APP_VERSION)/...${RESET}"
+	cp -f $(CURRENT_DIRECTORY)/temp_$(APP_VERSION)/$(BASEWAR) $(PRINT_OUTPUT)
+	@echo "${GREEN}Removing temp directory${RESET}"
+	rm -rf $(CURRENT_DIRECTORY)/temp_$(APP_VERSION)
+
+# FIXME default target is always master
+# Remove when ready to be merged
+.PHONY: createsnapshot
```
oterral: Can we use the same as in mf-geoadmin  `make SNAPSHOT=true deploydev`

loicgasser: Yes good point I want it to be as close as what we have in geoadmin.

loicgasser:
[Geoadmin branch working with this branch](https://mf-geoadmin3.dev.bgdi.ch/gal_make_chsdi/)

loicgasser:
we need to adapt https://github.com/geoadmin/deploy

procrastinatio:
In _ /srv/tomcat/tomcat1/conf/server.xml_ you have:

```<Host name="localhost"  appBase="webapps"
            unpackWARs="true" autoDeploy="true">```
which is the same on _dev_, _int_ and _prod_

**Comment by loicgasser** [comment]
Thanks @procrastinatio , I figured I need to be patient with tomcat to extract this file, it can take up to 20 seconds before the extract is performed.

**Comment by loicgasser** [comment]
But good to know where this configuration comes from !

File: rc_int [hunk 3]
```diff
@@ -0,0 +1,19 @@
+export APACHE_BASE_PATH=main
+export API_URL=//mf-chsdi3.int.bgdi.ch
+export DBHOST=pg.bgdi.ch
```
gjn: small thing, but I think this should be pg-sandbox.bgdi.ch for integration env.

procrastinatio:
Event. also add a `cleanbranch` target removing code and apache config for various branches.

rc_int
loicgasser: At the moment `buildout_int.cfg` extends `buildout.cfg` and we have `pg.bgdi.ch` see https://github.com/geoadmin/mf-chsdi3/blob/master/buildout.cfg#L105
This may be an error in current buidlout system though. @ltclm what instance are we suppose to point to in `int`?

loicgasser:
@procrastinatio yes we can do that to. PRs welcome though. ;)

rc_int
gjn: Yes, it's currently wrong

pg.bgdi.ch should be limited to production systems only. This is to avoid that in some testing/integration environment a run-wild tests consumes all our connections there.

All other system should use pg-sandbox (or pg-0.dev)

Makefile
```diff
@@ -0,0 +1,381 @@
...
+	sudo /etc/init.d/tomcat-tomcat1 restart
+	@echo "${GREEN}It may take a few seconds for $(PRINT_OUTPUT) directory to appear...${RESET}"
+
+# FIXME default target is always master
+# Remove when ready to be merged
+.PHONY: deploydev
+deploydev:
+	@if test "$(SNAPSHOT)" = "true"; \
+	then \
+			scripts/deploydev.sh -s $(GIT_BRANCH); \
```
oterral: Weird indent

Makefile
```diff
@@ -0,0 +1,381 @@
...
+		--var "shortener_allowed_hosts=$(SHORTENER_ALLOWED_HOSTS)" \
+		--var "shortener_allowed_domains=$(SHORTENER_ALLOWED_DOMAINS)" $< > $@
+
+requirements.txt:
+	@echo "${GREEN}File requirements.txt has changed${RESET}";
+.venv: requirements.txt
+	@echo "${GREEN}Install Cython before any deps as some need it to compile with optimizations...${RESET}";
+	@if [ ! -d $(INSTALL_DIRECTORY) ]; \
+	then \
+		virtualenv $(INSTALL_DIRECTORY) --system-site-packages; \
```
oterral: indent

oterral:
I'm testing the branch :)

loicgasser:
Know issue is the fact that msgfmt is not on in nor prod. So can't po to mo on those env.

rc_int
loicgasser: yes I fix

README.md
```diff
@@ -42,24 +42,20 @@ Add .boto to your environment
...
 [Nagios Check for Dynamodb Dumps](https://dashboard.bgdi.ch/cgi-bin/nagios3/extinfo.cgi?type=2&host=ip-10-220-4-46.eu-west-1.compute.internal&service=DynamoDB+backup)
 
-Bootstrap your build environment:
-
-    python bootstrap.py --version 1.5.2 --distribute --download-base http://pypi.camptocamp.net/distribute-0.6.22_fix-issue-227/ --setup-source http://pypi.camptocamp.net/distribute-0.6.22_fix-issue-227/distribute_setup.py
-
 Create a developer specific build configuration:
 
-    cp buildout_ltgal.cfg buildout_<username>.cfg
+    cp rc_example rc_user_<username>.cfg
```
oterral: Why don't  we use rc_XXXX like in mf-geoadmin ?

oterral: and to be consistent with other file we should remove the extension `.cfg`

loicgasser: Because we don't want to commit the personal configs in the main project, for clarity sake and because of make user target.

loicgasser: that's an error in the doc thks no cfg

Makefile
loicgasser: fixed

README.md
loicgasser: See .gitignore

oterral: we should do the same in mf-geoadmin also

loicgasser: yes it's planned.

Makefile
```diff
@@ -0,0 +1,382 @@
...
+.PHONY: help
+help:
+	@echo "Usage: make <target>"
+	@echo
+	@echo "Possible targets:"
+	@echo "- all                Build the app"
+	@echo "- user               Build the user specific version of the app"
+	@echo "- dev                Build the dev version of the app"
+	@echo "- int                Build the int version of the app"
+	@echo "- prod               Build the productive version if the app"
```
oterral: What is the purpose of dev, int , prod. is it used by deploydev,int,prod ?

loicgasser: yes it is used in `deploy/hooks/post-restore-code` it is not a necessity, if people don't like it we can remove it.

oterral: I guess we can remove it from the help because I just run it to see what happens with it and obviously I create a conflict between apache confs.

and then I wanted to revert back but `make user` doesn't do a `make clean` . If it's not in the help nobody will use it.

loicgasser: good point i'll remove it

loicgasser:
https://github.com/geoadmin/mf-chsdi3/issues/1854

loicgasser:
@procrastinatio concerning the deletebranch target
https://github.com/geoadmin/mf-chsdi3/commit/818f7caf06fe510bba753080b3a1fd8cd26ded91

loicgasser:
@procrastinatio @gjn @oterral @danduk82 

Antother reason to completely isolate ourselves from the python packages and re-install gdal in the virtualenv!

docutils missing in int/prod and causes the app to crash.
When using `python-virtualenv` we need to use the option ` --system-site-packages` if we want to use the GDAL install of the server.

Now when we are building the app in dev, it finds this global dependency and as a result doesn't install it in the `python virtualenv`.

Then we deploy the app and the package `docutils` is not on int!!!
So it can't be found and unless we rebuild the whole app on the int or prod instances (which we clearly want to avoid) there is nothing to do.

In dev the package is installed:

```
~$ /usr/bin/python
Python 2.7.3 (default, Mar 13 2014, 11:03:55) 
[GCC 4.7.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
import docutils
```

On int:

```
~$ /usr/bin/python
Python 2.7.3 (default, Mar 13 2014, 11:03:55) 
[GCC 4.7.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import docutils
  File "<stdin>", line 1, in <module>
ImportError: No module named docutils
>>> 
```

So I am in favour of totally isolating ourselves from the system.

loicgasser:
@procrastinatio Some news regarding mapproxy getcap?

procrastinatio:
@loicgasser GetCapabilities ? GetCapabilities are served by `mf-chsdi3`whiles tiles either by S3 or Mapproxy (https://github.com/geoadmin/service-mapproxy). Since yesterday we have no more `mapproxy`in chsdi3 :-)

loicgasser:
Great, thanks a lot! This means that IGN won't be impacted if we merge this one, one extra check.

loicgasser:
I adapted the PR so that the difference between test/int/prod server don't impact the deploy process anymore. So it is no longer a requirement.

gjn:
ok.

As discussed on Tuesday, we will wait for the merge of this until after next weeks deploy.

procrastinatio:
@loicgasser I still have worries about building thins on production instance. I made some test using `Debian packages`(.deb), building the project somewhere and using a standard `apt-get install` command elsewhere. I see several advantages:
- Build are really immutable
- Deploing (and reverting) is a matter of seconds and can be done in parallel, not quarter of an hour
- We have strict dependcies check. I was able to run `chsdi3` with a pretty vanilla Wheezy with 300 packages (instead of our monsters +1200 packages)
- We may have several packages (doc doesn't need to be deployed or build everywhere)
- Another idea is to run the new and the old version in parallel (branch style), and to deploy by switching version in the apache config.
- etc...

loicgasser:
@gjn 

Yes I understand that, after skietouren stuff.

@procrastinatio 

> I still have worries about building thins on production instance. I made some test using Debian 
packages(.deb), building the project somewhere and using a standard apt-get install command elsewhere. I see several advantages:

On int and prod we only build templates and replace the `.war`. (See `post-restore-code`) Current build does rebuild the whole app. In principle I think it would be a good thing to create a Debian package per CHSDI version, how do you think this would take? We could use this version until the Debian package install is ready.

loicgasser:
@procrastinatio Where can  I find your source code?

gjn:
@loicgasser This needs a rebased.

I want this in master ASAP. What needs to be done first?

gjn:
Also, a squash would be nice.

gjn:
a sample user file would be nice, with ALL variables that NEED to be adapted.

gjn:
What exactly is rebuild during deploy in a) the snapshot directory and b) the target instances?

Is running of nosetests within snapshot before deploy to target still supported?

loicgasser:
> I want this in master ASAP. What needs to be done first?
> Also, a squash would be nice.

yes I'm on it.

> What exactly is rebuild during deploy in a) the snapshot directory and b) the target instances?

a) From dev server, I clean the generated files, source with the correct target and recreate the template launch the tests from dev with int or prod variables. (pretty much like before) Per default when a snapshot is deployed we now use the `deploy.cfg` of the snapshot but you can define a custom deploy configuration as discussed. (see doc)

b) Once on the instance in the post-restore-code we compile and replace the war, we regenerate the templates only (no doc, translations etc..) this time (in case the deploy command is used directly) 

> Is running of nosetests within snapshot before deploy to target still supported?

yes see a)

Let's discuss it tomorrow. We also have a new line of servers and I have a question regarding your ci configuration files.

loicgasser:
To have a look at the new doc:
https://github.com/geoadmin/mf-chsdi3/tree/gal_makefile

gjn:
@procrastinatio @loicgasser ready to merge?

gjn:
Maybe not yet, status Andrea pending.

loicgasser:
I'll start preparing geoadmin and chsdi on the new instances.

loicgasser:
TO INTEGRATE changes in https://github.com/geoadmin/mf-chsdi3/pull/1920

PR #1824, opened 2015-12-23
Change Massstab label in printed PDF

gjn:
This is a little present for our Zeitreihen Gurus.

It changes the `Massstab` label in the PDF to `Druckmassstab`.

@davidoesch Could you give the go for this? And verify rumantsch

@cedricmoullet Could you verify french translation?

@AFoletti Could you verify italian translation?

### DIFF
--- a/print/config.yaml
+++ b/print/config.yaml
@@ -297,7 +297,7 @@ layouts:
               fontSize: 7
               align: center
               text: |
-                Massstab 1: ${format %,d scale}
+                Druckmassstab 1: ${format %,d scale}
                 Gedruckt am ${now dd.MM.yyyy HH:mm}
                 ${shortLink}
             - !text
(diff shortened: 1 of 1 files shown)


AFoletti:
IT is good to me, but I would not capitalize "stampa"

davidoesch:
a no go for this 
 "Druckmassstab" - is more confusing for the normal user than it actually helps. It is still the map scale. what the users sees is a map, a map does have a scale. (take a printed map from swisstopo: it clearly states scale and not print scale)
User will start to ask WHY we changed it . since scale IS correct. for the PDF size choosen. 

The Term Druckmassstab does simply not exist (google it or see https://de.wikipedia.org/wiki/Ma%C3%9Fstab_(Kartografie) )

Sorry to be here picky.. 


to make the zeitreisen guys happy: https://github.com/geoadmin/mf-geoadmin3/issues/2931

PR #1825, opened 2015-12-23
Fix rarely failing varnish tests

gjn:
I hope that this will fix the 1-in-2 failing tests we currently have. The timestamp used can conicide with an previous timestamp and giving back wrong results (contain geometry if used with timestamp of previous request that had correct referer).

Note that this probably fixes the sometimes failing tests, but we have a general problem with this
service and the caching mechanism that needs to be adressed.

I'll open an issue for that and auto-merge this fix.

### DIFF
--- a/chsdi/tests/e2e/test_varnish.py
+++ b/chsdi/tests/e2e/test_varnish.py
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
 
 import os
-import random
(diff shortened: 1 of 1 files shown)


gjn:
@danduk82 This will hopefully fix the failing builds we talked about.

gjn:
jenkins test

PR #1833, opened 2016-01-06
Correct modelname for globalstrahlung

gjn:


gjn:
@AFoletti a little correction for your branch.

PR #1835, opened 2016-01-07
Add order by distance parameter to identify

gjn:
@AFoletti this is a PR for your solarenergie branch. We need this to optimize the sonnendach application.

This PR adds 2 new parameters to identify requests, which will both be undocumented:

a) `limit` to be able to limit result set. Currently, all results up to 200 are returned. With this parameter, we can now specify a lower limit. Note that in any case, never more that 200 features are returned.

b) `order` to be able to sort result according to distance to request point. You have to specify `order=distance` for it to have any effect. There's an internal security guard (see comments) that should assure that not too many results need to be ordered, as this could potentially impact the db heavily.

@loicgasser I'll merge this directly into Ambrogios branch but would be glad if you could review it.

### DIFF
--- a/chsdi/lib/validation/mapservice.py
+++ b/chsdi/lib/validation/mapservice.py
@@ -34,6 +34,8 @@ class MapServiceValidation(MapNameValidation):
             'esriGeometryPolygon',
             'esriGeometryEnvelope'
         )
+        self._limit = None
+        self._order = None
 
     @property
     def where(self):
@@ -94,6 +96,14 @@ class MapServiceValidation(MapNameValidation):
     @property
     def offset(self):
         return self._offset
+
+    @property
+    def limit(self):
+        return self._limit
+
+    @property
+    def order(self):
+        return self._order
 
     @where.setter
     def where(self, value):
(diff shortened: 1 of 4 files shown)


chsdi/lib/validation/mapservice.py
```diff
@@ -235,3 +245,15 @@ def offset(self, value):
...
+    def limit(self, value):
+        if value is not None:
+            if value.isdigit() and int(value) >= 0:
+                self._limit = int(value)
+            else:
+                raise HTTPBadRequest('Please provide a positive integer for the parameter limit')
+
+    @order.setter
+    def order(self, value):
+        self._order = value
```
loicgasser: The only valid value for now is distance. So you could check that the same way. first `if value is not None` and then `if value == 'distance`. The rest is the bad params.

gjn: Pondered that. Didn't want to give any indication that we even have an order parameter.

But you are probably right. We should treat it the same. I'll adapt.

loicgasser: Additionally it should work only if a  spatial filter is defined. (e.g. the parameter geometry is defined)
I know it's the case in the code but could make a check here.

chsdi/views/features.py
```diff
@@ -369,7 +380,13 @@ def _get_features_for_filters(params, models, maxFeatures=None, where=None):
                 query = query.filter(timeInstantColumn == params.timeInstant)
 
             # Add limit
-            query = query.limit(maxFeatures) if maxFeatures is not None else query
+            flimit = maxFeatures
+            if params.limit is not None:
+                if flimit is not None:
+                    flimit = params.limit if params.limit <= flimit else flimit
+                else:
+                    flimit = params.limit
```
loicgasser: minor, you could simply do:
`min(flimit, params.limit)`

loicgasser:
small comments, looks good. Thanks for the tests and validation.

gjn:
Did adapt according to comments. Better code now.

Thanks alot for the review.

gjn:
jenkins test

gjn:
jenkins test

gjn:
Unfortunately, can't use min(x, y) directly. It will return None when either of the parameters are None.

loicgasser:
yes ok

gjn:
found a combination by filtering for None first. But None is smaller than everything in python, so min(...None...) always returns None.

loicgasser:
nice

PR #1837, opened 2016-01-08
New instances for prod deployement

gjn:


loicgasser:
Shall we change the number of WSGI threads as we now have a smaller instance? Current setup with 30 on prod and 15 on dev.

gjn:
Certainly something to look at. Currently, instances seem to run fine with the 30 thread setting.

PR #1840, opened 2016-01-11
Increase coverage chsdi.views.search

mariavkourouni:


gjn:
Cool. Thanks alot.

loicgasser:
Ok for the extra tests but I disagree with the changes from assertEqual to assertTrue. In gemeral prefer using assertEqual as it provides better error messages. I made a PR recently to address this issue and now we re changing it back.
@mariavkourouni any reason to change that that I might have missed?

mariavkourouni:
@loicgasser I totally agree with you.
The changes were reverted because of a rebase issue I had.

PR #1845, opened 2016-01-13
[POC Shop]: Add area computation support using an arbitrary polygon for clipping

loicgasser:
`/rest/services/ech/GeometryServer/cut?geometryType=esriGeometryEnvelope&geometry=545000,145000,555000,155000&layers=all:ch.bafu.bundesinventare-bln&groupby=bln_name`

Service name, code organisation is subject to changes.
Base code is there, we need to figure out where we want to define the property on which we'll `GROUP BY` .

### DIFF
--- a/chsdi/__init__.py
+++ b/chsdi/__init__.py
@@ -95,6 +95,7 @@ def main(global_config, **settings):
     config.add_route('stationboard', '/stationboard/stops/{id}')
     config.add_route('stationboard_destination', '/stationboard/stops/{id}/destinations')
     config.add_route('faqlist', '/rest/services/{map}/faqlist')
+    config.add_route('cut', '/rest/services/{map}/GeometryServer/cut')
 
     # Some views for specific routes
     config.add_view(route_name='dev', renderer='chsdi:templates/index.mako')
--- a/chsdi/lib/validation/geometryservice.py
+++ b/chsdi/lib/validation/geometryservice.py
@@ -1,0 +1,134 @@
+# -*- coding: utf-8 -*-
+
+
+from pyramid.httpexceptions import HTTPBadRequest
+
+from chsdi.lib.validation import BaseFeaturesValidation
+from chsdi.esrigeojsonencoder import loads
+
+
+class GeometryServiceValidation(BaseFeaturesValidation):
+
+    def __init__(self, request):
+        super(GeometryServiceValidation, self).__init__(request)
+
+        self.totalArea = False
+
+        self._clipper = None
+        self._geometry = None
+        self._geometryType = None
+        self._returnGeometry = None
+        self._layers = None
+        self._groupby = None
+        self._chargeable = None
+
+        clipper = request.params.get('clipper')
+        geometry = request.params.get('geometry')
+        geometryType = request.params.get('geometryType')
+
+        self.esriGeometryTypes = (
+            'esriGeometryPolygon',
+            'esriGeometryEnvelope'
+        )
+        self.geometryTypes = (
+            'Polygon',
+            'MultiPolygon'
+        )
+
+        # No parameter -> we want the totalPerimeter
+        if clipper is None and geometry is None and geometryType is None:
+            self.totalArea = True
+
+        if not self.totalArea:
+            if not clipper:
+                self.geometry = geometry
+                self.geometryType = geometryType
+            self.clipper = clipper
+            self.groupby = request.params.get('groupby')
+        self.layers = request.params.get('layers', 'all')
+
+    @property
+    def clipper(self):
+        return self._clipper
+
+    @property
+    def geometry(self):
+        return self._geometry
+
+    @property
+    def geometryType(self):
+        return self._geometryType
+
+    @property
+    def layers(self):
+        return self._layers
+
+    @property
+    def groupby(self):
+        return self._groupby
+
+    @property
+    def chargeable(self):
+        return self._chargeable
+
+    @clipper.setter
+    def clipper(self, value):
+        # Either a geometry or a clipper must be defined
+        if all((self._geometry, self._geometryType)):
+            return
+        if value is None:
+            raise HTTPBadRequest('Please provide a clipper (Required if geometry is not defined)')
+        temp = value.split(':')
+        if len(temp) != 2:
+            raise HTTPBadRequest(
+                'Unexpected clipper syntax. It should use the following syntax (layerBodId:featureId)')
+        else:
+            self._clipper = temp
+
+    @geometry.setter
+    def geometry(self, value):
+        if self._clipper:
+            return
+        if value is None:
+            raise HTTPBadRequest('Please provide the parameter geometry (Required if clipper is not defined)')
+        try:
+            self._geometry = loads(value)
+        except ValueError:
+            raise HTTPBadRequest('Please provide a valid geometry')
+
+    @geometryType.setter
+    def geometryType(self, value):
+        if self._clipper:
+            return
+        if value is None:
+            raise HTTPBadRequest('Please provide the parameter geometryType (Required if clipper is not defined)')
+        if value not in self.esriGeometryTypes:
+            raise HTTPBadRequest(
+                'Please provide a valid geometry type. Available: (%s)' % ', '.join(self.esriGeometryTypes))
+        self._geometryType = value
+
+    @layers.setter
+    def layers(self, value):
+        if value is None:
+            raise HTTPBadRequest('Please provide a parameter layers')
+        if value == 'all':
+            self._layers = value
+        else:
+            try:
+                layers = value.split(':')[1]
+                self._layers = layers.split(',')
+            except:
+                HTTPBadRequest('There is an error in the parameter layers')
+
+    @groupby.setter
+    def groupby(self, value):
+        if value is not None:
+            self._groupby = value.split(',')
+
+    @chargeable.setter
+    def chargeable(self, value):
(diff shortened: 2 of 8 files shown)


loicgasser:
Here are a couple of test links using our current layers. I made it so that it works with all the layers.

@oterral @p1d1d1 @gjn @pauloamado 
Do you think the following links make sense?

[Test Link one layer no price distinction and bbox](https://mf-chsdi3.int.bgdi.ch/gal_clip_area/rest/services/ech/GeometryServer/cut?geometryType=esriGeometryEnvelope&geometry=545000,145000,555000,170005&layers=all:ch.swisstopo.images-swissimage.metadata)
[Test Link one layer price distinction and bbox](https://mf-chsdi3.int.bgdi.ch/gal_clip_area/rest/services/ech/GeometryServer/cut?geometryType=esriGeometryEnvelope&geometry=545000,145000,555000,170005&layers=all:ch.swisstopo.images-swissimage.metadata&groupby=datenstand)
[Test Link 2 layers with price distinction and a bbox](https://mf-chsdi3.int.bgdi.ch/gal_clip_area/rest/services/ech/GeometryServer/cut?geometryType=esriGeometryEnvelope&geometry=545000,145000,555000,170005&layers=all:ch.swisstopo.images-swissimage.metadata,ch.swisstopo.swissboundaries3d-gemeinde-flaeche.fill&groupby=datenstand,kanton)
[Test Link 2 layers with price distinction and a complex polygon for clipping](https://mf-chsdi3.int.bgdi.ch/gal_clip_area/rest/services/ech/GeometryServer/cut?geometry={"rings":[[[675000,245000],[670000,255000],[680000,260000],[690000,255000],[685000,240000],[675000,245000]]]}&geometryType=esriGeometryPolygon&layers=all:ch.swisstopo.images-swissimage.metadata,ch.swisstopo.swissboundaries3d-gemeinde-flaeche.fill&groupby=datenstand,kanton)

@p1d1d1 @pauloamado 
How accurate do we need to be in square kilometers?

p1d1d1:
It seems to me a good start, but don't understand the use case with 2 layers... we don't need it.
For the accuracy you can round to 2 digits after the comma.

Please just remember that the service should be able to read a geometry accepting an object Id as input, so something like:

intersect?geometryType=esriGeometrypolygon&geometry=ch.swisstopo.swissboundaries3d-kanton-flaeche.fill=2&layers=all:ch.swisstopo.swissimage-product

And now I see the use case for 2 layers: swissimage has two perimeters (stopo_master.public.ganzer_datensatz_si_level2_25 and stopo_master.public.ganzer_datensatz_si_level2_50). If we don't want to make an "union" of these to layers in a new layer (as I suggested in the document), then we can do something like:

intersect?geometryType=esriGeometrypolygon&geometry=ch.swisstopo.swissboundaries3d-kanton-flaeche.fill=2&layers=all:ch.swisstopo.swissimage-25,ch.swisstopo.swissimage-25

But his implies that on the side Shop they cannot use the technical name on the product (ch.swisstopo.swissimage-product in this case): instead they should use the "perimeter_layer"

p1d1d1:
PS: looking at your implementation in detail I guess that on the Shop side they shall in any case use in "layers" the "perimeter_layer": is that right?

Alternatively they can send the product layer (ch.swisstopo.swissimage-product) and then is up to your service to understand which "perimeter_layer" is associated to the "product layer".
On the other hand this second solution could vehiculate a false idea: that one could clip a raster layer.

Don't know, but maybe the first option is the best one!?

loicgasser:
> For the accuracy you can round to 2 digits after the comma.

ok thanks


> Please just remember that the service should be able to read a geometry accepting an object Id as input, so something like:

> intersect?geometryType=esriGeometrypolygon&geometry=ch.swisstopo.swissboundaries3d-kanton-flaeche.fill=2&layers=all:ch.swisstopo.swissimage-product

I would replace:
`geometry=ch.swisstopo.swissboundaries3d-kanton-flaeche.fill=2`
with
`geometry=ch.swisstopo.swissboundaries3d-kanton-flaeche.fill:2`

It's dangerous to have an `=` sign within a parameter. And we can remove the `geometryType` as this won't be used in this case. And also change the parameter name as this could be confusing.

So yes the id of the feature we clip with could be defined in this request. (With a different syntax but it's a detail). This will come soon.

> And now I see the use case for 2 layers: swissimage has two perimeters (stopo_master.public.ganzer_datensatz_si_level2_25 and stopo_master.public.ganzer_datensatz_si_level2_50). If we don't want to make an "union" of these to layers in a new layer (as I suggested in the document), then we can do something like:

> intersect?geometryType=esriGeometrypolygon&geometry=ch.swisstopo.swissboundaries3d-kanton-flaeche.fill=2&layers=all:ch.swisstopo.swissimage-25,ch.swisstopo.swissimage-25

> But his implies that on the side Shop they cannot use the technical name on the product (ch.swisstopo.swissimage-product in this case): instead they should use the "perimeter_layer"

No, they can use the technical name, it's not an issue. That's the purpose of the `groupby` parameter which at the moment can be either:
- Defined in the layers configuration service and sent as an extra parameter (current implementation). They can use a dedicated metadata service to discover which attribute determine the price of different parts of the same layer. Or sent via the `GET price` service if needed.
- Hard coded server side (the same way we hard code the queryable attributes for instance)

> PS: looking at your implementation in detail I guess that on the Shop side they shall in any case use in "layers" the "perimeter_layer": is that right?

> Alternatively they can send the product layer (ch.swisstopo.swissimage-product) and then is up to your service to understand which "perimeter_layer" is associated to the "product layer".
On the other hand this second solution could vehiculate a false idea: that one could clip a raster layer.

> Don't know, but maybe the first option is the best one!?

Is the user supposed to use directly ch.swisstopo.swissimage-produc?
Or will he have a custom `metadata` layer like this one: (e.g. ch.swisstopo.images-swissimage.metadata)
https://s.geo.admin.ch/696c28aac9

This is not clear to me yet. Anyway not a problem as we can use any model back-end.

loicgasser:
Also for the use case where we need to clip a layer with a particular feature.
Not changing anything to the current code do:

- Perform a `GET Feature Resource` request, namely (id): http://api3.geo.admin.ch/rest/services/api/MapServer/ch.swisstopo.images-swissimage.metadata/1267-33
 to retrieve the geometry
- And then use the geometry with the current service

We'd loose a bit of performance, and it may be problematic if the request gets too long.

p1d1d1:
The issue is that:
* not always there is a .metadata layer available
* sometimes the perimeter of a product is not exactly the merge of the .metadata layer.

Your idea of using the .metadata layer is nice, would avoid to use other layers, but this stategy cannot alway be used.

The situation is: each product has a "perimeter" layer (the tables ganzer_datensatz in our DB) associated and this is hardcoded/configured server side. I wouldn't like to integrate these tables as .metadata layers. (We just need to check that we use the same geometries on our side and in the GDWH: i will check this)

I would keep the public interface as simple as possible. Somethig like:

/clip/?layers=all:ch.swisstopo.swissimage-product&geometry=""

Where:

ch.swisstopo.swissimage-product is the product and geometry can be either a bbox or an object ID.

loicgasser:
Ok then so we simply need to link our current models to these product perimeters.

Not a problem technically, thank you for the clarification.

loicgasser:
[Clip with Broye Vully no price distinction](https://mf-chsdi3.int.bgdi.ch/gal_clip_area/rest/services/ech/GeometryServer/cut?clipper=ch.swisstopo.swissboundaries3d-bezirk-flaeche.fill:2222&layers=all:ch.swisstopo.images-swissimage.metadata)
[Clip with Broye Vully price distinction](https://mf-chsdi3.int.bgdi.ch/gal_clip_area/rest/services/ech/GeometryServer/cut?clipper=ch.swisstopo.swissboundaries3d-bezirk-flaeche.fill:2222&layers=all:ch.swisstopo.images-swissimage.metadata&groupby=datenstand)

I still want to refactor some of this code and add tests, but functionality is there.

loicgasser:
address updated

ltclm:
swissimage is a special case. 
it is a mixture of dop25 perimeter and dop50 perimeter, see:
http://cw-datenstand.prod.bgdi.ch/?initialState=swissimage_resolution&reset_session

Information about the resolution perimeter:
https://github.com/geoadmin/cw-datenstand/blob/master/server_conf/datenstand/datenstand.map.in#L1524

in the old cartoweb solution, there is a lot of valuable information in the mapfiles.

p.e. **swisstlm3d** perimeter perimeter geometry and **fixed area value**
https://github.com/geoadmin/cw-toposhop/blob/master/cartoweb3/projects/toposhop/server_conf/vector/vector.map#L7266 

p.e. **swissimage level 2 dop 25**
https://github.com/geoadmin/cw-toposhop/blob/master/cartoweb3/projects/toposhop/server_conf/silevel2/silevel2.map#L499

@p1d1d1 this information has to be checked by the data owner and mapped to our chsdi layers configuration.

loicgasser:
Thank you for your input and the very useful rferences @ltclm. We need to discuss the integration of these layers for this service, but the functionality should remain the same as the one implemented in this PR.

p1d1d1:
@ltclm I cannot access the github repository you mentioned in your comment.
By the way I guess I will also check with the GDWH. With the new DI process we should take the info from the GDWH so that we just have one source for the data.

gjn:
Replaced by https://github.com/geoadmin/mf-chsdi3/pull/2083. I'll close this one.

PR #1846, opened 2016-01-14
Update oev legends

alcapat:


loicgasser:
chsdi/static/images/legends/ch.bav.haltestellen-oev_it.png is slightly bigger than the original

gjn:
@alcapat what about the sizes?

Issue #1795, opened 2015-12-03
[Tooltip window]: remove "Information" row


mariokeusen:
UC: open app, add a layer with a tooltip, click an object
Result: see screenshot. We have the window title "Objekt-Information" and then above the attribute content we have an additional row "Information"
Expected/question: as discussed with @davidoesch we could get rid of this row "Informatin" (because redundant with window title) in order to save space

@AFoletti @ltclm @cedricmoullet @gjn : does someone see some risks, if we remove this row?
![tooltip_info_label](https://cloud.githubusercontent.com/assets/4577749/11563961/e9c14af8-99d6-11e5-8166-cfa5d8f03be0.PNG)

cedricmoullet:
Fine for me.

2015-12-03 16:00 GMT+01:00 mariokeusen <notifications@github.com>:

> UC: open app, add a layer with a tooltip, click an object
> Result: see screenshot. We have the window title "Objekt-Information" and
> then above the attribute content we have an additional row "Information"
> Expected/question: as discussed with @davidoesch
> <https://github.com/davidoesch> we could get rid of this row "Informatin"
> (because redundant with window title) in order to save space
>
> @AFoletti <https://github.com/AFoletti> @ltclm <https://github.com/ltclm>
> @cedricmoullet <https://github.com/cedricmoullet> @gjn
> <https://github.com/gjn> : does someone see some risks, if we remove this
> row?
> [image: tooltip_info_label]
> <https://cloud.githubusercontent.com/assets/4577749/11563961/e9c14af8-99d6-11e5-8166-cfa5d8f03be0.PNG>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/geoadmin/mf-chsdi3/issues/1795>.
>



-- 
Twitter: http://twitter.com/cedricmoullet
Linked In: http://www.linkedin.com/in/cedricmoullet

<http://twitter.com/cedricmoullet>

AFoletti:
I think it makes sense.
I'll wait for @ltclm and @gjn answers and then remove it if accepted.

gjn:
Fine with me.

PR #1868, opened 2016-01-28
Remove info header in tooltip

loicgasser:
Fix for https://github.com/geoadmin/mf-chsdi3/issues/1795


loicgasser:
@davidoesch ok with that see reference in the ticket.

davidoesch:
Ok

PR #1869, opened 2016-01-28
Reorganize templates

loicgasser:
No functional change, just a reorganization of some templates.


loicgasser:
@procrastinatio could you have a look, it's quite basic.

Issue #1709, opened 2015-10-15
Layer search: words in title with more weight than in description


gjn:
When doing a search with type `layers`, the words in the layer title should weight more than the words in the description.

It should be in a way that part-word matches in the title weight slighlty more than full word matches in the description.

PR #1870, opened 2016-01-28
Give a higher imp to words in the title

loicgasser:
This will affect the ranking in the layer search.

This is to prioritize titles over description fields works with:
https://github.com/geoadmin/service-sphinxsearch/pull/256
and fixes:
https://github.com/geoadmin/mf-chsdi3/issues/1709

### DIFF
--- a/chsdi/views/search.py
+++ b/chsdi/views/search.py
@@ -77,7 +77,8 @@ class Search(SearchValidation):
             for key, value in enumerate(self.results['results']):
                 translation = re.search(r'.*(<i>[\s\S]*?<\/i>).*', value['attrs']['label']) or False
                 if translation:
-                    self.results['results'][key]['attrs']['label'] = value['attrs']['label'].replace(translation.group(1), '<i>%s</i>' % str(self.request.translate(translation.group(1)).encode('utf-8')))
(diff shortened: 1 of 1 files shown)


chsdi/views/search.py
```diff
@@ -167,6 +168,13 @@ def staging_filter(staging):
         self.sphinx.SetLimits(0, layerLimit)
         self.sphinx.SetRankingMode(sphinxapi.SPH_RANK_WORDCOUNT)
         self.sphinx.SetSortMode(sphinxapi.SPH_SORT_EXTENDED, '@weight DESC')
+        # Weights defaults to 1
+        self.sphinx.SetFieldWeights({
+            '@title': 4,
+            '@detail': 2,
+            '@layer': 1
```
gjn: I would switch weights for @layer and @detail.

loicgasser: layer is the technical name, that's for us only and if you have the technical name well you usually only have one result. but I can change it.

loicgasser: @gjn still want me to change it?

gjn:
We keep it as it is. But we merge this not for tomorrow's deploy.

PR #1877, opened 2016-02-02
Look for available serachField on all models [ltfoa_solarenergie_daecher]

loicgasser:
One layer can have several models and sometimes only some of those models have the so called `queryable` properties. This PR takes this fact into account.
 
@AFoletti 
I see that you're using a view `view_solarenergie_daecher_gs` which is probably huge!
A find operation is quite demanding if we cannot make use of the indices. Any chance we might have a real table for that?

For `df_uid`: Do we need other queries than an excat match? Otherwise I would say we don't need it in the find because it's also the `primary key` and `GET Feature` would be more approriate.

### DIFF
--- a/chsdi/models/__init__.py
+++ b/chsdi/models/__init__.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from sqlalchemy.ext.declarative import declarative_base
-from sqlalchemy import engine_from_config
+from sqlalchemy import engine_from_config, Column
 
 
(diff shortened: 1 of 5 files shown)


AFoletti:
jenkins test

AFoletti:
it look good to me. Thanks!

AFoletti:
@loicgasser , should we not remove the master PR? The solarenergie branch will be merged in a couple of days (I hope)

loicgasser:
It won't create any rebase issue as it is the same commit. In case, this is not merged soon enough I'd like to have this bug fixed.

PR #1878, opened 2016-02-02
Look for available serachField on all models [master]

loicgasser:
same as https://github.com/geoadmin/mf-chsdi3/pull/1877 but for master

One layer can have several models and sometimes only some of those models have the so called queryable properties. This PR takes this fact into account.


gjn:
We'll merge the other PR soon (I expect tomorrow), so I close this.

Issue #1591, opened 2015-07-31
Keep old-ga* files in service


gjn:
We have old, undocumented api files with ol3.4 (old-ga.js, old-ga-debug.js, old-ga.css) that are currently used as emergency patch by one if our clients. As they have a longer update process, they need those files to stay around.

Remove those files only once this ticket is resolved.

gjn:
Client informed us that he's able to go to production with the official version on 10.11.2015 (maybe 20.10.2015).

PR #1884, opened 2016-02-04
Remove old unused files from api

gjn:
KOW!

Fix for https://github.com/geoadmin/mf-chsdi3/issues/1591

All thos old api files should not be needed anymore.


loicgasser:
Kill of the week? Have we checked that it is not used in Kibana?

loicgasser:
please merge if you're sure we won't be bothered on googleGroup. I like!

gjn:
I'm not sure :=). But it has been communicated several times already.

Thanks for the review.

PR #1885, opened 2016-02-04
Get remote datenstand from geojson

loicgasser:
I tried to make as impact less as I could.
See:
https://github.com/geoadmin/mf-geoadmin3/issues/2366

We could test this function which would in turn test the existence of the geojson and the timestamp field but I would need to add it in the e2e tests I think. do you agree?

### DIFF
--- a/chsdi/templates/legend.mako
+++ b/chsdi/templates/legend.mako
@@ -113,7 +113,9 @@
     </tr>
     <tr>
       <td>${_('Datenstand')}</td>
-% if times:
+% if legend['datenstand']:
+      <td>${legend['datenstand']}</td>
+% elif times:
   % if len(times) == 4:
       <td>${times}</td>
   % elif len(times) == 6:
--- a/chsdi/views/layers.py
+++ b/chsdi/views/layers.py
@@ -3,7 +3,7 @@
 import os
 import decimal
 import datetime
-
+import requests
 
 from pyramid.view import view_config
 from pyramid.renderers import render_to_response
@@ -92,7 +92,8 @@ def legend(request):
                 layerMetadata['attributes']['dataStatus'] = params.translate('None') + params.translate('Datenstand')
     legend = {
         'layer': layerMetadata,
-        'hasLegend': _has_legend(layerId, params.lang)
+        'hasLegend': _has_legend(layerId, params.lang),
+        'datenstand': _get_geojson_datenstand(params, layerId)
     }
     response = render_to_response(
         'chsdi:templates/legend.mako',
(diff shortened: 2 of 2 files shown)


loicgasser:
[Test Link](http://mf-chsdi3.dev.bgdi.ch/gal_remote_datenstand/rest/services/all/MapServer/ch.bafu.hydroweb-messstationen_gefahren/legend?lang=en)

gjn:
As we need `Datenstand` for other applications as well (smo), would it make sense to create a `datenstand` service for each layer?

loicgasser:
As part of sphinx index? @ltclm What do you think? It would be enough?

ltclm:
i would say, kiss:

we can easily extract the datenstand of this json with 
```bash
$ curl -I "http://data.geo.admin.ch/ch.bafu.hydroweb-messstationen_gefahren/ch.bafu.hydroweb-messstationen_gefahren_de.json"
HTTP/1.1 200 OK
Accept-Ranges: bytes
Age: 0
Content-Length: 108665
Content-Type: text/plain
Date: Fri, 05 Feb 2016 15:09:29 GMT
ETag: "9b5a4a11f543da0e3a32ecaff052b408"
Last-Modified: Fri, 05 Feb 2016 15:00:09 GMT
Server: AmazonS3
Via: 1.1 varnish-v4
X-Cache: MISS
X-Varnish: 179018379
``` 
or fme and publish it with our standard method which is a table deploy of bod_master.public.dataset_timestamps each 4 hours.

ltclm:
we just have to change 
https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/templates/legend.mako#L116
that we can pass full dates
`` %Y-%m-%d %H:%M:%S ``

ltclm:
i have added the layer to the fme workbench, which is executed every 2h.
it makes a HEAD request on the json file and writes the last-modified date to the mentioned table.

gjn:
Thanks @ltclm . @loicgasser Do you think that the last modified date is enough?

ltclm:
i think the database solution is working fine, 
i would not add this to sphinx, we should seperate between fulltextsearch information (sphinx) and layer configuration and attributes (bod).

loicgasser:
Thks @ltclm it much better this way.
We now have a FME workbench here: https://github.com/geoadmin/bgdi-scripts/blob/master/syncserv/fme/bod_datenstand_update.fmw
code still need to be adapted but we can close this one.

gjn:
Nice solution for this. Thanks alot!

PR #1886, opened 2016-02-05
Extract JS code in FAQ doc and create testable faqlist service

mariavkourouni:
[FAQ doc test-link](http://mf-chsdi3.dev.bgdi.ch/kom_faqlist/api/faq/index.html)

[FAQ list JSON](http://mf-chsdi3.dev.bgdi.ch/kom_faqlist/rest/services/all/faqlist)

### DIFF
--- a/chsdi/static/doc/source/api/faq/index.rst
+++ b/chsdi/static/doc/source/api/faq/index.rst
@@ -59,20 +59,20 @@ Which layers are available ?
 
 Some layers can’t be freely used. These layers are accessible by the way of `swisstopo web access - WMTS documentation <http://www.swisstopo.admin.ch/internet/swisstopo/en/home/products/services/web_services/webaccess.html>`_
 
-Here is a list of the layers that requires a swisstopo web acesss:
+The list below includes the layers that require swisstopo web acesss:
 
 .. raw:: html
 
    <body>
-      <div id="notfree" style="margin-left:10px;margin-bottom:24px;"></div>
+      <div id="chargeableLayers" style="margin-left:10px;margin-bottom:24px;"></div>
    </body>
 
-Here is a list of all the freely accessible layers:
+The following list contains all the free accessible layers:
 
 .. raw:: html
 
    <body>
-      <div id="free" style="margin-left:10px;margin-bottom:24px"></div>
+      <div id="notChargeableLayers" style="margin-left:10px;margin-bottom:24px"></div>
    </body>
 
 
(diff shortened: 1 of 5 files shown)


chsdi/views/faqlist.py
```diff
@@ -0,0 +1,66 @@
+# -*- coding: utf-8 -*-
+
+import requests
+from requests.exceptions import RequestException
+from pyramid.view import view_config
+from pyramid.httpexceptions import (HTTPNotFound, HTTPBadRequest)
+
+
+@view_config(route_name='faqlist', renderer='json')
```
loicgasser: we could use jsonp, it's useful sometimes

chsdi/views/faqlist.py
```diff
@@ -0,0 +1,66 @@
...
+def faqlist(self):
+    try:
+        layersConfigURL = self.route_url('layersConfig', map='api')
```
loicgasser: please use db directly and avoid round trips.

```python
from chsdi.models import LayersConfig
query = request.db.query(LayersConfig)
```

chsdi/views/faqlist.py
```diff
@@ -0,0 +1,66 @@
...
+                if 'chargeable' in layer and not layer['chargeable']:
+                    try:
+                        no_chargeable_layers.append(layer_name)
+                    except:
+                        pass
+        except RequestException:
+            pass
+
+    if (len(queryable_layers) + len(searchable_layers) + len(chargeable_layers) + len(no_chargeable_layers)) == 0:
+        raise HTTPBadRequest()
```
loicgasser: why is that? `HTTPBadRequest` should be not thrown here. You can use that when parameters are bad/wrong.

chsdi/views/faqlist.py
```diff
@@ -0,0 +1,66 @@
...
+    chargeable_layers = []
+    no_chargeable_layers = []
+
+    for layer_name in layers_api:
+        try:
+            layer = layers_api[layer_name]
+            if 'parentLayerId' not in layer and not str(layer_name).endswith(suffix3d):
+                if 'queryable' in layer and layer['queryable']:
+                    try:
+                        queryable_layers.append(layer_name)
```
loicgasser: what are you afraid of here? no need to `try/except` every time.

loicgasser:
I'll come back to it later when adapted. come to my office if you have questions.

loicgasser:
@mariavkourouni I was thinking that you can also adapt the faq page at the same time.

chsdi/views/layers.py
```diff
@@ -160,6 +160,56 @@ def insertValueAt(field, attrName, value):
...
+                notChargeableLayers.append(l)
+
+    queryableLayers.sort()
+    tooltipLayers.sort()
+    searchableLayers.sort()
+    chargeableLayers.sort()
+    notChargeableLayers.sort()
+    return {
+        'translations': translations,
+        'queryableLayers': queryableLayers,
```
loicgasser: for less code simply do:
```python
queryableLayers': queryableLayers.sort(),
```

mariavkourouni: Like this it was returning null, probably because of the comma exactly after `.sort()`

loicgasser:
yes makes sense that it returns null, good alternative.

Looking good. We just need 1 or 2 integration test for this service and I ll merge. thanks! @gjn maybe could have a look to make sure this is what you want.

AFoletti:
What's the better way to test it?

loicgasser:
@AFoletti 

2 ways to do that:
1) Test that the service is responding correctly with integration tests
2) When deploying to int create a diff with the previous version of this service to check that the new/updated layers possess the correct features.

gjn:
Both 1) and 2) is part of @mariavkourouni tasks.

mariavkourouni:
yes, I am going to do them!

mariavkourouni:
1) Integration tests are done for the service
2) Diff-script is ready, so when deploying to int, the comparison between phases could be implemented

loicgasser:
@mariavkourouni looks good please squash and rebase

chsdi/tests/integration/test_mapservice.py
```diff
@@ -34,6 +34,14 @@ def test_metadata_chargeable_false(self):
...
         self.assertEqual(resp.content_type, 'application/json')
 
+    def test_faqlist(self):
+        resp = self.testapp.get('/rest/services/api/faqlist', status=200)
+        self.assertEqual(resp.content_type, 'application/json')
+
+    def test_faqlist_topic_all(self):
+        resp = self.testapp.get('/rest/services/all/faqlist', status=200)
+        self.assertEqual(resp.content_type, 'application/json')
+
```
gjn: Please check here that we have at least x layers in the results. (x high, but 2-3 smaller than current count).

Also, check for correct structure (by checking for the 4 lists, and 1-2 layers in each list)

chsdi/static/doc/source/api/faq/index.rst
```diff
@@ -173,16 +123,16 @@ Not all the layers have a tooltip. Here is a complete list of all the layers tha
 Which layers are searchable?
 ****************************
 
-We define a layer as searchable when its features can be searched. Here is a list of all searchable layers:
+We define a layer as searchable when its features can be searched. Here there is the list of all searchable layers:
```
loicgasser: you corrected the language but I think it's wrong

chsdi/static/doc/source/api/faq/index.rst
```diff
@@ -173,16 +123,16 @@ Not all the layers have a tooltip. Here is a complete list of all the layers tha
...
 .. raw:: html
 
   <body>
-    <div id="searchable" style="margin-left:10px;margin-bottom:24px;"></div>
+    <div id="searchableLayers" style="margin-left:10px;margin-bottom:24px;"></div>
   </body>
 
-How can I accessed the tiles ?
-******************************
+How can I have access to the tiles ?
```
loicgasser: :+1:

chsdi/static/doc/source/api/faq/index.rst
loicgasser: ok I fix myself

gjn:
Thanks alot @mariavkourouni and @loicgasser . Good improvement.

PR #1890, opened 2016-02-08
add helper function to format datenstand dates with time

ltclm:
the function splits the text, detects YYYYMM and YYYYMMDD dates and reformats them to DD.MM.YYYY
will be used to publish dates with time https://github.com/geoadmin/mf-chsdi3/pull/1885

### DIFF
--- a/chsdi/lib/helpers.py
+++ b/chsdi/lib/helpers.py
@@ -298,11 +298,26 @@ def center_from_box2d(box2D):
     ]
 
 
-def parse_date_string(dateStr):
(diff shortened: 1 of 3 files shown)


loicgasser:
Ok that's nice I'am adding a test now. Thks a lot @ltclm !

loicgasser:
So we don't have the time as before, but we now have the date and it comes from the DB. I think this should be enough.

ltclm:
i will add the time part as soon as this pr is deployed to prod

loicgasser:
ok great thanks @ltclm I'll merge this one then as we're deploying tomorrow.

PR #1895, opened 2016-02-09
Fix translation oev in extended tooltip

pfanguin:


loicgasser:
minor but from DB schema i can see that `betriebspunkttyp` column is theoretically nullable.
same for `verkehrsmittel`.

The code doesn't reflect that. I know we didn't care too much so far, but I sincerly think this could avoid a lot of mistakes/reduce the amount of code we need in the makos.

In the model declaration we would simply need to add:

`betriebspunkttyp = Column('betriebspunkttyp', Text, nullable=False)`

instead of

`betriebspunkttyp = Column('betriebspunkttyp', Text)`

@AFoletti , @ltclm, @daguer , @alcapat 

Would it be ok for furture updates or new integration to add this info? It must reflect what we have in DB.

```python
>>> from chsdi.models.vector.uvek import OevHaltestellen
>>> OevHaltestellen.betriebspunkttyp
<sqlalchemy.orm.attributes.InstrumentedAttribute object at 0x2d2d490>
>>> OevHaltestellen.betriebspunkttyp.nullable
True
>>> OevHaltestellen.betriebspunkttyp.nullable
```

that would be the added value. would make attentive everyone to the importance of knowing those things.

PR #1896, opened 2016-02-09
add rss feed script

rfrancel:
@davidoesch could you please check if this is what you want? You might want to change title and description in the channel.
test link: http://mf-chsdi3.dev.bgdi.ch/ltfrr/releasenotes/
@loicgasser could you please review to see if the method is optimal?

linked to issue: https://github.com/geoadmin/mf-geoadmin3/issues/3039


davidoesch:
Any Chances to get here a nicer styling in the output? Open the RSS in feedly - it is all in one string. Having carriage returns line breaks tables and links supported would be useful. Currently barely readable

rfrancel:
Yes I agree I will look at that

De : David Oesch [mailto:notifications@github.com]
Envoyé : mardi 9 février 2016 12:50
À : geoadmin/mf-chsdi3 <mf-chsdi3@noreply.github.com>
Cc : Francelet Rebecca swisstopo <Rebecca.Francelet@swisstopo.ch>
Objet : Re: [mf-chsdi3] add rss feed script (#1896)


Any Chances to get here a nicer styling in the output? Open the RSS in feedly - it is all in one string. Having carriage returns line breaks tables and links supported would be useful. Currently barely readable

scripts/rssFeedGen.py
```diff
@@ -0,0 +1,95 @@
+from bs4 import BeautifulSoup
+from datetime import datetime
+import time
+from email.Utils import formatdate
+import PyRSS2Gen
+import StringIO
+import re
+
+def extract_releases(html):
+    html_doc = open(html, 'r').read()
```
loicgasser: Usually better to do:
```python
with open(html, 'r') as f:
    soup = BeautifulSoup(f.read(), "lxml")
```

This way the file is closed automatically when you are out of the `with` context.

scripts/rssFeedGen.py
```diff
@@ -0,0 +1,95 @@
...
+            if length == 1:
+                data = cols[0].text
+            else:
+                status = cols[0].text
+                details = cols[1].text
+                data = data + status + ": " + details + "\n "
+    else:
+        data = "No change in the geodata"
+    return data
+
```
loicgasser: usually from here we would use:

```python
if __name__ == '__main__':
    your code to execute in the script
````

File: scripts/rssFeedGen.py [hunk 3]
```diff
@@ -0,0 +1,95 @@
...
+from email.Utils import formatdate
+import PyRSS2Gen
+import StringIO
+import re
+
+def extract_releases(html):
+    html_doc = open(html, 'r').read()
+    soup = BeautifulSoup(html_doc, "lxml")
+    divRelease = soup.find("div", {"id":"release-notes"})
+    releases = divRelease.findAll("div", {"id":re.compile("release-")})
```
loicgasser: maybe use a class select `section` instead within the div containing all the releases

loicgasser: sorry you can ignore this comment

loicgasser:
In general try to use only simple quotes when possible here we have a mix.

scripts/rssFeedGen.py
```diff
@@ -0,0 +1,95 @@
...
+    return releases
+
+def id_to_pubDate(r):
+    id_str = r.findNext("span").get("id")
+    date_str = ''
+    for i in id_str.split():
```
loicgasser: this achieves nothing really it puts a string in a list.
What you want to do is split on `-`
So:
```python
date_str = id_str.split('-')[1]
```
remove the loop and everything inside

scripts/rssFeedGen.py
```diff
@@ -0,0 +1,95 @@
...
+import re
+
+def extract_releases(html):
+    html_doc = open(html, 'r').read()
+    soup = BeautifulSoup(html_doc, "lxml")
+    divRelease = soup.find("div", {"id":"release-notes"})
+    releases = divRelease.findAll("div", {"id":re.compile("release-")})
+    return releases
+
+def id_to_pubDate(r):
```
loicgasser: I would prefer `id_to_pub_date` to be consistent

scripts/rssFeedGen.py
```diff
@@ -0,0 +1,95 @@
...
+        for l in i:
+            if l.isdigit():
+                date_str = date_str + l
+    date_str = date_str + " 00:00:00"
+    date_obj = datetime.strptime(date_str, '%Y%m%d %H:%M:%S')
+    date_rss = datetime.strftime(date_obj, "%a, %d %b %Y %H:%M:%S %z")
+    return date_rss
+
+def extract_api_changes(r):
+    api_changes = r.findNext("a",{"href":re.compile("https://github.com/geoadmin/mf-chsdi3")}, text="Full changelog")
```
loicgasser: I would remove the text arg and use `https://github.com/geoadmin/mf-chsdi3/compare/` until compare. It would be slightly more robust.

scripts/rssFeedGen.py
```diff
@@ -0,0 +1,95 @@
...
+    if api_changes != None:
```
loicgasser: usually `is not None` is preffered

scripts/rssFeedGen.py
```diff
@@ -0,0 +1,95 @@
...
+        api_change = "-Bug Fixes\n-Full changelog: " + api_changes.get("href")
+    else:
+        api_change = ''
+        for li in r.findAll("li"):
+            api_change = "- " + li.text + "\n  "
+    return api_change
+
+def extract_map_changes(r):
+    map_changes = r.findNext("a",{"href":re.compile("https://github.com/geoadmin/mf-geoadmin3")}, text="Full changelog")
```
loicgasser: same here use `re.compile('https://github.com/geoadmin/mf-geoadmin3/compare')`

scripts/rssFeedGen.py
```diff
@@ -0,0 +1,95 @@
...
+    if map_changes != None:
```
loicgasser: `is not None`

scripts/rssFeedGen.py
```diff
@@ -0,0 +1,95 @@
...
+    # parse information from html
+    date_rss = id_to_pubDate(r)
+    api_change = extract_api_changes(r)
+    map_change = extract_map_changes(r)
+    data = extract_geodata(r)
+    # create feeds
+    items.append(PyRSS2Gen.RSSItem(
+            title = r.findNext("h2").text[:-1],
+            link = "//api3.geo.admin.ch/releasenotes/" + r.findNext("a").get("href"),
+            description = "API & APPLICATIONS\n API\n  Bugfixes\n  Full changelog: " + api_change + "\n MAP\n  Bugfixes\n  Full changelog: " + map_change + "\nGEODATA\n  " + data,
```
loicgasser: For the description, you should be able to put directly all the html we have within the release.
According to RSS specs you can put html in the description field but you just need to encode it.
see: http://cyber.law.harvard.edu/rss/encodingDescriptions.html

You could maybe get rid of extract_api_changes, extract_map_changes, extract_map_changes all together.

loicgasser: See a RTS flux as an exemple:
https://www.rts.ch/espace-2/programmes/a-vous-de-jouer/?flux=rss

@davidoesch this is normal you can't read it as is.
@rfrancel If you want nicely formatted html in the description field you can use `prettify()` method of Beautifulsoup

loicgasser:
It works fine by me. I added a couple of comments but I think it's fine.
I'll help you integrate it in the build process.
Could you check your styling by adding `scripts` in https://github.com/geoadmin/mf-chsdi3/blob/master/buildout.cfg#L206
and then launch a buildout.

Good job! Don't forget to edit the head of the document, see this link:
http://www.petefreitag.com/item/384.cfm

loicgasser:
Put your RSS http://mf-chsdi3.dev.bgdi.ch/ltfrr/releasenotes/rss2.xml

in http://www.davidjuth.com/rest-demo-jquery-rss.aspx

You can see that at the moment the html is not correctly interpreted.

davidoesch:
If the content can't be properly interpreted we should add at least let it open the release notes site. Eg if you add http://www.edparsons.com/2016/02/im-back-talking-aviation-films/ as rss feed into feedly (most used rss reader) you get a nice description and on click on the description it opens the nicely formatted text - obviously from the site itself- in feedly. This is what we need . if just feed the user all in plain text without formatting- users are lost... M

loicgasser:
You ll get formatting that's not an issue it s just a matter off escaping it correctly. we ll add a JS module to have the feed nicely formatted on the welcome page I was thinking. We could also add our twitts directly in the rss feed as well. that s a option you just need to activate on the Twitter account.
That being said it will help us have some visibilty on the web.

davidoesch:
Only activating release notes. We will activate Twitter in the rss feed off the cms

rfrancel:
@davidoesch I integrated the html in the rss feed. However, it takes the css from the reader. In Feedly,  the result is not really nice, especially with Chrome (IE is better than chrome, I couldn't check with firefox). 
http://mf-chsdi3.dev.bgdi.ch/frr_rss_feed/releasenotes/rss2.xml

@loicgasser I adapted the script with your comments. Marc helped to integrate the buildout to generate the rss. If you could have a look to check if it's ok like that, that would be nice.

davidoesch:
Hi Rebecca... Works like a charm now on feedly. Thanks.

scripts/rssFeedGenFull.py
```diff
@@ -0,0 +1,84 @@
...
+        title = 'GeoAdmin - RSS Feed',
+        link = '//api3.geo.admin.ch/releasenotes/',
+        description = "The latest news about GeoAdmin application's changes, new and updated data available on map.geo.admin.ch",
+        lastBuildDate = time.strftime('%a, %d %b %Y %H:%M:%S %z'),
+        items = items
+    )
+ 
+    # write into xml file
+    rss = rss.to_xml('utf-8')
+    xml = open('chsdi/static/doc/build/releasenotes/rss2.xml', 'w')
```
loicgasser: with open ;)

rfrancel:
sorry..

scripts/rssFeedGenFull.py
```diff
@@ -0,0 +1,84 @@
...
+    return data
+
+def data_to_description(data):
+    data = data.decode('utf-8', 'ignore')
+    description = data.encode('ascii', 'ignore')
+    return description
+
+if __name__ == '__main__':
+    items = []
+    releases = extract_releases('chsdi/static/doc/build/releasenotes/index.html')
```
loicgasser: Minor: maybe check if this file exists before trying to open it.

```python
try:
    extract_releases(pathToReleasenotes)
except IOError as e:
    print '%s does not exist' % pathToReleasenotes
    sys.exit(1)
```

It is possible not to have it as the doc is generated.

loicgasser:
That's some good looking code.

A possible improvement would be to update this file only when we have new release notes, not compulsary though.

rfrancel:
@loicgasser I added the error handling you suggested and as discussed, will wait for the make to be integrated to handle the update of the feeds. Thank you for your comments and help!

loicgasser:
thanks!

PR #1901, opened 2016-02-09
Use query only on models where prop is available

loicgasser:


loicgasser:
@procrastinatio quick review please this is wanted for tomorrow.

gjn:
Shouldn't it be bad request if we try to use a where clause with an unknown column for at least one layer?

Might be philosophical question.

loicgasser:
> Shouldn't it be bad request if we try to use a where clause with an unknown column for at least one layer?

Here it's to handle the case where we have different models for the same layer with different columns/queryable attributes. Currently you can use the `where` clause where with only one layer at a time.

gjn:
Understood. Thanks for the explanation.

PR #1905, opened 2016-02-11
[WIP] Pseudo static file for layersConfig

procrastinatio:
In order to put geoadmin in S3, we need a `layersConfig` service which behave like a file, i.e. instead of [layersconfig?lang=rm](http://mf-chsdi3.dev.bgdi.ch/mom_layersconfig_lang/rest/services/all/MapServer/layersConfig?lang=rm) the service may be requested as [layersConfig.rm](http://mf-chsdi3.dev.bgdi.ch/mom_layersconfig_lang/rest/services/all/MapServer/layersConfig.rm.json).

Unfortunately, the whole localize buisiness in pyramid is done at a very early stage (before routing in this respect), and is hard to overcome. `mod_rewrite`can do the trick, 

See https://github.com/geoadmin/mf-geoadmin3/issues/3092 for rational behing this PR

Note: `.rm` is stopped by BIT proxy 

This PR is only adding some functionalities, not changing anything existing.

### DIFF
--- a/apache/wsgi.conf.in
+++ b/apache/wsgi.conf.in
@@ -35,6 +35,11 @@ RedirectMatch permanent ^${apache_entry_path}/iipimage/(.*)$ ${apache_entry_path
 # Redirect no-slash target to slashed version
 RedirectMatch ^${apache_entry_path}$ ${apache_entry_path}/
 
+
+# Makes layersConfig looks like a real file
+RewriteRule ^${apache_entry_path}/(rest/services/all/MapServer/layersConfig)\.(de|fr|en|it|rm)\.json ${apache_entry_path}/$1?lang=$2 [PT]
+RewriteRule ^${apache_entry_path}/[0-9]+/(rest/services/all/MapServer/layersConfig)\.(de|fr|en|it|rm)\.json ${apache_entry_path}/$1?lang=$2 [E=${apache_base_path}setyearcache:true,PT]
+
 # If URI has numbers at the start, we cache a year
 # This allows a client to create their own cache and
 # update at his discretion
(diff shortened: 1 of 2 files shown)


loicgasser:
@procrastinatio can we close as I don't see ever making it to master. I understand that as a testing feature only. Or am I missing something?

procrastinatio:
@loicgasser This is for geoadmin3. In S3, all services are static files (even /src). For development on mf-geoadmin3 dev, files in **/prd** as statified services, but **/src** still uses the services. But if developpers prefer to rebuild *geoadmin3* after every BOD change and always uses static files, I don't care.

loicgasser:
Ok, I understand now, but we're adding an apache rule for dev that will be present in prod as well that we don't need. So I would add this rule conditionnally in `${deploy_target}  == 'dev'` for instance.

loicgasser:
`chsdi/tests/apache/test_apache.py`  should be in `e2e` tests.

gjn:
Don't we treat this in the deploy build of mf-geoadmin3? We have both topics service and layersConfig file(s) as files inside prd/cache directory there.

gjn:
Needs discussion as I don't understand yet.

loicgasser:
Would be good to spend some time discussing it after vacations. I get it but there might be a better way. @procrastinatio can you arrange a meeting?

procrastinatio:
You do not have a fancy web server on S3. The two requests `toto?lang=fr`or `toto?lang=de`, gives you only `toto`, i.e. the very same file, or maybe I miss something very important about S3.
The alternative is to have a separates `index.html` files for every languages (ugly) or put translations elsewhere (in `locales/de.json` for instance, big breaking changes)

gjn:
I think we need to adapt mf-geoadmin3 as well to request the file like configurations. Then we could get rid of https://github.com/geoadmin/mf-geoadmin3/blob/master/apache/app.mako-dot-conf#L43

Could you try to do that?

gjn:
Not needed as we create those files directly in mf-geoadmin3 during deploy.

PR #1912, opened 2016-02-16
Improve full text search perfs

loicgasser:
Using `like` in conjuction with `lower` is faster than `ilike`. @ltcm can confirm I think.

### DIFF
--- a/chsdi/lib/filters.py
+++ b/chsdi/lib/filters.py
@@ -34,15 +34,15 @@ def filter_by_map_name(query, model, mapName):
     if mapName != 'all':
         clauses = []
         if mapName in ('api', 'swissmaponline'):
-            clauses.append(model.maps.ilike('%%%s%%' % mapName))
+            clauses.append(model.maps.like('%%%s%%' % mapName.lower()))
         else:
             # add background layersConfig
             if model.__tablename__ == LayersConfig.__tablename__:
                 isBgLayer = True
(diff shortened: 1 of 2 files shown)


gjn:
Doens't this require that all texts in the db to be lower case?

loicgasser:
Yes.

Topics are always lower case.

`full_text_search` is used in https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/views/layers.py#L49
The fullTextSearch, layerBodId and geoCatId are lower case.

`full_text_search` is also used in the find service.
https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/views/features.py#L471
This might be an issue. I may only change it for the topic search to be on the safe side.

gjn:
Yes. It would be nice to have sample tests for all use cases you mentioned.

I assume the `fullTextSearch` column of the layers is a special index with lower case only. Is this correct?

loicgasser:
yes

loicgasser:
let's just do that for the topics.

gjn:
Any metrics on the performance improvements?

loicgasser:
just a stackoverflow http://stackoverflow.com/questions/20336665/lower-like-vs-ilike
It's a minor change as it just affects the topic search

gjn:
ok. Please merge once build is ok.

PR #1913, opened 2016-02-16
Fix style and remove folder from linting

loicgasser:
mapproxy folder does not exist anymore and some style fix.


loicgasser:
Trivial self merge

PR #1914, opened 2016-02-16
Default resolutions fior 21781 should not be in layersConfig

loicgasser:
It fixes a regression introduced in https://github.com/geoadmin/mf-chsdi3/commit/d7da86e536fe48e8b8a5225372149efaa4e090b9#diff-2e90e144f9058f27411cceaa0039ca9d

The default resolutions are currently always sent. This PR fixes this issue.
It also introduces custom resolutions for tileMatrixSet different from the default one for native tiles in WGS84. (not used yet) See: https://github.com/geoadmin/mf-geoadmin3/issues/3124

It adds some tests and improves the overall traceback on failure.

### DIFF
--- a/chsdi/models/bod.py
+++ b/chsdi/models/bod.py
@@ -50,6 +50,9 @@ class Bod(object):
 
 
 class LayersConfig(Base):
+
+    defaultMatrixSet21781 = '21781_26'
+
     __tablename__ = 'view_layers_js'
     __table_args__ = ({'schema': 're3', 'autoload': False})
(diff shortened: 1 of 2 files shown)


loicgasser:
@gjn This PR fixes an issue where the default resolutions are is the layersConfig service when they should not. See: https://github.com/geoadmin/mf-geoadmin3/blob/master/src/components/map/MapService.js#L28

Again Jenkins errors are not related.

gjn:
Code LGTM.

loicgasser:
thanks for the double check!

PR #1920, opened 2016-02-18
Do not build on int,prod

loicgasser:
This PR does 2 things
1. We don't build on int and prod instances anymore when deploying the main project to int or prod
2. We avoid invoking git in snapshot directory by using the `.last-git-branch` file generated during deploydev

@gjn this is untested as I know you're currently deploying and it something to test on master

We could take the same approach in geoadmin if this is proven to be successful.

### DIFF
--- a/deploysnapshot.sh
+++ b/deploysnapshot.sh
@@ -19,22 +19,27 @@ cwd=$(pwd)
 # Go into snapshot directory to run nose-tests
 cd $SNAPSHOTDIR/chsdi3/code/chsdi3
 
-# Run nose tests with target cluster db
-if [ -z $3 ] || [ $3 != "notests" ]
+if [ "$2" == "int" ]
 then
-    if [ "$2" == "int" ]
-    then
-      echo "Running nose tests with integration cluster in $SNAPSHOTDIR"
-      ./nose_run.sh -i
-    fi
+  if [ -z $3 ] || [ $3 != "notests" ]
+  then
+    echo "Running nose tests with integration cluster in $SNAPSHOTDIR"
+    ./nose_run.sh -i
+  fi
+  buildout/bin/buildout -c buildout_int.cfg
+elif [ "$2" == "prod" ]
+then
+  if [ -z $3 ] || [ $3 != "notests" ]
+  then
+    echo "Running nose tests with production cluster in $SNAPSHOTDIR"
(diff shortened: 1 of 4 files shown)


gjn:
Agree with general approach. Please wait for tests until tomorrow.

deploysnapshot.sh
```diff
@@ -19,22 +19,27 @@ cwd=$(pwd)
...
-      ./nose_run.sh -p
-    fi
-else
-    echo "You have disabled nosetests!"
+  if [ -z $3 ] || [ $3 != "notests" ]
+  then
+    echo "Running nose tests with integration cluster in $SNAPSHOTDIR"
+    ./nose_run.sh -i
+  fi
+  buildout/bin/buildout -c buildout_$2.cfg
```
gjn: bould be explicit int here.

deploysnapshot.sh
```diff
@@ -19,22 +19,27 @@ cwd=$(pwd)
...
+elif [ "$2" == "prod" ]
+then
+  if [ -z $3 ] || [ $3 != "notests" ]
+  then
+    echo "Running nose tests with production cluster in $SNAPSHOTDIR"
+    ./nose_run.sh -p
+  fi
+  buildout/bin/buildout -c buildout_$2.cfg
```
gjn: explicit prod here

deploysnapshot.sh
loicgasser: yes note that `"$2" == "int"` so it cannot be anything else

loicgasser:
yep I should probably install the only print-war in post-restore-code as well.
Another good reason to remove the print from chsdi in a near future.
I am not sure what we're doing in nose_run is very relevant. I would personally get rid of it and lauch the tests directly in deploysnapshot after building the app with correct variables.

gjn:
just looked at nose_run myself. seems like it's needed to have dedicated environment for those runs. (we specify dedicated port in the nose-specific configs). But this might be solved differently as well.

Yes, print install is still needed. @procrastinatio Pressure for v3 separated...

procrastinatio:
@gjn Separating the mapfish print using the current settings (vhosts + make + deploy) can be done very easily, as V3 will take time.

gjn:
ok. let's separate print first then. service-print.

@procrastinatio please take the lead.

PR #1921, opened 2016-02-18
Enable querying of null and not null values

loicgasser:
This is the first part of https://github.com/geoadmin/mf-geoadmin3/issues/2013

[Test link null](http://mf-chsdi3.dev.bgdi.ch/gal_query_null/rest/services/all/MapServer/identify?geometryFormat=geojson&lang=en&layers=all:ch.bafu.gewaesserschutz-klaeranlagen_reinigungstyp&timeInstant=2014&where=andere_stoffe+is+null)
[Test link not null](http://mf-chsdi3.dev.bgdi.ch/gal_query_null/rest/services/all/MapServer/identify?geometryFormat=geojson&lang=en&layers=all:ch.bafu.gewaesserschutz-klaeranlagen_reinigungstyp&timeInstant=2014&where=andere_stoffe+is+not+null)

@oterral Can you take care of the front-end part?
And also fix the bug where the second condition doesn't provide the button for the suggestions?
See screen shot for condition 2:
suggestions_missing_cond2
After that I think we can remove the `(beta)` in the title of this tool.

Empty strings should be set to null in DB as this is the purpose of null values.

FYI: This can be merged safely even if the front-end part has not been adapted yet.

### DIFF
--- a/chsdi/lib/helpers.py
+++ b/chsdi/lib/helpers.py
@@ -165,7 +165,7 @@ def format_query(model, value):
                 match.group(2),
                 match.group(3),
                 match.group(4).replace(u'\\', u'\\\\')
-                              .replace(u'\'', u'\\\'')
+                              .replace(u'\'', u"\''")
                               .replace(u'_', u'\\_'),
                 match.group(5)
             ))
@@ -181,7 +181,11 @@ def format_query(model, value):
             res.append(val)
         return res
 
-    extractMatches = lambda x: x[0] if x[0] != '' else x[1]
+    def extractMatches(x):
+        for v in x:
+            if v != '':
+                return v
+        return v
 
     def getOperator(values):
         supportedOperators = [' and ', ' or ']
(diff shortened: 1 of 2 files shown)


chsdi/lib/helpers.py
```diff
@@ -193,7 +197,7 @@ def getOperator(values):
             return operator
         return ''
 
-    regEx = r'(\w+\s(?:ilike|not ilike)\s(?:\'%)[^\'%]+(?:%\'))|(\w+\s(?:=|\!=|>=|<=|>|<)\s[^\s]+)'
+    regEx = r'(\w+\s(?:ilike|not ilike)\s(?:\'%)[^\'%]+(?:%\'))|(\w+\s(?:=|\!=|>=|<=|>|<)\s[^\s]+)|(\w+\s(?:is|is not)\s(?:null))'
```
gjn: Not sure here, but a query containing `is bullshit` would pass this, not?

loicgasser: Just tested the resulting where is an empty string, meaning that we don't use any filter when it should return a bad request. so I think this could be improved. I'll adapt.

loicgasser: or just no result, what would be best?

gjn: I would probably be explicit: `(?:is null`|`is not null)`

loicgasser: yes will do

loicgasser:
@gjn Unrelated jenkins errors see:
https://github.com/geoadmin/mf-chsdi3/issues/1939
https://github.com/geoadmin/mf-chsdi3/pull/1935

I update the PR and also fixed a bug in the query where the `'` was not escaped correctly.

loicgasser:
rebased

gjn:
LGTM. Please merge when build is good.

PR #1925, opened 2016-02-19
Fix type in prod configuration

gjn:
@loicgasser This was the cause of the jenkins api3.geo.admin.ch job failure. Simple type.

It would have broken prod build :=)

I'll merge, move your test_tag and we should be good.


gjn:
jenkins build the PR well.

loicgasser:
:+1: Great! We would have tested the prod variables eventually. Many thanks for this find

Issue #1924, opened 2016-02-19
WSGI configurations in new Makefile setup


gjn:
In buildout setup, the *.ini file used for the mod_wsgi was as follows:

for dev/int/prod in main, it was always production.ini
for user specific environments, it was always development.ini
(was `config-file` paramter of [modwsgi] buildout section.

With Makefile setup, it's as follow:
for dev and all user specifics, it's development.ini
for int/prod it's productions.ini

I think it should be productions.ini in dev main evnironment as well. development.ini should be strictly for development work.

loicgasser:
Agreed, we don't need debugging in main.

PR #1926, opened 2016-02-22
Use production on dev main

loicgasser:
https://github.com/geoadmin/mf-chsdi3/issues/1924


gjn:
Maybe it would make sense to have an rc_user, which is base for all user specific configs (source rc_user in each user specific)...

loicgasser:
A little excessive just for mod_wsgi parameter, what else could we put there?

loicgasser:
wsgi_config sorry

gjn:
It's more mental help. user != dev.

loicgasser:
ok so actually copy the whole dev config but the wsgi_config? or source rc_dev and source rc_user in rc_example? If anything I would prefer the first.

PR #1928, opened 2016-02-22
[ON HOLD] Update iframe example

oterral:
Update for https://github.com/geoadmin/mf-geoadmin3/pull/3164

[Test](http://mf-chsdi3.dev.bgdi.ch/teo_postmtest/examples/geoadmin_iframe.html)

### DIFF
--- a/chsdi/static/doc/examples/geoadmin_iframe.html
+++ b/chsdi/static/doc/examples/geoadmin_iframe.html
@@ -43,6 +43,10 @@
         <th>Feature ID</th>
         <td class="feature-id"></td>
       </tr>
+      <tr>
+        <th>Feature name</th>
+        <td class="feature-name"></td>
+      </tr>
     </table>
     <script type="text/javascript" src="/static/js/jquery-2.0.3.min.js"></script>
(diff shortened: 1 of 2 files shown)


loicgasser:
LGTM in geoadmin and CHSDI but this may potentially break http://www.hydrodaten.admin.ch/

@davidoesch Can you warn hydroweb people to support per default 
https://github.com/geoadmin/mf-chsdi3/pull/1928/files

and also the old behaviour until we deploy a new version of CHSDI please?

davidoesch:
@loicgasser do not push this to production BEFORE we got an OK from BAFU that they tested it. @mariokeusen : can you provide BAFU this testlink so they can check this feature (it will later on as well make the tooltip not necessary anymore, and later enhance as well the communication between iframe and cms - one of the open points we had with tom)

loicgasser:
@davidoesch yes of course that's why I contacted you. Here is a test link: they simply need to change the host. I think that the best for them is to support both temporarly. [Test Host](https://mf-geoadmin3.int.bgdi.ch/fix_3146/)

gjn:
This is an interface that we need to keep stable in the future. So we should really put some effort into it to get it right.

loicgasser:
Not relevant anymore as examples are on codepen now and we decided to use antoher approach

PR #1929, opened 2016-02-23
manage no geometry in search example

rfrancel:
@loicgasser I removed the changes for the test. Indeed, I think you'll be able to read the code without testing.. Thank you!


chsdi/static/doc/examples/geoadmin_search.js
```diff
@@ -95,5 +99,6 @@ $('#search').on('typeahead:selected', function(evt, location, suggName) {
     popup.setPosition([0,0]);
     view.fitExtent(extent, map.getSize());
```
loicgasser: indent within the else please

PR #1941, opened 2016-02-25
Remove git when deploying to dev

loicgasser:
same as in https://github.com/geoadmin/mf-geoadmin3/pull/3174

we save 800Mb transfer when deploying with this approach.


Makefile
```diff
@@ -7,7 +7,7 @@ BRANCH_TO_DELETE :=
 CURRENT_DIRECTORY := $(shell pwd)
 DEPLOYCONFIG ?=
 DEPLOY_TARGET ?=
-GIT_BRANCH ?= $(shell git rev-parse --symbolic-full-name --abbrev-ref HEAD)
+GIT_BRANCH := $(shell if [ -f '.venv/deployed-git-branch' ]; then cat .venv/deployed-git-branch 2> /dev/null; else git rev-parse --symbolic-full-name --abbrev-ref HEAD; fi)
```
gjn: This will remove the abbility to overwrite GIT_BRANCH on the cmdline. The README.md should reflect this (there's an example there).

loicgasser: yes you're right forgot about this one

loicgasser:
updated and rebased

gjn:
Just a thought: how about also doing this after `rm -rf .git*`

```
git init
git add .
git commit -m "Initial commit"
```

Therefore, we could see if snapshot changed after it was created.

PR #1946, opened 2016-02-25
extend pre-commit hook, check for unstaged files

ltclm:


ltclm:
does not make sense, sorry.
this will prevent partial commits

gjn:
it's too bad that hooks are local only and can't be shared.

PR #1947, opened 2016-02-26
Fix cadastral webmap link tests

gjn:
This is to fix failing nightly tests of the cadastral webmap links.


loicgasser:
so ids start at 1 instead of 0. Makes sense.

Issue #1957, opened 2016-03-02
[extended tooltips] layout is broken


mariokeusen:
UC: open app, select and add a layer with an extended tooltip, pick an object and open extended tooltip
Result: the usual layout of the ext. HTML-tooltip has disappeared. It seems to be fully missing
Expected: extended tooltips are layouted as usual and as designed.

Recognized on Chrome and IE and for all ext. tooltips I tested (ca. 10). Important examples:
Teileinzugsgebiete: https://api3.geo.admin.ch/rest/services/energie/MapServer/ch.bafu.wasser-teileinzugsgebiete_2/77215/extendedHtmlPopup?lang=de Important UC here is print: if you want to print the extentded tooltip into a pdf, the map is blurred in E-W direction.
Luftfahrthindernisse: https://api3.geo.admin.ch/rest/services/aviation/MapServer/ch.bazl.luftfahrthindernis/1199/extendedHtmlPopup?lang=de
Windenergieanlagen: https://api3.geo.admin.ch/rest/services/ech/MapServer/ch.bfe.windenergieanlagen/facility_STB/extendedHtmlPopup?lang=de

gjn:
@AFoletti what happened?

loicgasser:
css are missing I think. I ll fix it.

loicgasser:
@alcapat what about a squashTM test on one of those

gjn:
Fixed and deployed.

PR #1961, opened 2016-03-02
Fix ext. tooltip on deploy

loicgasser:
Fix for https://github.com/geoadmin/mf-chsdi3/issues/1957

Because of https://github.com/geoadmin/mf-chsdi3/blob/master/deploy/hooks/post-restore-code#L21

the css for the extended tooltip were missing

@gjn in case you deploy again tomorrow


gjn:
Yes, I think we need to deploy today. Apps only. I will take care of it. Please keep the master stable today.

PR #1963, opened 2016-03-03
[POC] Add test layer meteo

loicgasser:
Not to be merged yet


AFoletti:
@loicgasser the .geojson extension is it important? I save now to .json but it's 15 seconds to adapt it

loicgasser:
no it is not

procrastinatio:
Why do we put data in mf-chsdi3 ? I really want data outside.

loicgasser:
because it's a POc
It will be on S3.

PR #1966, opened 2016-03-07
[WIP] Add grid support in identify/htmlpopup/getfeature

loicgasser:
PR in own branch using the newly created gatilegrid module

So it works I need to adapt a few things and we'll be good.
Note that performance is not great using boto directly.
Maybe we want to perpare batches of requests and perform them in parallel.

for 21 cells it takes around 1.5 secs to prepare. -> to be discussed

Source code for Grid: https://github.com/loicgasser/gatilegrid/blob/master/gatilegrid/grid.py

### DIFF
--- a/chsdi/templates/htmlpopup/windatlas50.mako
+++ b/chsdi/templates/htmlpopup/windatlas50.mako
@@ -1,7 +1,10 @@
 <%inherit file="base.mako"/>
 
 <%def name="table_body(c, lang)">
-  <iframe src="/rest/services/all/MapServer/${c['layerBodId']}/${c['id']}/extendedHtmlPopup?lang=${lang}" width="400" height="400" frameborder="0"  style="border: 0"></iframe>
+  <%
+    baseUrl = request.registry.settings['api_url']
+  %>
+  <iframe src="${baseUrl}/rest/services/all/MapServer/${c['layerBodId']}/${c['id']}/extendedHtmlPopup?lang=${lang}" width="100%" height="400" frameborder="0"  style="border: 0"></iframe>
 </%def>
 
 <%def name="extended_info(c, lang)">
@@ -15,6 +18,7 @@
   topRight = coordinates[2]
   center = [(bottomLeft[0] + topRight[0]) / 2, (bottomLeft[1] + topRight[1]) / 2]
   dhm_altitude = h.filter_alt(r.getVal(center[0], center[1]))
+  center = ', '.join([str(round(center[0], 2)), str(round(center[1], 2))])
   altitude = int(c['layerBodId'].split('ch.bfe.windenergie-geschwindigkeit_h')[1])
  
   props = c['properties']
--- a/chsdi/templates/htmlpopup/windatlas50.mako
+++ b/chsdi/templates/htmlpopup/windatlas50.mako
@@ -1,37 +1,52 @@
 <%inherit file="base.mako"/>
+<%!
+from chsdi.lib.raster.georaster import get_raster
+r = get_raster('COMB')
+%>
+
 
 <%def name="table_body(c, lang)">
-  <%
-    baseUrl = request.registry.settings['api_url']
-  %>
-  <iframe src="${baseUrl}/rest/services/all/MapServer/${c['layerBodId']}/${c['id']}/extendedHtmlPopup?lang=${lang}" width="100%" height="400" frameborder="0"  style="border: 0"></iframe>
+<%
+coordinates = c['geometry']['coordinates'][0]
+bottomLeft = coordinates[0]
+topRight = coordinates[2]
+center = [(bottomLeft[0] + topRight[0]) / 2, (bottomLeft[1] + topRight[1]) / 2]
+dhm_altitude = h.filter_alt(r.getVal(center[0], center[1]))
+center = ', '.join([str(round(center[0], 2)), str(round(center[1], 2))])
+altitude = int(c['layerBodId'].split('ch.bfe.windenergie-geschwindigkeit_h')[1])
+
+props = c['properties']
+
+baseUrl = request.registry.settings['api_url']
+
+%>
+<!-- html output -->
+    <tr><th class="cell-left">${_('tt_bfe_hoehe_ueber_grund')}</th>         <td>${altitude or '-'}</td></tr>
+    <tr><th class="cell-left">${_('tt_bfe_koordinaten')}</th>               <td>todo load grid module with col/row</td></tr>
+    <tr><th class="cell-left">${_('tt_bfe_hoehe_gelaende')}</th>            <td>${dhm_altitude or '-'}</td></tr>
+    <tr><th class="cell-left">${_('tt_bfe_geschw_wind_durchschnitt')}</th>  <td>${props['v_mean']}</td></tr>
+    <tr><td colspan=2><iframe src="${baseUrl}/rest/services/all/MapServer/${c['layerBodId']}/${c['featureId']}/extendedHtmlPopup?lang=${lang}&iframe=true" width="100%" height="400" frameborder="0"  style="border: 0" ></iframe></td></tr>
 </%def>
 
 <%def name="extended_info(c, lang)">
 
 <%
-  from chsdi.lib.raster.georaster import get_raster
+coordinates = c['geometry']['coordinates'][0]
+bottomLeft = coordinates[0]
+topRight = coordinates[2]
+center = [(bottomLeft[0] + topRight[0]) / 2, (bottomLeft[1] + topRight[1]) / 2]
+dhm_altitude = h.filter_alt(r.getVal(center[0], center[1]))
+center = ', '.join([str(round(center[0], 2)), str(round(center[1], 2))])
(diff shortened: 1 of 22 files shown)


gjn:
Awesome!

Let's discuss this tomorrow, but I think we could very well live with point only identify on grids. Hard to see a grid where more makes sense.

What I think is more important is to suppor the other rest service methods for such kind of layers (as indicated in the regular grid doc).

loicgasser:
Yes or limit it to 20 features. Need to adapt htmlpopup getfeature and we should beaucoup good.

loicgasser:
Because of s3 credentials issues this branch cannot be deployed yet. (works only locally)
Will do once this is solved.

loicgasser:
@gjn @ltclm  @davidoesch [No DB grid test link](https://mf-geoadmin3.dev.bgdi.ch/gal_grid/?layers=ch.bfe.windenergie-geschwindigkeit_h50&topic=ech&lang=fr&bgLayer=ch.swisstopo.pixelkarte-farbe&X=158070.00&Y=555000.00&zoom=5)

I would still adpat the css for a better a look and feel in geoadmin for the extended tooltip but it's a good start.

davidoesch:
the extended tooltip is a link and not in the tooltip
but besides that: nicely done and fast

loicgasser:
Yes I agree but you'll have to talk to the data owner about that (or someone else), it seems to be a must for him, or at least that's my understanding of the situation through the grapevine.

loicgasser:
FYI: it's an iframe (because of the external JS resources we don't want in geoadmin)

loicgasser:
jenkins test

davidoesch:
@mariokeusen : please check this:

chsdi/templates/htmlpopup/windatlas50.mako
```diff
@@ -0,0 +1,738 @@
...
+  %>
+  <iframe src="${baseUrl}/rest/services/all/MapServer/${c['layerBodId']}/${c['id']}/extendedHtmlPopup?lang=${lang}" width="100%" height="400" frameborder="0"  style="border: 0"></iframe>
+</%def>
+
+<%def name="extended_info(c, lang)">
+
+<%
+  from chsdi.lib.raster.georaster import get_raster
+
+  r = get_raster('COMB')
```
loicgasser: @gjn @danduk82 This also means that we need to keep the DTM on non-legacy instances (as it is the case at the moment)

davidoesch: an iframe in the simple tooltip is a mess. how to solve this? we need an
expert tooltip

ltclm:
thanks loic, i think we should keep the basic tooltip basic!
by the way, stable id / link to object would make sense for this layer. is this doable?

gjn:
AFAIK, wanted from data owner side @ltclm and @AFoletti can verify:

for basic tooltip: 3-4 values in table + windrose
for extended tooltip: as it currently is

Does it make sense to only put the windrose in the iframe?
(btw: iframe is not a bad solution and not bricolage if done well. we have to avoid bad effects as 2 scroll-bars etc, and it needs to be responsive and working in mobile. But I think that's doable).

Impressive work, @loicgasser

davidoesch:
Hi all
There is a detailed spec in the data integration formular on which the data owner and we agreed. Stick to it. Trust the process. You can store the d3js code in data.geo.admin.ch in the corresponding b layer directory if you don't want to have it in mf-chsdi

gjn:
We take d3 from a cdn.

cedricmoullet:
Don't we have d3 already in mf-geoadmin3 ?
I don't like to be dependent of a cdn.

2016-03-09 23:08 GMT+01:00 Gilbert Jeiziner <notifications@github.com>:

> We take d3 from a cdn.
>
> —
> <https://github.com/geoadmin/mf-chsdi3/pull/1966#issuecomment-194531397>.
>



-- 

ltclm:
@davidoesch should try to convince them to not use the basic tooltips for stuff like that, thats why we have extended tooltips. I think it is wrong to add this to the basic tooltip.

gjn:
cdn is more reliable then our current setup. In fact, cloudfront is a cdn (and the one for d3 is cloudflare). For release libraries, it makes sense.

I think we should think about using cdn for the d3 we are using in geoadmin3, not the other way around.

loicgasser:
I think it's possible technically possible to have the tooltip nicely formatted in the small tooltip with a little css and an eventual parameter in the extended one. We currently do have d3js in geoadmin but:
1. it is not the same version
2. it relies on an extra lib d3-tip
3. the current implementation would pollute the scope

The better solution that also involves a lot more dev in geoadmin (0 at the moment) would be to perform front-end templating. Now if you can convince them otherwise I'd be happy glad not to hack our tooltips. I put the tooltip stuff on hold for now. Let me know what is the decision.

loicgasser:
I guess the grid approach is accepted though and we can go further with this approach though?

ltclm:
@loicgasser 
maybe we could implement a get parameter for the extended tooltip, which returns only the small widndrose diagram for the basic tooltip?

loicgasser:
yes it is possible to cutomize this the way we want. for instance `extendedHtmlPopup?iframe=true` would only return a subset of the extended tooltip or even a completely new template. making it look nice and fit our current tooltip is far from undoable.

chsdi/templates/htmlpopup/windatlas50.mako
danduk82: @gjn shall I re-sync all profile stuff? Currently only the indexes are there:

du -hs /var/local/profile/swissalti3d/
508K	/var/local/profile/swissalti3d/

gjn:
I would suggest to not change the interface of the services for this.

We can use mako includes on server side to solve this. The windrose part could be mako templated that is included in both base and extended tooltip.

Also, htmlPopup service _has_ to be self contained. It's a service we publish and others can use. It can't and shouldn't depend on what we have in mf-geoadmin3 (as libraries).

chsdi/templates/htmlpopup/windatlas50.mako
ltclm: please take mf-0.dev.bgdi.ch as reference:
```bash
859M    dhm25_25_matrix
102G    swissalti3d
103G    insgesamt
```

ltclm:
here is a [testlink](https://mf-geoadmin3.dev.bgdi.ch/ltclm/?topic=dev&lang=de&bgLayer=ch.swisstopo.pixelkarte-farbe&catalogNodes=9900&layers=ch.bfe.windenergie-geschwindigkeit_h50,ch.bfe.windenergie-geschwindigkeit_h75&X=163000.05&Y=512999.96&zoom=9&layers_visibility=true,false)

with a slightly modified, db-based tooltip.
basic tooltip shows some attributes and small windrose (iframe with extended tooltip)
extended tooltip as specified

Please let me know what you think of this, so that i can add it to the 

the extended tooltip can be called with an additional parameter ``iframe=true`` for a reduced output:
[extended tooltip small](https://mf-chsdi3.dev.bgdi.ch/ltclm/rest/services/dev/MapServer/ch.bfe.windenergie-geschwindigkeit_h50/1462318/extendedHtmlPopup?lang=de&iframe=true)
[extended tooltip normal](https://mf-chsdi3.dev.bgdi.ch/ltclm/rest/services/dev/MapServer/ch.bfe.windenergie-geschwindigkeit_h50/1462318/extendedHtmlPopup?lang=de)

https://github.com/geoadmin/mf-chsdi3/commit/e66971f0ef9c53dc8d82a893ccce484226454c3a
marcel

loicgasser:
version small is nice I would probably put the graph under though.

loicgasser:
I mean the legend

gjn:
Repeated code in the mako for the windrose.

Why not make a mako function for the windrose d3 drawing and call it both in base (in the iframe with smaller graph) and extended (without iframe, bigger graph)?

AFoletti:
I think it's pretty great! Wonderful job.
As for the layout I would not touch it, for 2 reasons:
- it was delivered directly by the customer (script and css. We did not code it)
- I think it looks nice and compact

gjn:
Ah, srcdoc is not supported on IE11, so we can't directly put code in the iframe.

Anyway, we should still try to not repeat 300+ lines of code in the same file.

loicgasser:
yes that's just for testing purposes I guess he'll adapt it for sure.
@ltclm adapt the duplicated code and make a pr in this branch. I'll update the branch to have a nice demo in geoadmin.

chsdi/templates/htmlpopup/windatlas50.mako
gjn: Can't we call the height service? (roundtrip via varnish, I know, but it's fast service)

We could avoid putting 300GB of data on the non-legacy instances.

loicgasser: yes we could but nobody dared to remove it yet. Let's round trip if you prefer.

davidoesch:
For the normal tooltip Legend ID not needed. Can be moved to the large tooltip (expert)

loicgasser:
is now part of https://github.com/geoadmin/mf-chsdi3/pull/1980

PR #1968, opened 2016-03-09
Refactor catalog service and improve coverage

mariavkourouni:
[Test - catalog tree graph](http://mf-chsdi3.dev.bgdi.ch/kom_catalog/rest/services/ech/CatalogServer)

### DIFF
--- a/chsdi/tests/integration/test_catalog.py
+++ b/chsdi/tests/integration/test_catalog.py
@@ -4,22 +4,24 @@ from chsdi.tests.integration import TestsBase
 from chsdi.models.bod import Catalog
 from sqlalchemy.orm import scoped_session, sessionmaker
 from chsdi.views.catalog import create_digraph
+from chsdi.lib.filters import filter_by_geodata_staging
 
 
 class TestCatalogService(TestsBase):
 
     def test_nodes_connection(self):
         try:
+            geodata_staging = self.testapp.app.registry.settings['geodata_staging']
             session = scoped_session(sessionmaker())
             topics = self.testapp.get('/rest/services', status=200)
             for t in topics.json['topics']:
                 topic = t.get('id')
                 query = session.query(Catalog).filter(Catalog.topic == topic)\
-                    .filter(Catalog.staging == 'prod')\
                     .order_by(Catalog.orderKey)
+                query = filter_by_geodata_staging(query, Catalog.staging, geodata_staging)
                 rows = query.all()
                 if (rows):
-                    graph, meta, root_id = create_digraph(rows, 'fr', 'prod')
+                    graph, meta, root_id = create_digraph(rows, 'fr')
                     nodes = graph.nodes()
                     if len(nodes) != len(rows):
                         for row in rows:
@@ -46,9 +48,6 @@ class TestCatalogService(TestsBase):
 
     def test_catalog_wrong_map(self):
         self.testapp.get('/rest/services/foo/CatalogServer', status=400)
-
-    def test_catalog_topic_dev(self):
-        self.testapp.get('/rest/services/dev/CatalogServer', status=500)
 
     def test_catalog_ordering(self):
         resp = self.testapp.get('/rest/services/inspire/CatalogServer', params={'lang': 'en'}, status=200)
--- a/chsdi/tests/integration/test_catalog.py
+++ b/chsdi/tests/integration/test_catalog.py
@@ -1,9 +1,33 @@
 # -*- coding: utf-8 -*-
 
 from chsdi.tests.integration import TestsBase
+from chsdi.models.bod import Catalog
+from sqlalchemy.orm import scoped_session, sessionmaker
+from chsdi.views.catalog import create_digraph
 
 
 class TestCatalogService(TestsBase):
+
+    def test_nodes_connection(self):
+        try:
+            session = scoped_session(sessionmaker())
+            topics = self.testapp.get('/rest/services', status=200)
+            for t in topics.json['topics']:
+                topic = t.get('id')
+                query = session.query(Catalog).filter(Catalog.topic == topic)\
+                    .filter(Catalog.staging == 'prod')\
+                    .order_by(Catalog.orderKey)
+                rows = query.all()
+                if (rows):
+                    graph, meta, root_id = create_digraph(rows, 'fr', 'prod')
+                    nodes = graph.nodes()
+                    if len(nodes) != len(rows):
+                        for row in rows:
+                            if row.id not in nodes:
+                                raise Exception('%s %ss is unconnected leaf' % (row.category, row.layerBodId))
+        finally:
+            if session:
+                session.close()
 
     def test_catalog_no_params(self):
         resp = self.testapp.get('/rest/services/blw/CatalogServer', status=200)
@@ -23,6 +47,9 @@ class TestCatalogService(TestsBase):
     def test_catalog_wrong_map(self):
         self.testapp.get('/rest/services/foo/CatalogServer', status=400)
 
+    def test_catalog_topic_dev(self):
+        self.testapp.get('/rest/services/dev/CatalogServer', status=500)
+
     def test_catalog_ordering(self):
         resp = self.testapp.get('/rest/services/inspire/CatalogServer', params={'lang': 'en'}, status=200)
         self.assertEqual(resp.content_type, 'application/json')
@@ -34,6 +61,10 @@ class TestCatalogService(TestsBase):
             link = '/rest/services/ech/CatalogServer?lang=' + lang
             resp = self.testapp.get(link)
             self.assertEqual(resp.status_int, 200, link)
+
+    def test_layersconfig_with_callback(self):
+        resp = self.testapp.get('/rest/services/blw/MapServer/layersConfig', params={'callback': 'cb'}, status=200)
+        self.assertEqual(resp.content_type, 'application/javascript')
 
     def test_all_catalogs(self):
 
--- a/chsdi/views/catalog.py
+++ b/chsdi/views/catalog.py
@@ -11,7 +11,8 @@ from chsdi.lib.sqlalchemy_customs import remove_accents
 from chsdi.lib.filters import filter_by_geodata_staging
 
 
-# https://github.com/networkx/networkx/blob/master/networkx/readwrite/json_graph/tree.py#L16
+# Custom networkx like function
+# Inspired by https://github.com/networkx/networkx/blob/master/networkx/readwrite/json_graph/tree.py#L16
 def tree_data(G, root, attrs, meta):
     if G.number_of_nodes() != G.number_of_edges() + 1:
         raise TypeError("G is not a tree.")
--- a/chsdi/views/catalog.py
+++ b/chsdi/views/catalog.py
@@ -2,11 +2,81 @@
 
 from pyramid.view import view_config
 from pyramid.httpexceptions import HTTPNotFound
+import networkx as nx
+from itertools import chain
+from pyramid.httpexceptions import HTTPInternalServerError
 
 from chsdi.models.bod import Catalog
 from chsdi.lib.validation import MapNameValidation
 from chsdi.lib.sqlalchemy_customs import remove_accents
 from chsdi.lib.filters import filter_by_geodata_staging
+
+
+# https://github.com/networkx/networkx/blob/master/networkx/readwrite/json_graph/tree.py#L16
+def tree_data(G, root, attrs, meta):
+    if G.number_of_nodes() != G.number_of_edges() + 1:
+        raise TypeError("G is not a tree.")
+    if not G.is_directed():
+        raise TypeError("G is not directed.")
+
+    id_ = attrs['id']
+    children = attrs['children']
+    if id_ == children:
+        raise nx.NetworkXError('Attribute names are not unique.')
+
+    def add_children(n, G):
+        nbrs = G[n]
+        if len(nbrs) == 0:
+            return []
+        children_ = []
+        nbrs = sorted(nbrs)
+        for child in nbrs:
+            d = dict(chain(G.node[child].items(), [(id_, child)]))
+            d['category'] = meta[d['id']]['category']
+            d['staging'] = meta[d['id']]['staging']
+            d['label'] = meta[d['id']]['label']
+            # only for sorting according to 'orderKey'
+            if ('orderKey' in meta[d['id']]):
+                d['orderKey'] = meta[d['id']]['orderKey']
+            if (d['category'] == 'layer'):
+                d['layerBodId'] = meta[d['id']]['layerBodId']
+            else:
+                d['selectedOpen'] = meta[d['id']]['selectedOpen']
+            c = add_children(child, G)
+            if c:
+                d[children] = c
+            children_.append(d)
+            if ('orderKey' in d):
+                order_key = lambda x: x['orderKey']
+                children_ = sorted(children_, key=order_key)
+        for d in children_:
+            d.pop('orderKey', None)
+        return children_
+
+    data = dict(chain(G.node[root].items(), [(id_, root)]))
+    data[children] = add_children(root, G)
+    return data
+
+
+def initialize_graph(G, rows, lang, stage):
+    meta = {}
+    for r in rows:
+        node_parent_id = getattr(r, 'parentId')
+        node_id = getattr(r, 'id')
+        if node_id not in meta:
+            meta[node_id] = r.to_dict(lang)
+        G.add_edge(node_parent_id, node_id)
+        if not node_parent_id:
+            root_id = node_id
+    return G, meta, root_id
+
+
+def create_digraph(rows, lang, stage):
+    G = nx.DiGraph()
+    graph, meta, root_id = initialize_graph(G, rows, lang, stage)
+    edge_list = list(nx.bfs_edges(graph, root_id))
+    G = nx.DiGraph(edge_list)
+    return G, meta, root_id
 
 
 class CatalogService(MapNameValidation):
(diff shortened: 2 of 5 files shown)


chsdi/views/catalog.py
```diff
@@ -87,8 +87,6 @@ def cleanNode(node):
                     path = node['path'].split('/')
                     if len(path) > 0:
                         root = path[0]
-                else:
-                    return nodes_final
```
gjn: Are we sure about this? Why was the code there in the first place? Maybe there is a use-case that we don't see?

loicgasser: I share the same concern that Gilbert. I need proof.

The best way to test that is to extract the code that creates the trees -> functional tests given a list of entries and nodes.
We could probaly rewrite this piece of code as well and use a recursive function.

chsdi/views/catalog.py
```diff
@@ -100,7 +98,7 @@ def cleanNode(node):
                         if idx is not None:
                             try:
                                 nodes_final[root]['children'][idx]['children'].append(node)
-                            except:
+                            except:  # pragma: no cover
```
loicgasser: once code is an function you should be able to test this case as well.

gjn:
@mariavkourouni I'll put this to WIP, but please don't forget about it :)

mariavkourouni:
@gjn ok!

mariavkourouni:
[testlink](http://mf-chsdi3.dev.bgdi.ch/kom_catalog/rest/services/ech/CatalogServer)

loicgasser:
build error is unrelated

chsdi/views/catalog.py
```diff
@@ -2,13 +2,52 @@
...
+import networkx as nx
+from itertools import chain
 
 from chsdi.models.bod import Catalog
 from chsdi.lib.validation import MapNameValidation
 from chsdi.lib.sqlalchemy_customs import remove_accents
 from chsdi.lib.filters import filter_by_geodata_staging
 
 
+def tree_data(G, root, attrs, meta):
```
loicgasser: please put the reference from networkx

chsdi/tests/integration/test_catalog.py
```diff
@@ -26,8 +26,6 @@ def test_catalog_wrong_map(self):
     def test_catalog_ordering(self):
         resp = self.testapp.get('/rest/services/inspire/CatalogServer', params={'lang': 'en'}, status=200)
         self.assertEqual(resp.content_type, 'application/json')
-        self.assertTrue('AGNES' in resp.json['results']['root']['children'][0]['children'][0]['children'][0]['label'])
-        self.assertTrue('Geoid in CH1903' in resp.json['results']['root']['children'][0]['children'][0]['children'][1]['label'])
```
loicgasser: Why?

loicgasser:
It would be nice to have unit tests here don't you think? What is the total coverage on the catalog service now?
Did redeploy the branch for testing purposes?

loicgasser:
sorry I just saw the testlink you provided

mariavkourouni:
the coverage is 96%  lines: 16, 18, 23 (only for the tree_data function)

loicgasser:
You can easily cover the remaining lines with some functional tests. We should be able to detect corrupted graphs as well.
Also could you take into account the `orderKey`?

chsdi/views/catalog.py
```diff
@@ -2,13 +2,52 @@
...
+                d['staging'] = meta[d['id']]['staging']
+                d['label'] = meta[d['id']]['label']
+                if (d['category'] == 'layer'):
+                    d['layerBodId'] = meta[d['id']]['layerBodId']
+                else:
+                    d['selectedOpen'] = meta[d['id']]['selectedOpen']
+            c = add_children(child, G)
+            if c:
+                d[children] = c
+            children_.append(d)
```
loicgasser: here you could either sort the children according to the order key if not None or use the `insert` method of python lists. To insert at specific location.

chsdi/views/catalog.py
```diff
@@ -29,97 +80,13 @@ def catalog(self):
...
-                                nodes_final[root]['children'][idx_1]['children'][idx_2]['children'][idx_3]['children'].append(node)
-                            except:
-                                pass
-
-        return nodes_final
+        lang = self.lang
+        G = nx.DiGraph()
+        graph, meta, root_id = self.initialize_graph(G, rows, lang)
+        edge_list = list(nx.bfs_edges(graph, root_id))
+        Gr = nx.DiGraph(edge_list)
```
loicgasser: You could reuse G as this is one is not needed anymore

loicgasser:
Also with this new awesome imlementation we don't need
depth and path anymore.

So you could remove them from:
https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/models/bod.py#L398

loicgasser:
Jenkins build error is unrelated.
I saw that id is not on category `root` and `catxx` anymore. I think this is a good thing as they are not used in geoadmin anyway.

gjn:
Good effort. I like it!

Any way to easily compare old output with new? Can't afford to 'change' catalogs.

chsdi/views/catalog.py
loicgasser: Also at this point you can check
```python
if len(rows) != len(Gr.nodes()):
    raise HTTPInternalServerError('Catalog tree for topic %s has unconnected leafs' % self.mapName)
```

loicgasser:
A tip, if you create 2 new functions outside the CatalogService class, it might be easier to perform functional tests.

loicgasser:
@gjn 2 main advantages with this approach.

We can detected corrupted graphs
We can create catalogs of virtually infinite depth

gjn:
As said, I like it.

We just need to be able to compare this output with old version for all catalogs.

loicgasser:
yes I was just documenting. ;)

gjn:
@mariavkourouni Can we compare new output with old output?

mariavkourouni:
@gjn @loicgasser 
This script gives as result true for all the topics.
I just pushed it here for now, so you could have a look.

loicgasser:
Did you redploy the branch? Could you provide an online diff between the New and the old version ?

mariavkourouni:
@loicgasser yes, I redeployed.
[diff](https://github.com/geoadmin/bod/pull/9/commits/1a89f02f46a94595628212b5572ccede626f104f)

loicgasser:
@mariavkourouni please remove the unused attributes from the Catalog model. (e.g. path, depth)

loicgasser:
jenkins test

chsdi/views/catalog.py
loicgasser: ping @mariavkourouni

loicgasser:
> Catalog tree for topic inspire has unconnected leafs

That's a good thing to know. Can you tell me which ones?

loicgasser:
ping @mariavkourouni

mariavkourouni:
@loicgasser we have already deleted them together. 
Could you come to me when you have some time?

loicgasser:
jenkins test

loicgasser:
Yes I come in an hour.

mariavkourouni:
@loicgasser 
1) ch.swisstopo.geologie-geotechnik-gk200 (Carte géotechnique 200) and 
2) ch.swisstopo.geologie-geotechnik-mineralische_rohstoffe200 (Mat. premières minérales 200) are the unconnected leafs for inspire.

loicgasser:
https://map.geo.admin.ch/?topic=inspire&lang=fr&bgLayer=ch.swisstopo.pixelkarte-farbe&X=196300.00&Y=652000.00&zoom=3&layers=ch.swisstopo.geologie-geotechnik-gk200,ch.swisstopo.geologie-geotechnik-mineralische_rohstoffe200&layers_opacity=0.75,0.75&catalogNodes=284,165,183

So I checked in the catalog in prod for inspire (see link above) and it seems that the layers are found in the catalog.

Inpire -> Environment, biology and geology -> Geology

Can you have a look directly in the table try to figure out what's the difference?

mariavkourouni:
Both of them are double?

mariavkourouni:
unconnected

loicgasser:
Yes, we need to handle the case were the layers are productive but the category is not.

mariavkourouni:
@loicgasser and how to handle this case? Throw an exception or just ignore the productive layers?

chsdi/tests/integration/test_catalog.py
```diff
@@ -1,10 +1,60 @@
...
+                    .filter(Catalog.staging == 'prod')\
+                    .order_by(Catalog.orderKey)
+                rows = query.all()
+                if (rows):
+                    graph, meta, root_id = create_digraph(rows, 'fr', 'prod')
+                    nodes = graph.nodes()
+                    if len(nodes) != len(rows):
+                        for row in rows:
+                            if row.id not in nodes:
+                                print row.category + ' ' + row.layerBodId + ' is unconnected leaf'
```
mariavkourouni: when we fix the prod layers, raiseException()

mariavkourouni:
@loicgasser @mariokeusen can we change please the staging of the two children layers from prod to test? Otherwise we have unconnected leaves in prod...

ltclm:
we should not change the layers staging. prod layers in dev category have to be ignored. they should be part of  a prod category too, if not theyre not visible in inspire.

mariokeusen:
For the general behavior of the layers I agree with @ltclm : never change a layer's staging outside the data integration process.
For the case with category number 280 there was a wrong staging in the config table (dev instead of prod, no idea why) : I set it on prod and imported the change with the @alcapat 's tool.

mariavkourouni:
Yes, if we change the 280 in prod, we will not have unconnected leaves any more. Thanks!

chsdi/tests/integration/test_catalog.py
```diff
@@ -1,10 +1,60 @@
...
+        with self.assertRaises(TypeError):
+            tree_data(G, 1, {'children': 'children', 'id': 'id'}, meta)
+
+    def test_tree_data_attribute_names(self):
+        G = nx.DiGraph()
+        edge_list = [(15006, 457), (457, 458), (457, 491)]
+        G = nx.DiGraph(edge_list)
+        meta = {457: {u'category': u'cat72', u'staging': u'prod', 'selectedOpen': False, 'label': u'Grundlagen und Planung', u'topic': u'ech', u'depth': 1, 'orderKey': 1, 'parentId': 15006, u'path': u'root/cat72', 'id': 457}, 458: {u'categor    y': u'cat73', u'staging': u'prod', 'selectedOpen': False, 'label': u'Basiskarten', u'topic': u'ech', u'depth': 2, 'orderKey': 1, 'parentId': 457, u'path': u'root/cat72/cat73', 'id': 458}, 491: {u'category': u'cat76', u'staging': u'pr    od', 'selectedOpen': False, 'label': u'Ortsangaben, Referenzsysteme', u'topic': u'ech', u'depth': 2, 'orderKey': 4, 'parentId': 457, u'path': u'root/cat72/cat76', 'id': 491}}
+        with self.assertRaises(nx.NetworkXError):
+            tree_data(G, 15006, {'children': 'children', 'id': 'children'}, meta)
```
gjn: I think the 3 tests above should be functional tests (they don't touch any back-end).

chsdi/views/catalog.py
```diff
@@ -2,13 +2,83 @@
...
+
+
+def create_digraph(rows, lang, stage):
+    G = nx.DiGraph()
+    graph, meta, root_id = initialize_graph(G, rows, lang, stage)
+    edge_list = list(nx.bfs_edges(graph, root_id))
+    G = nx.DiGraph(edge_list)
+    return G, meta, root_id
+
+
```
gjn: Shouldn't we put these graph functions under helper?

chsdi/tests/integration/test_catalog.py
```diff
@@ -23,99 +73,21 @@ def test_catalog_existing_map_no_catalog(self):
...
-        try:
-            topics = self.testapp.get('/rest/services', status=200)
-            for t in topics.json['topics']:
-                topic = t.get('id')
-                # Get catalog
-                catalog = self.testapp.get('/rest/services/' + topic + '/CatalogServer', status=200)
-                # Get flat catalog table entries
-                query = DBSession.query(Catalog).filter(Catalog.topic == topic).filter(Catalog.staging == 'prod')
-                entries = query.all()
-                # Check if every node in the catalog is in view_catalog of db
```
gjn: Do we still tests this? "Check if every node in the catalog is in view_catalog of db"

chsdi/tests/integration/test_catalog.py
```diff
@@ -23,99 +73,21 @@ def test_catalog_existing_map_no_catalog(self):
...
-        self.testapp.app.registry.settings['geodata_staging'] = 'prod'
-        try:
-            topics = self.testapp.get('/rest/services', status=200)
-            for t in topics.json['topics']:
-                topic = t.get('id')
-                # Get catalog
-                catalog = self.testapp.get('/rest/services/' + topic + '/CatalogServer', status=200)
-                # Get LayersConfig for this topic
-                layersconf = self.testapp.get('/rest/services/' + topic + '/MapServer/layersConfig', status=200)
-                # Check if all layers of catalog are in LayersConfig
```
gjn: Do we still test this? "Check if all layers of catalog are in LayersConfig"

chsdi/views/catalog.py
loicgasser: I don't think we'll be using those anywhere else. I would keep them here as it is easier to read and helpers are already quite big.

loicgasser:
If you rebase, make it work with Jenkins and provide a diff between the old and the new service this might enter by the end of the week. Thanks for this tremendous effort.

mariavkourouni:
jenkins

loicgasser:
You can rebase and tests should work now

mariavkourouni:
[Diff between the old and the new catalog service](https://github.com/geoadmin/bod/pull/10/commits/c6c21fd4753912c33087b8a29e6b18104fffc715)

loicgasser:
please check you link

mariavkourouni:
[Diff between the old and the new catalog service](https://github.com/geoadmin/db/pull/2/commits/b14666eb4d5b9d9e881100bc266a30143478eb49)

loicgasser:
@mariavkourouni 

Damn, I get an error when I test locally. probably not your fault. We need to fix the catalog process. I think you can have a look at the gewiss topic.

```
======================================================================
ERROR: test_nodes_connection (chsdi.tests.integration.test_catalog.TestCatalogService)
----------------------------------------------------------------------
  File "/home/ltgal/mf-chsdi3/chsdi/tests/integration/test_catalog.py", line 22, in test_nodes_connection
    graph, meta, root_id = create_digraph(rows, 'fr', 'prod')
  File "/home/ltgal/mf-chsdi3/chsdi/views/catalog.py", line 76, in create_digraph
    graph, meta, root_id = initialize_graph(G, rows, lang, stage)
  File "/home/ltgal/mf-chsdi3/chsdi/views/catalog.py", line 71, in initialize_graph
    return G, meta, root_id
UnboundLocalError: local variable 'root_id' referenced before assignment
-------------------- >> begin captured logging << --------------------
chsdi: DEBUG: route matched for url http://localhost/rest/services; route_name: 'topics', path_info: u'/rest/services', pattern: '/rest/services', matchdict: {}, predicates: ''
chsdi: DEBUG: debug_authorization of url http://localhost/rest/services (view name u'' against context <pyramid.traversal.DefaultRootFactory instance at 0xce507a0>): Allowed (no authorization policy in use)
--------------------- >> end captured logging << ---------------------
```

loicgasser:
Error was related to the dev catalog. It is fixed and good to go! @gjn Please merge if ok with you as well.

chsdi/views/catalog.py
```diff
@@ -1,14 +1,86 @@
...
+from itertools import chain
+from pyramid.httpexceptions import HTTPInternalServerError, HTTPNotFound
 
 from chsdi.models.bod import Catalog
 from chsdi.lib.validation import MapNameValidation
 from chsdi.lib.sqlalchemy_customs import remove_accents
 from chsdi.lib.filters import filter_by_geodata_staging
 
 
+# https://github.com/networkx/networkx/blob/master/networkx/readwrite/json_graph/tree.py#L16
```
gjn: What is the reason to copy code here instead of importing it? The networkx module exists.

Can't we do something like `from networkx.readwrite.json_graph.tree import tree_data`?

@loicgasser

gjn:
Added a comment regarding code re-use.

chsdi/views/catalog.py
loicgasser: It's a custom one

gjn: The comment should mention it. Also, why was a custom one needed? Might it be worth to upstream the adaption?

loicgasser: Have you add a look at add_children function? It is for sorting the children according to orderKey. I don't think I can come up with something generic enough. Note that it is the last function we use to prepare the json.

chsdi/views/catalog.py
gjn: Ok. I think it's worth it to mention it in the code (that code is not 1-1, but inspired...)

loicgasser: done

chsdi/views/catalog.py
```diff
@@ -1,14 +1,87 @@
...
+from itertools import chain
+from pyramid.httpexceptions import HTTPInternalServerError, HTTPNotFound
 
 from chsdi.models.bod import Catalog
 from chsdi.lib.validation import MapNameValidation
 from chsdi.lib.sqlalchemy_customs import remove_accents
 from chsdi.lib.filters import filter_by_geodata_staging
 
 
+# Cutsom networkx like function
```
gjn: Typo. Please adapte and merge when done. Thanks alot for this, @loicgasser and @mariavkourouni .

loicgasser: :see_no_evil:

Issue #1969, opened 2016-03-09
Why are those layers supposed to have searchable features and don't have an index 


loicgasser:
Jenkins fails because of these 2 layers.

ch.swisstopo.geologie-gravimetrischer_atlas.metadata
ch.swisstopo.skitourenkarte-50.metadata

@ltclm @AFoletti @daguer @alcapat Any idea why?

ltclm:
i have set both layers to searchable false in the bod and i have deployed the bod to dev:

**ch.swisstopo.skitourenkarte-50.metadata** is not supposed to be searchable according to ap, i will fix the bod.

**ch.swisstopo.geologie-gravimetrischer_atlas.metadata**  @rfrancel please prepare a sphinx pr with a sphinx pr for this layer. this layer has to be searchable. after the merge of this pr you can set the layer in the bod to searchable againg.

gjn:
They also have the dev wms set. So other tests also failed.

PR #1970, opened 2016-03-09
Round is a resevered namespace

loicgasser:
I don't even know why we have that but it's BAD.

### DIFF
--- a/chsdi/lib/helpers.py
+++ b/chsdi/lib/helpers.py
@@ -107,11 +107,6 @@ def check_even(number):
     return False
 
 
-def round(val):
-    import math
-    return math.floor(val + 0.5)
-
-
 def format_search_text(input_str):
     return remove_accents(
         escape_sphinx_syntax(input_str)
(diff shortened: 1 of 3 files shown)


loicgasser:
Build error related to https://github.com/geoadmin/mf-chsdi3/issues/1969

gjn:
Interesting. If bad, we should remove the function from the helper library.

I've noticed a small difference between in-build round function and helper round.

round(-1.5) = -2.0
helper.round(-1.5) = -1.0

But it's only affecting negative numbers that are exactly -x.5.

loicgasser:
I saw that but I don't see the use case where altitudes would be negative. I believe that this is quite bad indeed.

gjn:
Agree.

Conflicts now with merged PR. Sorry about that.

Should we remove the round function from the helper library?

loicgasser:
yes I rebase now

loicgasser:
done

PR #1971, opened 2016-03-09
Functional tests should always be launched per default

loicgasser:
@gjin using:
```
${NOSE_CMD} chsdi/tests/ -e .*e2e.*
```
or
```
${NOSE_CMD} chsdi/tests/
```

functional tests were never launched

before we had 1609 tests per default
now we have 1676

Jenkins should be adapted once this is merged as I know we're using the first command above.


gjn:
`functional tests were never launched`

I wonder why that is? Any idea?

loicgasser:
I couldn't tell from the doc, I found:

  --cover-inclusive     Include all python files under working directory in
                        coverage report.  Useful for discovering holes in test
                        coverage if not all files are imported by the test

and

  --first-package-wins, --first-pkg-wins, --1st-pkg-wins
                        nose's importer will normally evict a package from
                        sys.modules if it sees a package with the same name in
                        a different location. Set this option to disable that

that may be relevant in nosetests doc

loicgasser:
damn just noticed there was no `__init__.py` in functional folder... So that's the reason. I'll add this file we should be good.

loicgasser:
@gjn I adpated the PR this way we don't have to change Jenkins.

loicgasser:
sphinxapi test is failing do not merge yet

loicgasser:
sphinxapi test uses ```sys.args``` as a result it takes the arguments of the nosestest cmd. I believe these tests were migrated but they initially were meant to be executed outside a testing framework.
This tests only the sphinxapi but it needs to have indexes in the background, that's why I move those in the integration tests.
Sphinxapi python lib is itself subject to no changes and is no more developed.
If we update sphinxserver and new api features come up, we may have to adapt it ourselves, but that's another story.

gjn:
Thanks alot for the explanation!

Looks good to me now.

PR #1974, opened 2016-03-10
Factorize Swissnames models

loicgasser:
I was doing my mvt tiles and corrected that on the way


gjn:
@cedricmoullet @davidoesch This is simply a re-factoring and looks good to me. I tested the affected layer and tooltips are working fine. Creating a testlink would involve some work. Do you need it?

cedricmoullet:
Looks good to me.

PR #1955, opened 2016-03-01
Make use of gatilegrid module

loicgasser:
source code can be found at:

https://github.com/loicgasser/gatilegrid

PR #1980, opened 2016-03-11
Grid Suppport

loicgasser:
https://github.com/geoadmin/mf-chsdi3/pull/1966
https://github.com/geoadmin/mf-chsdi3/pull/1955

@gjn @ltclm  @davidoesch [No DB grid test link](https://mf-geoadmin3.dev.bgdi.ch/gal_grid/?layers=ch.bfe.windenergie-geschwindigkeit_h50&topic=ech&lang=fr&bgLayer=ch.swisstopo.pixelkarte-farbe&X=158070.00&Y=555000.00&zoom=5)

Update from @ltclm for the tooltip. Looks pretty good now.


davidoesch:
Folks absolutely fantastic

to fix before commit

tooltip
- [ ] no legend needed since it is in mouseover/click ( we save a lot of precious space)
- [ ] align Link "zusatzinformationen" to the left
- [ ] The labling has to be multingual ("häufigkeit") or we have to remove it is not possible

Eyperttooltip
- [ ] The labling has to be multingual ("häufigkeit") or we have to remove it is not possible

loicgasser:
hopefully if the extent doesn't change :)

cedricmoullet:
Currently, a tooltip loads 9 js and 2 css. Would it be possible to combine/merge these files ?

loicgasser:
Yes, we could remove some libs we're not using as that are there per default in the extended tooltip.

If you want one file, we would need to make one build per layer.

loicgasser:
This is because of the loaderjs.

loicgasser:
https://github.com/geoadmin/mf-chsdi3/issues/1981

gjn:
Not completley convinced that's it's ideal to load extended tooltip in the basic tooltip in iframe. I think we have alternatives that would also adress the loading of loaderjs (loaderjs is loaded but it's never needed in the base tooltip). We could try either

a) to switch where we have the iframe parameter. Put the iframe=true parameter on the basic tooltip: 
`rest/services/all/MapServer/ch.bfe.windenergie-geschwindigkeit_h50/681_1397/htmlPopup?lang=fr&iframe=true` - then use this in the iframe of the extended tooltip.

b) instead of having iframe=true, we could have iframe service with same URL route as htmlPopup: ``rest/services/all/MapServer/ch.bfe.windenergie-geschwindigkeit_h50/681_1397/htmlIFrame?lang=fr` and using the same mako which delivers just the iframe content (windrose) and this service could then be used in both extended and base tooltip in an iframe. This service would return 404 for layers that don't support it (and it would not be documented)

Thoughts?

loicgasser:
a) In order to add an iframe parameter in the intial request, we'd have to put this info on the client side somewhere in the layersConfig. The added valule is that we can request both, a reqular tooltip without an embedded iframe (never requested in geoadmin though) and an iframe based tooltip. We'd probaly need to document it as people using our tooltips might wonder why we are inconsistent.

b) So you'd have a iframe in both tooltips? I would not put an iframe where we don't need one. Moreover "rose" dimensions varies accoding to where it is inserted (a regular or an extended tooltip). We'd also need to know this info client side as in a). Note that I am not against creating a new service for that.

Finally I personally think the iframe should be based on an extended tooltip template or a dedicated one as an extended tooltip represents an html page (head, body, links to resources etc..) and the regular tooltip is only an html fragment.

Now, for me it comes down to one main question I can't answer alone:

Do we want to generalize this approach and use it for more layers?

If we don't know yet, I'd be in favour of keeping the actual implementation and extend it if we need to, as I don't see any backward compatibility issue if we go either ways.

gjn:
Let's discuss it tomorrow.

ltclm:
@davidoesch what do you mean with that:
>  no legend needed since it is in mouseover/click ( we save a lot of precious space)

davidoesch:
In the standard Tooltip: only Windrose, without the legend
Am 15.03.2016 11:09 vorm. schrieb "Clausen Marcel" <notifications@github.com

ltclm:
We implemented the windrose code that we received from the data owner, it is as specified now.

davidoesch:
@cype To save precious space in the tooltip, I suggest to remove the legend . The legend of the windrose will be visible in the expert tooltip
-> can you adapt the d3js code?
unbenannt

cedricmoullet:
After discussion with @loicgasser,  I don't find particularly clever to add wind rose in the normal tooltip. It's really something for an advanced tooltip.

davidoesch:
A graph/Image itself makes sense , as in Watertemp station. show key values for the specific location.
form a tech point of view: what are the issues?

gjn:
Creating all the images is the issue. We would need to create it for each cell, as the data.json file.
Or we create it on the fly on the server.

gjn:
That being said, I don't think is bad to have an iframe in the normal tooltip and to present the windrose like this.

The questions is more about responsibility of this code and updates/maintencance/tests.

loicgasser:
@gjn 
Please review you'll need the boto credentials in your env. Ask Andrea.

gjn:
I'll do that.

chsdi/views/features.py
```diff
@@ -291,10 +291,7 @@ def _identify_grid(params, layerBodIds):
 
     pointCoordinates = geometry.coordinates
 
-    # TODO change bucket and profile names
-    # To work with pserve use your own profilename
-    testProfilename = params.request.registry.settings['boto_test_profilename']
-    bucket = get_bucket(profile_name=testProfilename, bucket_name='waf-wmts-test')
+    bucket = get_bucket(profile_name='chsdi_aws_admin_ro', bucket_name='mwks6dv2y5dsbbgg-vectortiles')
```
gjn: @danduk82 @loicgasser The idea is to dispatch this profile on all our mf0 machines?

loicgasser: According to Jira ticket yes

chsdi/lib/helpers.py
```diff
@@ -289,15 +292,29 @@ def parse_box2d(stringBox2D):
...
-def center_from_box2d(box2D):
+def is_box2d(box2D):
     # Bottom left to top right only
-    if box2D[0] > box2D[2] or box2D[1] > box2D[3]:
+    if box2D[0] > box2D[2] or box2D[1] > box2D[3] or len(box2D) != 4:
         raise ValueError('Invalid box2D.')
+    return True
+
+
+def extend_box2d(box2D, distance):
```
gjn: I think this function is never used.

chsdi/lib/helpers.py
```diff
@@ -289,15 +292,29 @@ def parse_box2d(stringBox2D):
...
+            box2D[3] + distance
+        ]
 
-    return [
-        box2D[0] + ((box2D[2] - box2D[0]) / 2),
-        box2D[1] + ((box2D[3] - box2D[1]) / 2)
-    ]
+
+def center_from_box2d(box2D):
+    if is_box2d(box2D):
```
gjn: In case if this being False, client code gets None and will probably fail, is it?

chsdi/models/bod.py
```diff
@@ -320,63 +317,46 @@ class ServiceMetadataEn(Base, ServiceMetadata):
...
-                {'level': 21, 'resolution': 5, 'scale': 17857.1882144, 'width': 375, 'height': 250},
-                {'level': 22, 'resolution': 2.5, 'scale': 8928.59410719, 'width': 750, 'height': 500},
-                {'level': 23, 'resolution': 2, 'scale': 7142.87528575, 'width': 938, 'height': 625},
-                {'level': 24, 'resolution': 1.5, 'scale': 5357.15646431, 'width': 1250, 'height': 834},
-                {'level': 25, 'resolution': 1, 'scale': 3571.43764288, 'width': 1875, 'height': 1250},
-                {'level': 26, 'resolution': 0.5, 'scale': 1785.71882144, 'width': 3750, 'height': 2500},
-                {'level': 27, 'resolution': 0.25, 'scale': 892.857, 'width': 7500, 'height': 5000},
-                {'level': 28, 'resolution': 0.1, 'scale': 357.1425, 'width': 18750, 'height': 12500}
-            ]
+            'compressionQuality': None,
```
gjn: The new lods are different than the old ones. Differences are in the 'scale' value only. When I create them as above with gaTileGrid, I get the following:

{'height': 1, 'width': 1, 'scale': 15118080.0, 'resolution': 4000.0, 'level': 0}
{'height': 1, 'width': 1, 'scale': 14173200.0, 'resolution': 3750.0, 'level': 1}
{'height': 1, 'width': 1, 'scale': 13228320.0, 'resolution': 3500.0, 'level': 2}
{'height': 1, 'width': 1, 'scale': 12283439.999999998, 'resolution': 3250.0, 'level': 3}
{'height': 1, 'width': 1, 'scale': 11338559.999999998, 'resolution': 3000.0, 'level': 4}
{'height': 1, 'width': 1, 'scale': 10393680.0, 'resolution': 2750.0, 'level': 5}
{'height': 1, 'width': 1, 'scale': 9448800.0, 'resolution': 2500.0, 'level': 6}
{'height': 1, 'width': 1, 'scale': 8503920.0, 'resolution': 2250.0, 'level': 7}
{'height': 1, 'width': 1, 'scale': 7559040.0, 'resolution': 2000.0, 'level': 8}
{'height': 1, 'width': 2, 'scale': 6614160.0, 'resolution': 1750.0, 'level': 9}
{'height': 1, 'width': 2, 'scale': 5669279.999999999, 'resolution': 1500.0, 'level': 10}
{'height': 1, 'width': 2, 'scale': 4724400.0, 'resolution': 1250.0, 'level': 11}
{'height': 2, 'width': 2, 'scale': 3779520.0, 'resolution': 1000.0, 'level': 12}
{'height': 2, 'width': 3, 'scale': 2834639.9999999995, 'resolution': 750.0, 'level': 13}
{'height': 2, 'width': 3, 'scale': 2456688.0, 'resolution': 650.0, 'level': 14}
{'height': 3, 'width': 4, 'scale': 1889760.0, 'resolution': 500.0, 'level': 15}
{'height': 5, 'width': 8, 'scale': 944880.0, 'resolution': 250.0, 'level': 16}
{'height': 13, 'width': 19, 'scale': 377951.99999999994, 'resolution': 100.0, 'level': 17}
{'height': 25, 'width': 38, 'scale': 188975.99999999997, 'resolution': 50.0, 'level': 18}
{'height': 63, 'width': 94, 'scale': 75590.4, 'resolution': 20.0, 'level': 19}
{'height': 125, 'width': 188, 'scale': 37795.2, 'resolution': 10.0, 'level': 20}
{'height': 250, 'width': 375, 'scale': 18897.6, 'resolution': 5.0, 'level': 21}
{'height': 500, 'width': 750, 'scale': 9448.8, 'resolution': 2.5, 'level': 22}
{'height': 625, 'width': 938, 'scale': 7559.039999999999, 'resolution': 2.0, 'level': 23}
{'height': 834, 'width': 1250, 'scale': 5669.279999999999, 'resolution': 1.5, 'level': 24}
{'height': 1250, 'width': 1875, 'scale': 3779.5199999999995, 'resolution': 1.0, 'level': 25}
{'height': 2500, 'width': 3750, 'scale': 1889.7599999999998, 'resolution': 0.5, 'level': 26}
{'height': 5000, 'width': 7500, 'scale': 944.8799999999999, 'resolution': 0.25, 'level': 27}
{'height': 12500, 'width': 18750, 'scale': 377.952, 'resolution': 0.1, 'level': 28}

getScale of gaTileGrid is also used in the WMTSGetCapabilities documents - so we really need to be sure with this.

chsdi/models/bod.py
```diff
@@ -320,63 +317,46 @@ class ServiceMetadataEn(Base, ServiceMetadata):
...
         'description': 'Configuration for the map (topic) ' + mapName,
         'copyrightText': 'Data ' + mapName,
         'layers': [],
-        'spatialReference': {"wkid": 21781},
+        'spatialReference': {"wkid": gagrid.spatialReference},
         'tileInfo': {
-            'rows': 256,  # tile width in pixel
-            'cols': 256,
+            'rows': gagrid.tileSizePx,  # tile width in pixel
+            'cols': gagrid.tileSizePx,
```
gjn: Returns a flow, before it was an int. Impact in front-end?

chsdi/templates/htmlpopup/kgs.mako
```diff
@@ -12,6 +12,9 @@
 </%def>
 
 <%def name="extended_info(c, lang)">
+    <link rel="stylesheet" type="text/css" href="${h.versioned(request.static_url('chsdi:static/css/blueimp-gallery'))}"/>
+    <script src="${h.versioned(request.static_url('chsdi:static/js/jquery-2.0.3.min.js'))}"></script>
```
loicgasser: I need to fix those links

chsdi/models/clientdata_dynamodb.py
```diff
@@ -73,7 +73,6 @@ def get(self):
 
 
 dynamodb_connection = DynamodbConnection()
-s3_connection = S3Connection()
```
gjn: We shared connections accross requests with this approach (at least, that' what I assume).

Now we don't anymore. What is the impact of that?

Might we share connections accross requests for each profile?

chsdi/templates/htmlpopup/energieforschung.mako
```diff
@@ -43,7 +43,9 @@ unterthema = 'unterthema_%s' %lang
 date_beginn = datetime.datetime.strptime(c['attributes']['projektbeginn'].strip(), "%Y-%m-%d").strftime("%d.%m.%Y")
 date_ende = datetime.datetime.strptime(c['attributes']['projektende'].strip(), "%Y-%m-%d").strftime("%d.%m.%Y")
 %>
-
+<link rel="stylesheet" type="text/css" href="${h.versioned(request.static_url('chsdi:static/css/blueimp-gallery.min.css'))}"/>
+<script src="${h.versioned(request.static_url('chsdi:static/js/jquery.min.js'))}"></script>
+<script src="${h.versioned(request.static_url('chsdi:static/js/blueimp-gallery.min.js'))}"></script>
```
gjn: We need to test all affected makos.

chsdi/templates/htmlpopup/windatlas50.mako
```diff
@@ -0,0 +1,1003 @@
+<%inherit file="base.mako"/>
+<%!
+from chsdi.lib.raster.georaster import get_raster
+
+r = get_raster('COMB')
```
gjn: I'd rather use the official service here (yes, means roundtrip...). But having it here measn having dtm data on all instances that we have this popup. This is officially not the case on our production instances, where we have this data only for the profile/height service instances (it's separated). And in the future, this service might be completely separated from chsdi3. In that moment, we have to use the service any way.

It's better to decouple this now.

chsdi/models/bod.py
loicgasser: scale value was slightly wrong before

chsdi/templates/htmlpopup/windatlas50.mako
```diff
@@ -0,0 +1,1003 @@
...
+props = c['properties']
+
+freq_total = props['freq_0']    + props['freq_30']
+freq_total += props['freq_60']  + props['freq_90']
+freq_total += props['freq_120'] + props['freq_150']
+freq_total += props['freq_180'] + props['freq_210']
+freq_total += props['freq_240'] + props['freq_270']
+freq_total += props['freq_300'] + props['freq_330']
+
+# gracefully check if url request has get parameter iframe
```
gjn: We need to re-think our iframe approach. Seems it will be needed with other popups as well (see https://github.com/geoadmin/mf-chsdi3/pull/1996) - and therefore it merits a more generic approach.

I envision an undocumented `iframe` service, that we can use to include in the standard html and extended html popups.

We would have:
`https://api3.geo.admin.ch/rest/services/api/MapServer/ch.bfe.windenergie-geschwindigkeit_h/1/iframe?content=windrose`
It's a full html page that can be parametrized. Each layer decideds if it needs one or not.

which then can be used inside
`https://api3.geo.admin.ch/rest/services/api/MapServer/ch.bfe.windenergie-geschwindigkeit_h/1/htmlPopup`
and
`https://api3.geo.admin.ch/rest/services/api/MapServer/ch.bfe.windenergie-geschwindigkeit_h/1/extendedHtmlPopup`

We can discuss it later.

chsdi/views/features.py
```diff
@@ -214,17 +218,14 @@ def insertTimestamps(header, comments):
...
-    )
+    comments = render('chsdi:templates/oereb_timestamps.mako', {
+        'data_imported': data_imported,
+        'data_created': data_created
+    })
     header = insertTimestamps(header, comments)
 
     # Only relation 1 to 1 is needed at the moment
-    layerVectorModel = [[oereb_models_from_bodid(layerBodId)[0]]]
+    layerVectorModel = [{layerBodId: [oereb_models_from_bodid(layerBodId)[0]]}]
```
gjn: Why this change?

chsdi/models/bod.py
loicgasser: https://github.com/loicgasser/gatilegrid/blob/master/gatilegrid/__init__.py#L141 for reference. I pretty sure it's correct

chsdi/models/clientdata_dynamodb.py
loicgasser: A connection is initialized only when we do:
```
s3_connection.get()
```

We use to share only the constructor to get those connections. Now that we have several possible connections on s3 we create a new object each time.

chsdi/templates/htmlpopup/windatlas50.mako
loicgasser: yes

chsdi/models/bod.py
gjn: Can we find out where those changed scale values are used? (All services that are affected).

Then we need to analyse the external effects. It might be that it doesn't matter at all - but I'd rather be sure.

chsdi/lib/helpers.py
loicgasser: Yes that's was my idea. This way we can handle the error in the service.

```python
if center_from_box2d(box2D) is None:
    raise HTTPBadRequest()
```

chsdi/models/bod.py
loicgasser: None, this service is not used. But I could use a cast here.

chsdi/models/bod.py
loicgasser: Assumed dpi in current WMTS is currently
`dpi = scale / (resoultion * pixelPerinches) = 13392891.1608 / (3750.0 * 39.37) = 90.71469755842011`
It's correct in the doc though. http://api3.geo.admin.ch/services/sdiservices.html#wmts

ltclm: ``90.7``  comes from here:
http://gis.stackexchange.com/questions/29671/mathematics-behind-converting-scale-to-resolution

chsdi/lib/helpers.py
gjn: but this is not done when we use `center_from_box2d` in code currently.

chsdi/models/bod.py
gjn: In the above link, it says that WMTS OGC specifications says to use 90.7...(corresponding to 0.28mm pixel size). From the OGC WMTS Implementation standard:

```
The scale denominator is defined with respect to a "standardized rendering pixel size" of
0.28 mm × 0.28 mm (millimeters). The definition is the same used in WMS 1.3.0 [OGC
06-042] and in Symbology Encoding Implementation Specification 1.1.0 [05-077r4].
Frequently, the true pixel size is unknown and 0.28 mm is a common actual size for
current displays.
```

So I think, for WMTS at least, we should continue to use 90.7. In the other services, we also publish 96dpi value and we can use the new onces.

Thoughts?

loicgasser: Let's keep on using 90.7, doc is correct as it states at 96.0 dpi. I'll adapt.

chsdi/lib/helpers.py
loicgasser: notice the use of ```is_box2d```, it raises

```ValueError('Invalid box2D.')```

if this is not a box2d.

File: chsdi/views/features.py (hunk 11 above)
Comment (loicgasser) [inline chsdi/views/features.py]: data structure changed in `_get_features_for_filters`

File: chsdi/templates/htmlpopup/windatlas50.mako (hunk 10 above)
Comment (loicgasser) [inline chsdi/templates/htmlpopup/windatlas50.mako]: Yes after this PR. It's already a lot of changes without adding a new service. But I like the idea.

**Comment by loicgasser** [comment]
S3 credentials ok now.

**Comment by gjn** [comment]
Should I rebase?

**Comment by loicgasser** [comment]
@ltclm your geojson are wrong i believe. Polygons must be an array of array of arrays.
Here we have arrays of arrays... Why has this changed?
http://geojson.org/geojson-spec.html#polygon

**Comment by loicgasser** [comment]
https://mf-chsdi3.dev.bgdi.ch/gal_gatilegrid_grid_support/rest/services/ech/MapServer/identify?geometry=559370,162609.99999999997&geometryFormat=geojson&geometryType=esriGeometryPoint&imageDisplay=959,1139,96&layers=all:ch.bfe.windenergie-geschwindigkeit_h50&mapExtent=545410,146680,564590,169460&returnGeometry=true&tolerance=10

**Comment by loicgasser** [comment]
@gjn I just did, I'll push making sure everything is ok.

**Comment by loicgasser** [comment]
@gjn geoms have been updated. You can test using the credentials in /home/ltgal/zoro

**Comment by gjn** [comment]
Do I really need those credentials? Aren't they in global boto config?

**Comment by loicgasser** [comment]
If you want to test using pserve

File: chsdi/lib/helpers.py (hunk 2 above)
Comment (gjn) [inline chsdi/lib/helpers.py]: @loicgasser Please remove this function as it's never used.

**Comment by loicgasser** [comment]
@ltclm I forgot to take you changes before rebasing which means that I lost you commits


 @ltclm	new coordinate source due to fixed geojson structure			bb222d1
 @ltclm	simplify windrose in basic tooltip			1918dce
 @ltclm	add windrose translation			cc292ff
 @ltclm	left-align windrose in basic tooltip			b998894


Could you do the following command

```
git checkout gal_gatilegrid_grid_support
git checkout -b mybackup
git checkout fetch origin gal_gatilegrid_grid_support:gal_gatilegrid_grid_support
git checkout gal_gatilegrid_grid_support
git cherry-pick bb222d1
git cherry-pick 1918dce
git cherry-pick cc292ff
git cherry-pick b998894
git push origin gal_gatilegrid_grid_support
```
Would be great sorry

File: chsdi/lib/helpers.py (hunk 3 above)
Comment (gjn) [inline chsdi/lib/helpers.py]: Well, this has been like this before this PR. But the ValueError raised from is_box2d is not handled in code. see [1] and [2]. Here's an example link of what I mean: http://mf-chsdi3.dev.bgdi.ch/gal_gatilegrid_grid_support//rest/services/inspire/SearchServer?searchText=berges&type=locations&bbox=551754.125,167918.328125,551306.5625,168514.625

It should fail not with internal server error, but with proper 404 error (bad parameter). But this could be done in separate PR.

[1] https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/views/historicalmaps.py#L31
[2] https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/views/search.py#L274

**Comment by loicgasser** [comment]
I think

```
git cherry-pick mybackup~3 mybackup~2 mybackup~1 mybackup
```

should work as well

ltclm:
done

loicgasser:
yeah! thanks and sorry for that

loicgasser:
@davidoesch Let's remove the rose on the left it is really not nice on mobile.
screenshot_2016-04-07-21-10-16

ltclm:
Windenergieanlagen tooltip is not working:
https://mf-geoadmin3.dev.bgdi.ch/gal_grid/?layers=ch.bfe.windenergie-geschwindigkeit_h100,ch.bfe.windenergieanlagen&topic=ech&lang=de&bgLayer=ch.swisstopo.pixelkarte-farbe&X=225602.10&Y=568063.27&zoom=8&layers_visibility=false,true


https://mf-chsdi3.dev.bgdi.ch/gal_gatilegrid_grid_support/1457537727/rest/services/ech/MapServer/ch.bfe.windenergieanlagen/turbine_56/htmlPopup?coord=568233.75,225631.24999999997&imageDisplay=977,805,96&lang=de&mapExtent=566842.0195655648,224595.85434745927,569284.5195655648,226608.35434745927

loicgasser:
Yes thanks for the notice. I ll fix that and try to know why our tests did not detect it.

loicgasser:
This was the problem: https://github.com/geoadmin/mf-chsdi3/pull/1980/commits/4e8d8fe65bda71b988e0cdd891836e98f1846251

davidoesch:
Hi regarding the position of the rose text and image: put the image below the rose text , in a new row and merge the to cols put the  image as left align in this cell. Got the OK for this from the dataowner

loicgasser:
ok thanks for the feedback will do

loicgasser:
Rebased, redeployed and here are the 5 windenergie layers:

[ch.bfe.windenergie-geschwindigkeit_h50](https://mf-geoadmin3.dev.bgdi.ch/gal_grid/?layers=ch.bfe.windenergie-geschwindigkeit_h50&topic=ech&lang=fr&bgLayer=ch.swisstopo.pixelkarte-farbe&X=158070.00&Y=555000.00&zoom=5)
[ch.bfe.windenergie-geschwindigkeit_h75](https://mf-geoadmin3.dev.bgdi.ch/gal_grid/?layers=ch.bfe.windenergie-geschwindigkeit_h75&topic=ech&lang=fr&bgLayer=ch.swisstopo.pixelkarte-farbe&X=158070.00&Y=555000.00&zoom=5)
[ch.bfe.windenergie-geschwindigkeit_h100](https://mf-geoadmin3.dev.bgdi.ch/gal_grid/?layers=ch.bfe.windenergie-geschwindigkeit_h100&topic=ech&lang=fr&bgLayer=ch.swisstopo.pixelkarte-farbe&X=158070.00&Y=555000.00&zoom=5)
[ch.bfe.windenergie-geschwindigkeit_h125](https://mf-geoadmin3.dev.bgdi.ch/gal_grid/?layers=ch.bfe.windenergie-geschwindigkeit_h125&topic=ech&lang=fr&bgLayer=ch.swisstopo.pixelkarte-farbe&X=158070.00&Y=555000.00&zoom=5)
[ch.bfe.windenergie-geschwindigkeit_h150](https://mf-geoadmin3.dev.bgdi.ch/gal_grid/?layers=ch.bfe.windenergie-geschwindigkeit_h150&topic=ech&lang=fr&bgLayer=ch.swisstopo.pixelkarte-farbe&X=158070.00&Y=555000.00&zoom=5)

davidoesch:
@loicgasser : 
[ ] make a left align of the gfx 
[ ] make non scroll size, 
[ ] the circles in the windrose should be behind the sectors (line of circles are covered by the sector according to @cype (dataowner)

loicgasser:
Left align OK no problem
Tooltip max height content is defined in geoadmin but I guess we can hack it...
Concerning circles -> yes we can and will

ltclm:
we received this windrose definition for the basic tooltip:
http://www.bfe-gis.admin.ch/testumgebung/wind/windrose_klein_ohne_legende.php

loicgasser:
@davidoesch 
> make non scroll size

It's a mess not that it is impossible to do, but you have to take into account all the use cases.
Mobile phones, tablets, selecting several features from different layers etc...
Moreover it is a dirty hack to override css in geoadmin via style tags in html fragments.
We now have some space on the right of the rose, I propose to put the link to object and link to extended tooltip there.
Everything else has been taken care of.

loicgasser:
jenkins test

davidoesch:
Keep it simple. If it will be maintenance hell ,- just let the scroll bar be scroll bar

loicgasser:
yes thank you for you understanding.

gjn:
@loicgasser @ltclm Congratulations for this work! Impressive and I think an appropriate solution for this kind of data.

I finally merge.

PR #1982, opened 2016-03-15
releasenotes 20160316

ltclm:


gjn:
@davidoesch @cedricmoullet Please validate change in api examples. We removed the self-hosted examples and links now lead directly to code-pen.

See http://mf-chsdi3.dev.bgdi.ch/api/examples.html

If good, we can merge the release notes.

cedricmoullet:
Shouldn't we add "Examples migrated to CodePen" ?

cedricmoullet:
With this example: http://codepen.io/geoadmin/pen/xVKLdV?editors=0010
image
Quality is not the same on big scale.

cedricmoullet:
In http://codepen.io/geoadmin/pen/NxQmvo?editors=0010 If a layer is added (1:25'000) for example, a rectangle can be drawn, but it doesn't appear. Probably a zindex issue.

gjn:
1) I'll add the change to the release notes
2) Cadastral Example: mapproxy configuration and nature of the beast. Same on old examples
3) Bug in codepen examples (was working well in old examples). I'll open an issue. Don't think it blocks the deploy.

loicgasser:
https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/static/doc/examples/README.md
To adapt codepens

loicgasser:
release notes updated

PR #1984, opened 2016-03-15
Adapt deployment

gjn:
This PR adds 2 new integration target machines.

It also assures that we don't have the 'varnish looses back-ends' effect described in comments in https://github.com/geoadmin/mf-geoadmin3/pull/3200 during the deploy.

We had this since forever, but it never really mattered because deployment was so slow. Now with a) make instead of buildout and b) reduced transfer of files to targets because of removed .git repository, we are much faster. This started to show the bad effect on deployment:

image

### DIFF
--- a/deploy/deploy.cfg
+++ b/deploy/deploy.cfg
@@ -21,7 +21,10 @@ content = Include /var/www/vhosts/mf-chsdi3/private/chsdi/apache/*.conf
 [remote_hosts]
 # mf0i
 int = ip-10-220-6-119.eu-west-1.compute.internal,
-      ip-10-220-6-105.eu-west-1.compute.internal
+      ip-10-220-6-41.eu-west-1.compute.internal,
+      ip-10-220-6-105.eu-west-1.compute.internal,
+      ip-10-220-6-208.eu-west-1.compute.internal
+
 
 # mf0 --> vpc_mf0_legacy_prod and vpc_mf0_prod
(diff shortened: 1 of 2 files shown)


gjn:
@davidoesch @cedricmoullet This is the same fix as https://github.com/geoadmin/mf-geoadmin3/pull/3200 for geoadmin3. It assures that varnish has enough time to detect that the instance is up again.

cedricmoullet:
Just for my understanding: why do we still need to deploy to legacy prod ?
(sorry, my Alzheimer...)

2016-03-16 7:25 GMT+01:00 Gilbert Jeiziner <notifications@github.com>:

> @davidoesch <https://github.com/davidoesch> @cedricmoullet
> <https://github.com/cedricmoullet> This is the same fix as
> geoadmin/mf-geoadmin3#3200
> <https://github.com/geoadmin/mf-geoadmin3/pull/3200> for geoadmin3. It
> assures that varnish has enough time to detect that the instance is up
> again.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/geoadmin/mf-chsdi3/pull/1984#issuecomment-197170425>
>



-- 

davidoesch:
I like the idea of 2 new target machines

gjn:
We have print and profile separated on those instances.

The new instance are integration only, as we only had 1 per type now - which was bad.

PR #1987, opened 2016-03-16
Use the same version on int and prod

loicgasser:
This has always been this way, but we can improve it.


gjn:
Agree. It should only be first step. We should have same mechanisms as in ga3 for the deploysnapshot: build it for the target environment with new version in the snapshot directory, transfer the files, rebuild only what's strictly needed on the targets. This is a good first step.

@cedricmoullet @davidoesch Internal changes that won't affect the app and affects the deploy. If you agree, we merge it.

davidoesch:
Make sense -ok

PR #1995, opened 2016-03-21
Deflate kml files

procrastinatio:
KML files on the vhosts, inlcudings thoses dezipped though the `ogcproxy`should be gezipped for performance sake.


gjn:
How does it perform with 20Mb file? (Our limit in map.geo.admin.ch). Metrics (Both compressing in apache and uncrompressing in browser) would be nice.

gjn:
Ping @procrastinatio

procrastinatio:
I would close this PR, because KMZ are hardly used (only 0.1% KMLs are KMZ), and more this will do nothing as the decompression is handled by python not apache.

gjn:
ok.

PR #1997, opened 2016-03-22
[WIP] releasenotes 20160330

ltclm:

PR #1999, opened 2016-03-24
Release note for next deploy 20160330

pfanguin:


AFoletti:
that's a duplicate of https://github.com/geoadmin/mf-chsdi3/pull/1997
@pfanguin could you please integrate your changes in said PR?

gjn:
Do we curently have a WMS layer in production with similar number of objects and similar symbology that we could use as comparison? Is bazl luftfahrthindernisse what we can expect?

AFoletti:
this PR is no more needed

PR #2000, opened 2016-03-24
haltestellen introduce polygon geometry for tootip intersection

ltclm:
this is a postgis based approach to solve the issues differences between label extent and feature geometry extent. I'm not convinced that vector tiles will help anything here https://github.com/geoadmin/mf-geoadmin3/issues/3065.

[Testlink]( https://mf-geoadmin3.dev.bgdi.ch/ltclm/?topic=ech&lang=de&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.bav.haltestellen-oev&X=247821.90&Y=682731.97&zoom=12 )

The new polygon geometry is generated on the fly based on the attributes ``verkehrsmittel`` and ``betriebspunkttyp``

The new db view bav.oev_haltestellen_tooltip can be reviewed here:
https://github.com/geoadmin/db/commit/dd65846e57816fcd93a8733273fe0e989d9ec5fc

### DIFF
--- a/chsdi/models/vector/uvek.py
+++ b/chsdi/models/vector/uvek.py
@@ -14,20 +14,52 @@ Base = bases['uvek']
 
 
 class OevHaltestellen (Base, Vector):
-    __tablename__ = 'oev_haltestellen'
+    __tablename__ = 'oev_haltestellen_tooltip'
     __table_args__ = ({'schema': 'bav', 'autoload': False})
     __template__ = 'templates/htmlpopup/oev_haltestellen.mako'
     __bodId__ = 'ch.bav.haltestellen-oev'
     __label__ = 'name'
     __extended_info__ = True
     __queryable_attributes__ = ['id', 'name']
+    __returnedGeometry__ = 'the_geom_point'
     id = Column('nummer', Integer, primary_key=True)
     name = Column('name', Text)
     tuabkuerzung = Column('tuabkuerzung', Text)
     betriebspunkttyp = Column('betriebspunkttyp', Text)
     verkehrsmittel = Column('verkehrsmittel', Text)
-    the_geom = Column(Geometry(geometry_type='GEOMETRY',
+    # polygon geometry spatial selection
+    the_geom = Column('bgdi_geom_poly', Geometry(geometry_type='GEOMETRY',
                                dimension=2, srid=21781))
+    # point geometry hilight
+    the_geom_point = Column('the_geom', Geometry(geometry_type='GEOMETRY',
+                                                               dimension=2, srid=21781))
+    __minscale__ = 1
+    __maxscale__ = 3000
+
+register('ch.bav.haltestellen-oev', OevHaltestellen)
+
+
+class OevHaltestellen (Base, Vector):
+    __tablename__ = 'oev_haltestellen_tooltip'
+    __table_args__ = ({'schema': 'bav', 'autoload': False,  'extend_existing': True})
+    __template__ = 'templates/htmlpopup/oev_haltestellen.mako'
+    __bodId__ = 'ch.bav.haltestellen-oev'
+    __label__ = 'name'
+    __extended_info__ = True
+    __queryable_attributes__ = ['id', 'name']
+    __returnedGeometry__ = 'the_geom_point'
+    id = Column('nummer', Integer, primary_key=True)
+    name = Column('name', Text)
+    tuabkuerzung = Column('tuabkuerzung', Text)
+    betriebspunkttyp = Column('betriebspunkttyp', Text)
+    verkehrsmittel = Column('verkehrsmittel', Text)
+    # polygon geometry spatial selection
+    the_geom = Column('bgdi_geom_poly_overview', Geometry(geometry_type='GEOMETRY',
+                               dimension=2, srid=21781))
+    # point geometry hilight
+    the_geom_point = Column('the_geom', Geometry(geometry_type='GEOMETRY',
+                                                               dimension=2, srid=21781))
+    __minscale__ = 3000
 
 register('ch.bav.haltestellen-oev', OevHaltestellen)
 
--- a/chsdi/models/vector/uvek.py
+++ b/chsdi/models/vector/uvek.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from sqlalchemy import Column, Text, Integer, Date, DateTime, ForeignKey
-from sqlalchemy.orm import relationship
+from sqlalchemy.orm import relationship, remote
 from sqlalchemy.types import Numeric
 from geoalchemy2.types import Geometry
 from sqlalchemy.dialects import postgresql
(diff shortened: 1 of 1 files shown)


ltclm:
@davidoesch please test here

davidoesch:
+1 tested on chrome FF and IE11: it works like a charm and mobile
-> perfect solution

gjn:
Special treatement for each layer. It's maintenance heavy and will need attention for every layer on every update. It needs to be integrated in DI process.

@ltclm @AFoletti feasable?

Maybe we add the as temporary fix for this layer - and wait until we have a solution with vector tiles and then compare results and effort required for each layer to maintain it.

ltclm:
there is only an "issue" when the extent of the label is not identical with the extent of the feature geometry. so manly point layers are affected, need some special treatement, there is no need to have this for all our layers.
I dont think that vector tiles would fix this particular problem.

AFoletti:
should be updates-transparent since the new geometry is prepared by the view itself. It's a very nice solution for the present case and can help us handle borderline layers where point symbolization is big. Another candidate is luftfahrthindernis

gjn:
I like the solution and the effect. Much improved.

I just want to avoid adding unnecessary complexity that increases our maintenance and testing tasks. Just adds another element to always consider. And views have their difficulty :=)

Speaking of tests: would be great to add some automated tests for this!

It's not only point layers, but also polyline layers, depending on how they are styled (I seem to remember similar troubles with IVS layers).

I've nothing against merging this, but I still hope that vector tiles would handle this better. @ltclm I wonder why you think vector tiles won't be able to handle this? Could you elaborate?

gjn:
If I understand correctly, when we change the symbology, we need to adapt the view. Correct?

ltclm:
yes if we change the icon size in mapserver we have to update the view definition too.

im not 100% sure how we will use vector tiles here.
if you will use the vector tiles to replace the source of the hover effect (hand), then i think it will improve the usability. because then we will use the original feature geomerty for hover. we will see... vector tiles will also introduce a lot of special treatements (symbology etc)...

gjn:
Understood.

The idea with vector tiles is to use the symbol for detection - and each feature will directly have it's id as part of the transfered vector data. We don't even need an identify request to find out which elements are under a cursor. That's the theory.

Of course, vector tiles adds a whole lot of complexity on it's own.

gjn:
@loicgasser Would like to have your opinion. To me, this could be merged for tomorrow.

chsdi/models/vector/uvek.py
```diff
@@ -14,20 +14,52 @@
...
+    # point geometry hilight
+    the_geom_point = Column('the_geom', Geometry(geometry_type='GEOMETRY',
+                                                               dimension=2, srid=21781))
+    __minscale__ = 1
+    __maxscale__ = 3000
+
+register('ch.bav.haltestellen-oev', OevHaltestellen)
+
+
+class OevHaltestellen (Base, Vector):
```
loicgasser: Use a different class name here.

chsdi/models/vector/uvek.py
```diff
@@ -14,20 +14,52 @@
 
 
 class OevHaltestellen (Base, Vector):
```
loicgasser: please factorize, have a look at:
https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/models/vector/stopo.py#L175
for instance

loicgasser:
not against that, but how does the identify request performs as compared to previousely?

ltclm:
i have tried to address the comments of @loicgasser, thanks for your help!

the impact on the query performance is massive:

```sql
-- original query
"Limit  (cost=0.28..8.56 rows=1 width=100) (actual time=2.374..2.376 rows=1 loops=1)"
"  ->  Index Scan using oev_haltestellen_the_geom on oev_haltestellen  (cost=0.28..8.56 rows=1 width=100) (actual time=2.373..2.374 rows=1 loops=1)"
"        Index Cond: (the_geom && '0103000020155500000100000005000000000000004F4B2341000000007CD80441000000004F4B2341000000001CD9044100000000774B2341000000001CD9044100000000774B2341000000007CD80441000000004F4B2341000000007CD80441'::geometry)"
"        Filter: (('01010000201555000000000000634B234100000000CCD80441'::geometry && st_expand(the_geom, 10::double precision)) AND _st_dwithin(the_geom, '01010000201555000000000000634B234100000000CCD80441'::geometry, 10::double precision))"
"Planning time: 0.290 ms"
"Execution time: 2.451 ms"

-- new query
"Limit  (cost=0.00..279.68 rows=201 width=100) (actual time=65.710..255.636 rows=1 loops=1)"
"  ->  Seq Scan on oev_haltestellen  (cost=0.00..11982.90 rows=8612 width=100) (actual time=65.708..255.632 rows=1 loops=1)"
"        Filter: st_dwithin(CASE WHEN (betriebspunkttyp = 'Haltestelle'::text) THEN st_envelope(st_setsrid(st_makeline(st_makepoint((st_x(st_envelope(the_geom)) - ((18 * array_length(regexp_split_to_array(verkehrsmittel, '\_'::text), 1)))::double precision) (...)"
"        Rows Removed by Filter: 25835"
"Planning time: 0.372 ms"
"Execution time: 255.769 ms"
```

I will prepare a materializev view with the necessary geometry indices for a performant identify request.

ltclm:
the matview has been created and is working fine:
```sql
"Limit  (cost=0.28..8.56 rows=1 width=164) (actual time=1.343..1.344 rows=1 loops=1)"
"  ->  Index Scan using oev_haltestellen_tooltip_bgdi_geom_poly on oev_haltestellen_tooltip  (cost=0.28..8.56 rows=1 width=164) (actual time=1.341..1.342 rows=1 loops=1)"
"        Index Cond: (bgdi_geom_poly && '010300002015550000010000000500000000000000EAD524410000000096400E4100000000EAD5244100000000E6400E4100000000FED5244100000000E6400E4100000000FED524410000000096400E4100000000EAD524410000000096400E41'::geometry)"
"        Filter: (('01010000201555000000000000F4D5244100000000BE400E41'::geometry && st_expand(bgdi_geom_poly, 5::double precision)) AND _st_dwithin(bgdi_geom_poly, '01010000201555000000000000F4D5244100000000BE400E41'::geometry, 5::double precision))"
"Planning time: 12.675 ms"
"Execution time: 1.428 ms"
```
uvek_master has been deployed to dev and int.
https://github.com/geoadmin/db/commit/3b35696b58e360fd445be505c98c91817c0bc0b0
So it is ready from my side.

chsdi/models/vector/uvek.py
```diff
@@ -675,7 +691,7 @@ class WindenergieanlagenFacility(Base, Vector):
     fac_y = Column('fac_y', Integer)
     fac_x = Column('fac_x', Integer)
     turbines = Column('turbines', Text)
-    __minscale__ = 100005
+    __minscale__ = 100006
```
gjn: Why this change?

gjn:
We shouldn't mix concerns in one PR. Could you remove the non-related changes?

ltclm:
yes, these changes should not be in this pr, i have removed them.

gjn:
Thanks alot. Another question:

How long does the update of the matview take? I'm asking because AFAIK, this is an auto-update layer.

ltclm:
it takes less than a second.

gjn:
@ltclm Thanks alot for this. It's a great user experience improvement!

PR #2002, opened 2016-03-29
fixes astra schwerunf

AFoletti:
Uses integer for population instead or real.
A change in the DB was also needed and is being pushed to INT just now


gjn:
LGTM. I'll merge once jenkins is happy and then I'll deploy to integration.

daguer:
It was not a good idea to modify this code.
The separators of thousands do not appear any more...

gjn:
@mariokeusen Can we live without the thousand separators?

gjn:
@daguer When doing something as specific as this, I would be great to add a comment in code so we can understand better why it was done.

mariokeusen:
When we are talking about the "population" field, then yes, we could live without the thousand separator, because they are simply not present in the data source. But I saw on INT that we have some thousand separators.

gjn:
Yes, we fixed it :)

PR #2008, opened 2016-03-30
Create last version

loicgasser:


gjn:
@loicgasser Doesn't seem to work. Instead of a version, we now have empty string as version...

gjn:
./venv/last-version exists but is empty.

Makefile
```diff
@@ -387,6 +386,9 @@ requirements.txt:
 	${PYTHON_CMD} setup.py develop
 	${PIP_CMD} install Pillow==3.1.0
 
+.venv/last-version:
+	test $(VERSION) != $(LAST_VERSION) && echo $(VERSION) > .venv/last-version || :
```
gjn: APP_VERSION instead of VERSION. I'll adapt.

PR #2012, opened 2016-03-30
Bafu layers not in bafu topic...

loicgasser:


mariokeusen:
Hope nothing has been changed on the topic configuration here. Because those layer are actually from BAFU, but are used in the aviation topic context only. Clearly definied exception.

loicgasser:
no we left it as is

PR #2017, opened 2016-04-05
update current timestamp documentation

ltclm:


loicgasser:
thks

gjn:
Shouldn't we mention that we advice to use 'default'? Or is it bad advice?

gjn:
Also, we should have informed the api list directly about this. It's quite an important change.

PR #2021, opened 2016-04-06
send lv95-like coordinates to SH portal

AFoletti:
SH portal does not work with lv03 coordinates. This sends pseudo-lv95 (not transformed) instead


gjn:
Testlink?

AFoletti:
[TESTLINK](http://mf-chsdi3.dev.bgdi.ch/ltfoa_cadastralwebmap_SH/rest/services/ech/MapServer/ch.kantone.cadastralwebmap-farbe/14/htmlPopup?coord=690385.3000000002,286063.05&imageDisplay=1920,1085,96&lang=en&mapExtent=690291.55,286013.025,690483.55,286121.525)

gjn:
Thanks alot.

Seems to me like their portal does not react to the coordinates at all. I can change them by hand and always get the same result. Am I doing something wrong?

AFoletti:
the trick is that the portal does not respond to click coordinates, but to mapgeoadmin view center. You can then change the tooltip coordinates as much as you want without getting any changes in the SG portal ;)

gjn:
ok. It reacts very roughly. I changed coordinates by 100 meter (the link to SH portal, not your test link) and it centered around same location.
Bigger changes and it reacts.

PR #2023, opened 2016-04-07
Correct documentation for iframe example

gjn:
The documentation is wrong in that the codepen exampled linked to does not contain a kml, but a GeoJSON layer.

The kml support is not yet implemented and under discussion here: https://github.com/geoadmin/mf-geoadmin3/pull/3199

@loicgasser Would be happy for a quick review.


gjn:
See faulty doc on official page: http://api3.geo.admin.ch/api/examples.html (last entry)

loicgasser:
+1 LGTM

PR #2028, opened 2016-04-13
added grids to relnotes

AFoletti:


loicgasser:
why is that? we updated the legend no?

PR #2033, opened 2016-04-14
Fix usage of assertRaises in unit tests

gjn:
Currently, these tests don't run at all (they
do nothing)

See official documentation
https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaises

@loicgasser We discussed this once. Please review.


loicgasser:
Thank you for that. Looks good.

PR #2045, opened 2016-04-26
Fix tests for esri encoders

loicgasser:
Since https://github.com/geoadmin/mf-chsdi3/commit/a5ef445ae07821654e7035307cf0affd63331c22#diff-73f43febd2482200bb4338623e93a959

json string is unwided in a different order but this is correct.
I also closed some polygons and multipolygons that were'nt before.

@gjn quick review merge please?


gjn:
hmm. meaning the original PR changed service results. Let's hope it's not used often...

loicgasser:
Well it changes the generated string order and not the content. When the json text is parsed the resulting dictionary will be the same.

gjn:
Yes, I interpreted incorrectly. I thought Polygons were being closed now - but that's not true. Sorry.

PR #2046, opened 2016-04-26
[WIP]: Sanatize ids during getFeature

loicgasser:
With the new geocover "shop" update we have primary keys of different types.

For some models the id is some text and for some other we have integers...

This is to handle the case where we would request a table using the wrong type.

I would add that this is a temporary and I have a greater scheme to handle all these border line cases.

@gjn Please have a look.

### DIFF
--- a/chsdi/views/features.py
+++ b/chsdi/views/features.py
@@ -12,7 +12,7 @@ import pyramid.httpexceptions as exc
 from sqlalchemy.orm.exc import NoResultFound, MultipleResultsFound
 from sqlalchemy.exc import InternalError, DataError
 from sqlalchemy.sql.expression import cast, func
-from sqlalchemy import Text, Integer, Boolean, Numeric, Date
+from sqlalchemy import Text, Integer, Boolean, Numeric, Date, Float
 from sqlalchemy import text
 from geoalchemy2.types import Geometry
 
@@ -318,21 +318,33 @@ def _get_features(params, extended=False):
             yield _get_feature_db(featureId, params, models)
 
 
+def is_valid_id_value(column, value):
+    column_type = column.type
+    python_type = column_type.python_type
+    try:
+        python_type(value)
+    except:
+        return False
+    return True
+
+
 def _get_feature_db(featureId, params, models):
     # One layer can have several models
+    feature = None
     for model in models:
-        query = params.request.db.query(model)
-        query = query.filter(model.id == featureId)
-        try:
-            feature = query.one()
-        except (NoResultFound, DataError):
-            feature = None
-        except MultipleResultsFound:
-            raise exc.HTTPInternalServerError('Multiple features found for the same id %s' % featureId)
+        if is_valid_id_value(model.id, featureId):
+            query = params.request.db.query(model)
+            query = query.filter(model.id == featureId)
+            try:
+                feature = query.one()
+            except (NoResultFound, DataError):
+                feature = None
+            except MultipleResultsFound:
+                raise exc.HTTPInternalServerError('Multiple features found for the same id %s' % featureId)
 
-        if feature is not None:
-            vectorModel = model
-            break
+            if feature is not None:
+                vectorModel = model
+                break
 
     if feature is None:
         raise exc.HTTPNotFound('No feature with id %s' % featureId)
@@ -551,7 +563,7 @@ def _format_search_text(columnType, searchText):
             return int(searchText)
         else:
             raise exc.HTTPBadRequest('Please provide an integer')
-    elif isinstance(columnType, Numeric):
+    elif isinstance(columnType, Numeric) or isinstance(columnType, Float):
         if re.match('^\d+?\.\d+?$', searchText) is not None:
             return float(searchText)
         else:
--- a/chsdi/tests/integration/test_features_service.py
+++ b/chsdi/tests/integration/test_features_service.py
@@ -133,6 +133,13 @@ class TestFeaturesView(TestsBase):
         self.assertIn('geometry', resp.json['feature'])
         self.assertEqual(resp.json['feature']['id'], 362)
 
+    def test_feature_pkey_type(self):
+        resp = self.testapp.get('/rest/services/all/MapServer/ch.swisstopo.geologie-geocover/GC-V-28', status=200)
+        self.assertEqual(resp.content_type, 'application/json')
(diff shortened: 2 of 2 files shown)


gjn:
Can't we integrate the data in such a way that both primary keys have the same type (string in this case?). The import should also assure that it's indeed unique accross multiple tables (as the service assumes as much).

chsdi/views/features.py
```diff
@@ -318,21 +318,33 @@ def _get_features(params, extended=False):
             yield _get_feature_db(featureId, params, models)
 
 
+def is_valid_id_value(column, value):
+    column_type = column.type
+    python_type = column_type.python_type
```
gjn: Where is this coming from? (From sqlAlchemy?)

loicgasser:
Yes

loicgasser:
We should though if I create a request from scratch I can end up on a 500 error.

gjn:
Yes, but that's already the case for other, 'correct' tables, no?

I'd rather adapt the data-integration. Can you talk about it with @pfanguin

loicgasser:
Looking into that.

loicgasser:
Okay so we can't change the ids of the grid layers as they are used by the toposhop.

We then then tried to use a hack by casting bgdi_id for the 7+ views associated to
 
ch.swisstopo.geologie-geocover
ch.swisstopo.geologie-geologischer_atlas

performance becomes really really terrible if you don't provide the scale (which is the case when you click a feature in the search) when requesting a feature in the grid. (35 sec)

The backend basically goes through 7+ views trying to match `GC-V-28` against a bgdi_id casted to text on the fly. -> totally unacceptable.

What's your take on that?

FYI: Looking at Jenkins I can tell that we have a lot "bad" mapping or "bad" tests.

ltclm:
Why do we use these views for the shop?

loicgasser:
I guess because we want to be able to order a digital map using some kind of custom pk25 grid.

I don't have time to find a better solution for tomorrow so I propose that we publish these 2 layers without the feature search (easily handled via layers configuration) warn the webshop people and find a proper solution for next deploy.

Regular Tooltip is still avaiable which is I believe the most important thing.

ltclm:
Yes, that should be enough for tomorrow, thanks loic.
if we use these views for the shop ***, we are performing queries on more than a million features to get the map sheet id's of 222 objects - complete nonsense in my eyes. 
We have a geocover map sheet layer and model in chsdi which can be used for the shop link and a geologischer atlas map sheet layer too, we should use these meta layers instead of the data layers. 

https://map.geo.admin.ch/?X=128814.02&Y=650797.82&zoom=0&lang=en&topic=ech&bgLayer=voidLayer&layers=ch.swisstopo.geologie-geocover.metadata,ch.swisstopo.gas25-papierkarte.metadata&layers_opacity=0.75,1

gjn:
Agree with @ltclm.
It's a data integration issue. Let's disable feature search and assure the data is well modelled and intgrated for our uses case.
Thanks for the analysis, @loicgasser.
Am 26.04.2016 9:08 nachm. schrieb Clausen Marcel <notifications@github.com>:Yes, that should be enough for tomorrow, thanks loic.
if we use these views for the shop ***, we are performing queries on more than a million features to get the map sheet id's of 222 objects - complete nonsense in my eyes. 
We have a geocover map sheet layer and model in chsdi which can be used for the shop link and a geologischer atlas map sheet layer too, we should use these meta layers instead of the data layers. 

https://map.geo.admin.ch/?X=128814.02&Y=650797.82&zoom=0&lang=en&topic=ech&bgLayer=voidLayer&layers=ch.swisstopo.geologie-geocover.metadata,ch.swisstopo.gas25-papierkarte.metadata&layers_opacity=0.75,1

loicgasser:
@gjn Why don't we use ch.swisstopo.geologie-geocover.metadata intead of ch.swisstopo.geologie-geocover?
For the geocover stuff, we already have everything we need with this layer. It's because it's a digital product? I do not understand the logic behind it.

gjn:
I have no idea. @p1d1d1 ?

p1d1d1:
so has been defined by data owner.

ltclm:
This layer does not contain the map sheet geometry and its primary key cannot be used as toposhop id. can we discuss this with the data owner?

p1d1d1:
Has been already discussed several times. They want the shop functionality on the pixel layer.
@ltclm the layer has 2 geometry layers for the tooltip... and this is scale dependent.
For zoom levels 0-5 (I guess) it uses the .metadata table (I guess), for > 5 the table with the vector features is used.

ltclm:
@p1d1d1 the shop link should only appear on the metadata table zoomlevels?

Do we need feature search on all zoomlevels or only the metadata zoomlevel?
If we need it on all zoomevels, i propose to add a new attribute to the feature tables and views which is ``NONE``  and which contains the Shop Key in the Metadata Shop Views. The current primary key of the feature tables should be removed from the queryable attributes, it is an autogenerated serial number!

ltclm:
As discussed with @p1d1d1 the feature search should be activated ONLY for the metadata zoomlevels ``GeocoverGridShop ``. @loicgasser is this doable?

AFoletti:
After a brief discussion with @ltclm and @rfrancel , our proposition is to handle that at BOD level and make the layer an aggregate.
This will allow to separate model classes assigns between mapsheet geometries and features geometries, thus solving the problem with a FSDI standard solution and without any hack

loicgasser:
Aggregate sounds good but we're always requesting the parent layer in geoadmin.
In CHSDI the concept of aggreagted layers does not really exist. 
Parentlayer id is always used and the only way to interogate only part of a layer is to provide a scale (which you don't want to provide when clicking a feature from the search)

Now we could extend this behaviour backend, model layers group which would give us more flexibility.

If only parent layer id is provided -> look though all the models (default behaviour)

If scale is provided (e.g. MapExtent and MapImageDisplay) -> query only models at the appropriate scale

If parentLayerId and Sublayerid(s) are provided -> query all sublayers only.

It's quite a developement but it might be necessary as we are handling more and more diverse models (grid, toposhop extents) and the actual solution is not ideal. 
Your thoughts?

p1d1d1:
Please consider that what you are discussing here applies also to ch.swisstopo.geologie-geologischer_atlas

loicgasser:
This can be closed. I'll open a ticket for the issue we talked about.

PR #2051, opened 2016-04-27
Fix shop image display

oterral:


oterral:
wait I modify only for mobile

loicgasser:
I saw that

oterral:
@loicgasser should be good

loicgasser:
yep thanks

PR #2056, opened 2016-05-02
add 2015 and 2016 to metadata layer legend

ltclm:


chsdi/static/js/ga-debug.js
```diff
@@ -57721,9 +57721,9 @@ ga.layer.create = function(layer, options) {
         return layerConfig['label'];
       }
     },
-    queryable: {
```
gjn: The changes in this file should not be in this PR. We should update the ol3 project first and re-create the files from there. Please remove the changes from this PR.

gjn:
There are log files added to the PR that should not be there. Same is true for other changes (in-code comment).

Please, review your PR as well...

ltclm:
i have removed the unwanted files. thanks for pointing that out.

Issue #1223, opened 2015-01-14
Test WMTS GetCap


loicgasser:
Check that for each `LayerRef`, there is an associated `Identifier`

loicgasser:
we have `<% layers = theme.fk_dataset_id.split(',')  %>` to determine the LayerRef
so from the themes and
`layers` from GetCap model...

procrastinatio:
Where is it missing ?

procrastinatio:
Should be a `NOT NULL`condition in the database.

See http://api3.geo.admin.ch/rest/services/aviation/1.0.0/WMTSCapabilities.xml

and *ch.bafu.schutzgebiete-aulav_auen*

cedricmoullet:
@loicgasser what are the next steps here ?

loicgasser:
@ltclm Has it been implemented in DB?

cedricmoullet:
@ltclm ?

ltclm:
> Should be a NOT NULLcondition in the database.
See http://api3.geo.admin.ch/rest/services/aviation/1.0.0/WMTSCapabilities.xml
and ch.bafu.schutzgebiete-aulav_auen

I am not sure whether I understood the problem here...
@procrastinatio / @loicgasser 
Every layer has to be in an inpsire theme?

loicgasser:
@rfrancel since you are working on the wmts, there tests we have to add to control the integrity of the WMTS getCap document.

In `Themes` check that for each `LayerRef` there is an associated `Identifier` in `Contents-Layer` .
Check that in each layer `Contents-Layer` the `TileMatrixSet` can be found in the `TileMatrixSet-Identifier` section.

http://www.gdal.org/frmt_wmts.html

Look if we could add the `ZeroBlockHttpCodes` and `ZeroBlockOnServerException` optional sections.

Doable for you?

rfrancel:
Yes I guess. I'll look into it.

loicgasser:
done

PR #2058, opened 2016-05-02
test if layers refs in layers defined

rfrancel:
@loicgasser  I added a test to check if the Layers Ref in theme refer to existing Identifier in Layers, but there is already a test testing the tilematrixSets: test_tilematrixsets_are_defined(self), L64.
Is this what you wanted?

linked to: https://github.com/geoadmin/mf-chsdi3/issues/1223


loicgasser:
yes you are right. LGTM

PR #2059, opened 2016-05-03
new legend pdf's for segeflugkarte and icao karte

ltclm:


gjn:
A little strange that we put those files in the api...

PR #2061, opened 2016-05-06
Assures that our git pre-commit hook is installed.

gjn:
This PR adds a couple of things:

1) It installs git hooks as part of the build process. This means that when you build the project, it is assured that the git hooks (pre-commit) is installed and executed.

2) We are now flexible in that we can handle the hooks with code inside the repository. No more manual adding of hooks as it was before

3) We integrate [git secrets](https://github.com/awslabs/git-secrets) project in the hooks to assure to not commit AWS related keys.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,7 +102,7 @@ help:
 .PHONY: all
 all: setup chsdi/static/css/extended.min.css templates potomo rss lint fixrights
 
-setup: .venv gdal node_modules
+setup: .venv gdal node_modules .venv/hooks
 
 templates: .venv/last-version apache/wsgi.conf apache/tomcat-print.conf print/WEB-INF/web.xml development.ini production.ini
 
@@ -377,6 +377,11 @@ production.ini: production.ini.in
 		--var "vector_profilename=$(VECTOR_PROFILENAME)" \
 		--var "shortener_allowed_domains=$(SHORTENER_ALLOWED_DOMAINS)" $< > $@
 
+.venv/hooks: .venv/bin/git-secrets ./scripts/install-git-hooks.sh
+	@echo "${GREEN}Installing git hooks${RESET}";
+	./scripts/install-git-hooks.sh
+	touch $@
+
 requirements.txt:
 	@echo "${GREEN}File requirements.txt has changed${RESET}";
 .venv: requirements.txt
@@ -388,6 +393,14 @@ requirements.txt:
 	fi
 	${PYTHON_CMD} setup.py develop
 	${PIP_CMD} install Pillow==3.1.0
+
+.venv/bin/git-secrets: .venv
+	@echo "${GREEN}Installing git secrets${RESET}";
+	rm -rf .venv/git-secrets
+	git clone https://github.com/awslabs/git-secrets .venv/git-secrets
+	cd .venv/git-secrets && make install PREFIX=..
+	(git config --local --get-regexp secret && git config --remove-section secrets) || cd
+	.venv/bin/git-secrets --register-aws
 
 .venv/last-version::
 	mkdir -p $(dir $@)
--- a/README.md
+++ b/README.md
@@ -198,31 +198,24 @@ To autocorrect most linting mistakes
 make autolint
   ```
 
-*Add a pre-commit hook*
+## Git hooks
 
-1. Create a pre-commit file
+3 git hooks are installed automatically when `make user` is called.
 
-  ```bash
-touch .git/hooks/pre-commit
-  ```
+All the hooks check that we don't accidently publish sensitive AWS keys to
+github - in the files as well as in the commit messages. We also execute
+`make lint` in the pre-commit hook.
 
-2. Copy/paste the following script
+Other checks can be added freely to any hook.
 
-  ```bash
-#!/bin/bash
+### `pre-commit` hook
 
-make lint
-if [[ $? != 0 ]];
-then
-  echo "$(tput setaf 1) Nothing has been commited because of styling issues, please fix it according to the comments above $(tput sgr0)"
-  exit 1
-fi
-  ```
+Called before committing changes locally. The commands in the `scripts/pre-commit.sh` script are executed.
 
-3. Make this it executable
+### `commit-msg` hook
 
-  ```bash
-chmod +x .git/hooks/pre-commit
-  ```
+Called before comitting changes locally and checks the commit message. The commands in the `scripts/commit-msg.sh` script are executed.
 
-Now commits will be aborted if styling is not respected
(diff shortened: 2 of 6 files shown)


gjn:
It might make sense to put this into a dedicated repository, generalize it and then use it in all of our projects. Let's first gain some experience with it. I'm not sure if this setup is robust enough.

scripts/commit-msg.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+PATH=.venv/bin:"$PATH"
+
+git secrets --commit_msg_hook -- "$@"
```
loicgasser: I still get
`.venv/bin/git-secrets: line 52: git secrets --aws-provider: command not found`

loicgasser: I would propose to use directly

`.venv/bin/git-secrets`

scripts/install-git-hooks.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+HOOKS=(pre-commit commit-msg prepare-commit-msg)
+
+# Adds the call to the local hook files to the .git/hooks commands
+# Creates the .git/hooks commands if they don't already exists
+for HOOK in ${HOOKS[@]}
+do
+  HOOK_FILE=.git/hooks/${HOOK}
+  HOOK_CMD="/bin/bash scripts/${HOOK}.sh"
```
loicgasser: minor but
`HOOK_CMD="./scripts/${HOOK}.sh"`
should be enough as we already have ` #!/bin/bash` at top of the pre-commit file

README.md
```diff
@@ -198,31 +198,24 @@ To autocorrect most linting mistakes
...
-  ```
+Called before committing changes locally. The commands in the `scripts/pre-commit.sh` script are executed.
 
-3. Make this it executable
+### `commit-msg` hook
 
-  ```bash
-chmod +x .git/hooks/pre-commit
-  ```
+Called before comitting changes locally and checks the commit message. The commands in the `scripts/commit-msg.sh` script are executed.
```
loicgasser: ok but what checks are we actually performing ?

scripts/commit-msg.sh
loicgasser: I's actually not related to this piece of code.

scripts/install-git-hooks.sh
```diff
@@ -0,0 +1,34 @@
...
+      echo $HOOK_CMD >> $HOOK_FILE
+      chmod +x $HOOK_FILE
+  fi
+
+done
+
+ALLOWED_KEYS=(01222222222222222222 03333333333333333333 02222222222222222222 MSLOSTPOINTERCAPTURE)
+for ALLOWED_KEY in ${ALLOWED_KEYS[@]}
+do
+  DUMMY=$(git config --get-all secrets.allowed | grep ${ALLOWED_KEY})
```
loicgasser: here you could do -> no DUMMY
`git config --get-all secrets.allowed | grep ${ALLOWED_KEY}  --silent`

loicgasser:
`pre-commit` and `prepare-commit-msg` are not doing anything yet right?

From doc:

pre-commit: Used to check if any of the files changed in the commit use prohibited patterns.

commit-msg: Used to determine if a commit message contains a prohibited patterns.

prepare-commit-msg: Used to determine if a merge commit will introduce a history that contains a prohibited pattern at any point. Please note that this hook is only invoked for non fast-forward merges.

Maybe add some of this info in our README, do something with commit-msg or prepare-commit-msg otherwise it's a little confusing..

In general I like it and I think it's a nice addition.

scripts/commit-msg.sh
gjn: Yes, I know about the error. I don't know why because when executing `git secrets --aws-provider`, it's working.

Officially, it's git secrets, so we should use that, IMO.

scripts/install-git-hooks.sh
gjn: Right.

README.md
gjn: This can be seen in each script...

gjn:
commit-msg and prepare-commit-msg do check commit messages. Either the -m parameter (commit-msg) or the commit message file (prepare-commit-msg). It's additional security, as potentially, users can put AWS keys in commit messages as well.

gjn:
The general idea is that `make all` installs all those hooks if they are not installed already. So basically, no one has to do anything has the hooks are automatically executed after you have build the project once.

loicgasser:
okay I couldn't get commit-msg and prepare-commit-msg to fail. I didn't think about adding an aws key though.

loicgasser:
> The general idea is that make all installs all those hooks if they are not installed already. So basically, no one has to do anything has the hooks are automatically executed after you have build the project once.

FYI: I had to remove my pre-commit hook because I had a mix between both after a make cleanall user.

gjn:
Yep, existing hooks continue to exists. At worst, it does twice the same thing.

gjn:
@loicgasser I've adapted according to your feedback.

Please have a final look and when good, merge. I want this in ASAP.

Makefile
```diff
@@ -100,12 +100,16 @@ help:
...
-all: setup chsdi/static/css/extended.min.css templates potomo rss lint fixrights
+all: setup installhook chsdi/static/css/extended.min.css templates potomo rss lint fixrights
 
 setup: .venv gdal node_modules
 
 templates: .venv/last-version apache/wsgi.conf apache/tomcat-print.conf print/WEB-INF/web.xml development.ini production.ini
 
+.PHONY: installhook
+installhook: .venv/bin/git-secrets ./scripts/install-git-hooks.sh
+	./scripts/install-git-hooks.sh
```
loicgasser: can we have materialized target to avoid installing it every time?

Makefile
```diff
@@ -389,6 +393,13 @@ requirements.txt:
 	${PYTHON_CMD} setup.py develop
 	${PIP_CMD} install Pillow==3.1.0
 
+.venv/bin/git-secrets: .venv
```
loicgasser: I would add
```
@echo "${GREEN}Installing git-secrets...${RESET}";
```
It's easier to read and you know if it's updated or not

gjn:
Good ideas. I've adapted.

loicgasser:
LGTM

gjn:
@hpchrist This is the amazon key linting you told us. It's active for this project for some time - if we don't encounter issues, we'll apply it to all public repositories.

hpchrist:
@gjn thanks, this certainly makes a lot of sense.

PR #2066, opened 2016-05-10
Fix primarykey type and classnames

loicgasser:
Thanks to https://github.com/geoadmin/mf-chsdi3/pull/2046
I realized some columns were not mapped correctly.


procrastinatio:
Looks OK, but I am not 100% convinced by classes name (still better though)

loicgasser:
Yes I rebased and fixed the cap letters missing. We need also some guidelines on how to integrate a layer in chsdi...

gjn:
@AFoletti Can you mention class names (CamelCase) in the DI process?

PR #2067, opened 2016-05-11
Ignore grids

loicgasser:
ping @oterral 

As they have a tooltip and are now productive layers we're trying to access a table that doesn't exist.
Here we just ignore the tests on those layers. Tests are already performed in the identify section anyway.


gjn:
+1.

PR #2073, opened 2016-05-11
adapt resource exists helper

rfrancel:
this PR adapts the "resource exists" helper so that it deals with connection errors, and does not make the app crash


loicgasser:
+ 1

gjn:
+1

PR #2075, opened 2016-05-13
Ltfoa windatlas tooltip

AFoletti:


AFoletti:
[TestLink](https://mf-chsdi3.dev.bgdi.ch/ltfoa_windatlas_tooltip/rest/services/ech/MapServer/ch.bfe.windenergie-geschwindigkeit_h50/1567_1137/htmlPopup?coord=642105,182242.49999999997&imageDisplay=1274,1085,96&lang=en&mapExtent=637035.0013480298,179636.74211416184,643405.0013480298,185061.74211416184)

(only API. Because of trivial change I did not deploy a geoadmin branch for testing)

loicgasser:
So we have 2 more info links?

AFoletti:
Yes: exactly that.
The whole is already really non-standard, so it makes maybe sense to remove the standard "more info" link and keep only the new one?

gjn:
2 links that point to same resource? Hmm. Not ideal. I would prefer to do the windrose smaller so the standard link is visible.

loicgasser:
Standard one is in base.mako, and if you add css to hide it, you'll also hide the other links.
reducing the size of the rose is a reasonable solution to me.

loicgasser:
it's actually doable if you hide the second child.
cssclass:nth-child(2)

gjn:
We should try to avoid too much customisation. I really would prefer to keep it as standard as possible; meaning to use existing linke and work with size of the elements that prevent the link from being visible in the tooltip.

procrastinatio:
Sorry, but these  tooltips craziness is going too far... Unmaintanable, untestable.

PR #2091, opened 2016-05-25
change id building class to correct error

rfrancel:
@gjn this should correct the fail. I could also remove this class cince this layer is not going to prod today.. What do you think is best?


rfrancel:
it was more than a typo that you noticed on the other PR, I corrected this also with the second commit

PR #2105, opened 2016-05-30
correct test identify query_time

rfrancel:
jenkins tests were failing because a test was not up to date. abortionaccomplished  is always null for ch.bazl.luftfahrthindernis. I replaced with a time query from another layer: ch.bafu.gefahren-historische_erdbeben


gjn:
I'm curious: why did this change in the luftfahrthindernisse? @AFoletti @rfrancel ?

rfrancel:
Actually there's only 12 data for which ther is a value for this attribute on the whole set, and not within the searched geometry. Do you prefer I change the geometry instead of the layer?

gjn:
@loicgasser fixed it as part of https://github.com/geoadmin/mf-chsdi3/pull/2106. Have a look. I think it's ok now.

PR #2107, opened 2016-05-30
correct link to wms in legend

rfrancel:
linked to issue https://github.com/geoadmin/mf-geoadmin3/issues/3218.
Assuming that the link has to redirect to the public wms (always the first in the table, as far as I know),
I adapted the legend template so that it redirects to this link. Here is a [test link](http://mf-chsdi3.dev.bgdi.ch/frr_wms_link_legend/rest/services/all/MapServer/ch.bafu.moose/legend?lang=fr).
But is this how it is supposed to be? Should the user be able to redirect to the swisstopo wms for example?

### DIFF
--- a/chsdi/templates/legend.mako
+++ b/chsdi/templates/legend.mako
@@ -38,6 +38,10 @@
       legend_url_pdf = False
   legend_url = host + '/static/images/legends/' + c['layerBodId'] + '_' + lang + '.png'
   times = c['attributes']['dataStatus'] if 'dataStatus' in c['attributes'] else '-'
+
+  wms_urls = str(c['attributes']['wmsUrlResource'])
+  wms_url = wms_urls.split(' ', 1)[0] 
+
 %>
 <div class="legend-container">
 <div class="legend-header">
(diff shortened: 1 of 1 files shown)


oterral:
Shouldn't it be a list of the wms  available ?

loicgasser:
In DB we have a text which is
```
https://wms.geo.admin.ch/?REQUEST=GetCapabilities&SERVICE=WMS&VERSION=1.0.0 | http://wms-swisstopowms.prod.bgdi.ch
```
I don't think the second address should be published and we don't have the full url to the GetCap anyway.

As a result I would only put the first one but it be better to have a list of strings to start with,

gjn:
We only publish to wms.geo.admin.ch (it's the official address used in map.geo and in [www.geo](http://www.geo.admin.ch/internet/geoportal/de/home/services/geoservices/display_services/services_wms.html#wms)

@AFoletti Does this fix work for all cases? If so, please merge this.

loicgasser:
rebased

PR #2111, opened 2016-06-01
Make sure the search text is not empty

loicgasser:
More robust validation of searchText, empty strings are not allowed.
Linked to https://github.com/geoadmin/mf-geoadmin3/pull/3351

### DIFF
--- a/chsdi/lib/validation/search.py
+++ b/chsdi/lib/validation/search.py
@@ -13,6 +13,11 @@ class SearchValidation(MapNameValidation):
 
     def __init__(self):
         super(SearchValidation, self).__init__()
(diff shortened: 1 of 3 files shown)


gjn:
Jenkins saying that [this test](identifyUrlTemplate) now fails. Obviously, we support bbox only search without any searchText - as described in our doc as well http://api3.geo.admin.ch/services/sdiservices.html#search

loicgasser:
obviously... I will adapt it though.

loicgasser:
I adapted the PR. Nothing will go through if not expected.

gjn:
Fantastic.

PR #2117, opened 2016-06-02
removed treasurehunt model

AFoletti:


AFoletti:
this has to be removed because the layer is a geojson one.
The layer now has tooltip=false in the BOD

gjn:
This is not an official layer, right? (WMTS, WMS, API3)

If not, it would break other services for this layer, like metadata service.

If yes, please merge.

AFoletti:
The layer is only present in the swisstopo topic (no API)

PR #2120, opened 2016-06-06
GeoJSON layers do not have GeoTables

loicgasser:


AFoletti:
we could then set tooltip=true for geojson layer without make tests fail, is that right?

loicgasser:
Yes, but the tooltip attribute is currently ignored for vector GeoJSON layers. It is assumed that they always have a tooltip in the description field. I will change this behaviour in geoadmin.

PR #2124, opened 2016-06-07
New tooltip and extended tooltip for baqua

pfanguin:
Test link : https://mf-geoadmin3.dev.bgdi.ch/?api_url=fap_baqua

Tooltip with photo: 
https://mf-chsdi3.dev.bgdi.ch/fap_baqua/rest/services/ech/MapServer/ch.bafu.gewaesserschutz-badewasserqualitaet/536/extendedHtmlPopup?lang=fr

Tooltip without photo: 
https://mf-chsdi3.dev.bgdi.ch/fap_baqua/rest/services/ech/MapServer/ch.bafu.gewaesserschutz-badewasserqualitaet/428/extendedHtmlPopup?lang=fr


pfanguin:
@mariokeusen : could you test the new tooltip and extended tooltip ?

pfanguin:
jenkins test

loicgasser:
can you rebase please? Is it to be ready for next deploy?

pfanguin:
Now i'ts ready

chsdi/models/vector/bafu.py
```diff
@@ -560,29 +560,31 @@ class StrukturgueteHochrheinSohle(Base, Vector):
 
 
 class GewaesserschutzBadewasserqualitaet(Base, Vector):
-    __tablename__ = 'gewaesserschutz_badewasserqualitaet'
+    __tablename__ = 'baqua'
     __table_args__ = ({'schema': 'wasser', 'autoload': False})
     __bodId__ = 'ch.bafu.gewaesserschutz-badewasserqualitaet'
-    __queryable_attributes__ = ['rbdsuname', 'groupid', 'bwname', 'nwunitname', 'rbdname']
+    __queryable_attributes__ = ['bwid', 'bwname', 'groupid', 'nwunitname', 'rbdsuname', 'gemeinde', 'canton', 'bwatercat', 'qualitaet_ch']
```
loicgasser: you could break down this line.

loicgasser:
Code is ok.

There is no primary key in DB on bgdi_id, please add it or change it to bwid as I think it makes more sense. (but add one please)

bwid is never bigger than 7 characters (sometimes 6), you could easily use:
```
varchar(7)
```
in DB and reflect it on the model.

Please fix the rights on this table as well (you forgot one).

Can you enforce NULL and NOT NULL in DB?

```sql
SELECT count(*) FROM wasser.baqua where groupid is null;
```

This is nullable for instance.

while

```sql
SELECT count(*) FROM wasser.baqua where groupid is null;
```

is not.

Then in the column definition add (I know this is new, but I want people to start paying attention to these things).

```python
bwname = Column('bwname', Text, nullable=False)
```

can you do that for all the columns please?

loicgasser:
very nice thank you! I wait for Jenkins and merge.

AFoletti:
@loicgasser I'm not 100% sure your "nullable=false" suggestion is something we should implement systematically.

If an attribute is **really** nullable or not, you can only find out in two ways:
- ask the datenherr (not ideal, since he probably does not know)
- look at the data model (would be ideal, but often there is no published data model)

To look at the data gives you an idea if for the present delivery you have nulls or not, but tells you nothing about future releases.
Filling our DBs with "NOT NULL" and our models with "nullable=False" is not the best way to avoid hassles in the future

loicgasser:
My feeling is that it's pretty easy to update and tells us more about what we can expect in a table.

Though I am not arguing that using NULL values where we have no data is always a MUST, I'm just saying that for the extra bit of time it takes, it will be useful to simplify the mako logic and so on.

ltclm:
We should not use varchar(xx) anymore in our tables and use the more flexible datatype text instead.

@loicgasser what is the advantage of null/not null mapping between datababase and sqlalchemy? 
this will force us to deploy / change code first if one of these attributes in the data will change from nullable=true to nullable=false or vice versa. in the worst case, a data update will crash chsdi... Back then we decided to get rid of autoload=true to have more flexibility on database side (add,remove,rename attributes) which was a good decision i think.

loicgasser:
With the current situation you can never quite be sure if a value exists or not.
Let s assume I want to get the length of a string. What does len(NULL) returns ?

If we know a column is nullable we may filter out unwanted results using the IS NOT NULL filter.
Knowing that a value is not NULL and is enforced in the DB is therefore useful.

I am not saying we should enforce not NULL values in tables that are updated daily and were we can t ensure the quality we indeed need flexibility. But for a yearly updated static table I think it makes sense. 
(which is the case here) 

Enforce what can be enforced safely and don t otherwise. 

Now let s assume I want to know which rows don t have data, If I don't know the default value (NULL, empty string and so on) how can I perform such an operation ? (we also need to know this default value if it exists) 

The flexible text column is probably fine for our use case, but know this is probably not very optimised memory wise.

ltclm:
Thanks for the explanations. 
Makes sense and yes, if possible we should try to find out during the data integration if a column can be NULL or not. So if we have a ``NOT NULL`` column (confirmed by data owner) in the table we will have to set the ``nullable=false`` parameter in the model. We will add this to the data integration documentation.

> The flexible text column is probably fine for our use case, but know this is probably not very optimised memory wise.

see here: https://www.postgresql.org/docs/current/static/datatype-character.html
> Tip: There is no performance difference among these three types, ... In most situations text or character varying should be used instead.

more detailed:  https://www.depesz.com/2010/03/02/charx-vs-varcharx-vs-varchar-vs-text/

PR #2116, opened 2016-06-02
Add deploy no print target

loicgasser:
@gjn as discussed
The PR adds the possibility to deploy the app only on the 2 main instances (without print and profile).

The motivation is that we don't really to update the print or profile function that often (a couple of times a year) but we always deploy on all 4 instances.

To test this PR try:

```
make deployintnoprint SNAPSHOT=201606011317 DEPLOYCONFIG=from_current_directory
```

We'll gain quite a bit pf time this way.

```
make deployint
```

still deploys the snapshot on all 4 instances.

PR #2126, opened 2016-06-07
Completly removing MapFish print

procrastinatio:
MapFish print v2 will continue its life in https://github.com/geoadmin/service-print

This PR removes all its artefacts in this project.

So merge, and deploy https://github.com/geoadmin/service-print/pull/4. Dito for https://github.com/geoadmin/mf-geoadmin3/pull/3365. 
Wenn triple tested and checked, merge that PR.

This replace PR https://github.com/geoadmin/mf-chsdi3/pull/2116 (deploy no print)


loicgasser:
Don t forget the hooks and scripts

loicgasser:
I would also remove the legacy targets in the configs files.

loicgasser:
You can remove all that is under https://github.com/geoadmin/mf-chsdi3/tree/master/print

Adapt gitignore.

gjn:
Legacy targets still needed for height/profile service.
Am 06.07.2016 4:15 nachm. schrieb Gasser Loïc <notifications@github.com>:I would also remove the legacy targets in the configs files.

loicgasser:
huuu, what's the point of removing the print if we still need to deploy chsdi on all 4 instances?

We need a service-height then, as we discussed it with @procrastinatio if I am not mistaken.
Let's focus on that and Jessie migration. We can do both at the same time.

procrastinatio:
@loicgasser _Service isolation_ (it's only the first step), it is stupid to deploy things that are not used, have not changed.

**service-alti** in progress https://github.com/geoadmin/service-alti

gjn:
yes, that's the goal.

Anyway, removing print from mf-chsdi3 could already be done.

loicgasser:
I will rebase/finish it and keep targets

loicgasser:
ping @procrastinatio and @gjn for a review

Makefile
```diff
@@ -440,10 +394,6 @@ fixrights:
 clean:
 	rm -rf production.ini
 	rm -rf development.ini
-	rm -rf apache/wsgi.conf
```
gjn: I think we need to keep this in the clean.

Makefile
```diff
@@ -440,10 +394,6 @@ fixrights:
...
-	rm -rf apache/tomcat-print.conf
-	rm -rf print/WEB-INF/web.xml
-	rm -rf apache/application.wsgi
```
gjn: Same here, we need to keep this.

Makefile
```diff
@@ -303,22 +273,6 @@ deploy/conf/00-branch.conf: deploy/conf/00-branch.conf.in
 	@echo "${GREEN}Creating deploy/conf/00-branch.conf...${RESET}"
 	${MAKO_CMD} --var "git_branch=$(GIT_BRANCH)" $< > $@
 
-apache/tomcat-print.conf.in:
```
gjn: This file is not removed, from what I can see. Please remove it.

gjn:
Added a couple of comments that I think need corrections. Also, a squash would be nice.

loicgasser:
rebased, squashed and adapted to comments
Thks for the review

loicgasser:
@gjin if we ever want to remove the useless target instances

loicgasser:
@gjn

PR #2132, opened 2016-06-15
Fix primary key for sphinxsearch

pfanguin:


pfanguin:
I removed bwid as primary key, because it's not a integer and so the search does not work.

PR #2154, opened 2016-07-06
Evd factorize models

loicgasser:
In response to https://github.com/geoadmin/mf-chsdi3/pull/2141/files

ping @alcapat please consider the following structure to factorize the models as it is easier to maintain.
Quick review and reassign it to me or gilbert when done.

Thanks !


alcapat:
LGTM.
Check all layers, it work's like before.
Thanks for factorization.

loicgasser:
Thanks

PR #2157, opened 2016-07-06
Allow empty tables

loicgasser:
I put back the earlier commented models to allow for rapid deploy a Friday at 17h00. :+1: 

ping @AFoletti and @ltclm if you have to publish an empty layer, simply define your layerId in `production.ini.in` and search for `empty_geotables`.
When layers have been finally updated remove them for the next deploy.


AFoletti:
that's nice!
But if I understand it well, automatic tests will still fail for the timespan between the removal of the layer in production.ini.in and the actual deploy of the data, right?

loicgasser:
No it should be good on chsdi side. 


Deploy the table to int and then prod 
Ensure metadata is updated on geoadmin side (if layers were previsouly set to test)

Before the next release simply remove the layer from production.ini.in

DB will then be okay in test and int 
Maybe I missed something ?

PR #2164, opened 2016-07-11
Update python libs

loicgasser:
https://github.com/geoadmin/mf-chsdi3/issues/2134

This PR updates all major dependencies.
Mainly we upgrade pyramid from 1.5 to 1.7

Full changelog can be found here:
http://docs.pylonsproject.org/projects/pyramid/en/latest/changes.html

Changes mainly aim at improving the security of the webserver.
Some intresting new features and nice improvements for use:

* Definition of custom exception views
* pserve --reload to not crash on syntax errors!!!

Main incompatibility issues come from callback function names that are now sanatized.
See:
https://github.com/Pylons/pyramid/pull/1627

Also, now one cannot pass the keyword self as parameter to function views. (Bad practise)


loicgasser:
@procrastinatio please test review

PR #2167, opened 2016-07-13
Use app staging rather than layer staging for wms

loicgasser:
This PR fixes the annoying bug where the link to WMS is wrong on int and dev.
Basically this means that all

- wms in dev will point to wms-bgdi dev
- wms in int will point to wms-bgdi int

ping @ltclm

### DIFF
--- a/chsdi/models/bod.py
+++ b/chsdi/models/bod.py
@@ -95,7 +95,9 @@ class LayersConfig(Base):
     def layerConfig(self, params):
         config = {}
         translate = params.translate
-        wmsHost = params.request.registry.settings['wmshost']
+        settings = params.request.registry.settings
+        geodataStaging = settings['geodata_staging']
+        wmsHost = settings['wmshost']
         for k in self.__dict__.keys():
             val = self.__dict__[k]
             if not k.startswith("_") and not k.startswith('geojsonUrl') and \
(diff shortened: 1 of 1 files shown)


loicgasser:
[Test Link](https://mf-geoadmin3.int.bgdi.ch/?api_url=gal_wmsaddress&lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers_timestamp=&layers=ch.bafu.gewaesserschutz-biologischer_zustand_diatomeen)

AFoletti:
so this PR makes the wms_url info from the BOD obsolete?

loicgasser:
Not really. If wms is different from wms bgdi it won't change it.

loicgasser:
and it doesn't change the URL in prod at all.

loicgasser:
@AFoletti You may have a good point. Stop me if I am wrong:

I propose we completely remove wmsUrl from the layers config service as it is always the same for WMS-BGDI.

In geoadmin simliarly to what we have done with the parameter `api_url`, we can force the app the point to a specific WMS. (`wms_url` ?)

So that you always get the chance to configure the app with a specific WMS for testing purposes.

Now I may miss some use cases. @ltclm ?

AFoletti:
yes, @gjn and I we discussed the issue some time ago and our final thought was exactly that: geoadmin should always use the wms at its same level exactly as he does for chsdi and DBs. Plus an URL paramenter in order to override it if needed
A solution like that would sole in my opinion TONS of problems

loicgasser:
I opened a PR.

We may also need to adapt:
https://github.com/geoadmin/ol3/blob/master/src/ga/layer/layer.js#L322

But we'll be able to  merge safely geoadmin and fix most of our problems.

gjn:
Ga PR is good, I guess. Will test in detail tomorrow.

JS API needs to be adapted, as you pointed out, before we can remove the information from the BOD.

gjn:
This PR is not needed anymore, right?

loicgasser:
I would keep api is would not work on int with WMS as well. so keep it, possibly merge it.

PR #2173, opened 2016-07-14
Remove height and profile

loicgasser:
This PR removes the profile and height from chsdi.

Of course as these services are public services, we'll need to adapt Varnish accordingly before this can be merged.

We may also want to change the URLs in the doc.


gjn:
Put to WIP...please focus on the other tasks. We have many open things.

loicgasser:
This can be merged. OGC proxy stuff won't influence the behaviour.

Issue #2178, opened 2016-07-18
Ensure all geometriesin DB do not exceed max size


loicgasser:
With
https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/views/features.py#L698
and
https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/views/features.py#L710

it is assumed that our geometries don't have more than 1'000'000 parts.

This is not the case, with the current approach we simply skip all geometries that exceeds this arbitrary limit.
The client on the other hand excepts a respone with a geometry property which he is likey not to receive.
This will to generate errors on the client side.

We should
1. Detect all `big geometries`
2. Simplify them accordingly.
3. Ensure that geometries we will publish in the future don't exceed this limit

@ltclm @alcapat @simonevonheute @daguer @gjn

ltclm:
hello, 
i dont agree in changing / reducing the precision of the original geometry (tilegeneration, wms, etc.), this is data manipulation. But we can introduce a simplifed geometry column for chsdi highlights.
p.e.
https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/models/vector/lubis.py#L17

loicgasser:
Sure that s what I had in mind sorry

gjn:
The above was introduced to protect our infrastructure. Huge geometries lead to timeouts (actually, it was the creation of the JSON string) - which consumed the apache connections, which in turn brought down all services.

We need simplified geometries for the web.

loicgasser:
I could easily detect all the tables where we have such geometries with a script. Then we need somebody to actually do the work.

procrastinatio:
Splitting or converting from `MULTI` to `SINGLE` large geometries is not an options ?

loicgasser:
This is now part of the integration process.

PR #2179, opened 2016-07-18
Raise error on big geoms

loicgasser:
Rather than silently removing geomtries, we inform the client why we can't provide this geometry.

Related to https://github.com/geoadmin/mf-chsdi3/issues/2178

### DIFF
--- a/chsdi/lib/exceptions.py
+++ b/chsdi/lib/exceptions.py
@@ -1,0 +1,8 @@
+# -*- coding: utf-8 -*-
+
+import pyramid.httpexceptions as exc
+
+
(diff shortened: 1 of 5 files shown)


loicgasser:
As general rule we should really avoid to put this kind of things directly in the code and make sure such geoms to reach the app.

loicgasser:
Not sure about this PR. Maybe we should wait for the simplification of the geometries. We can see that we have `layers=all` (all the layers of a topic) we have geometries that are too big and raise a `509` error. This is normaly never used but....

gjn:
+1 from me for this.

PR #2183, opened 2016-07-22
Update swisstlm3d legends and big pdf

daguer:


gjn:
Transparent backgrounds?

daguer:
I made as it is now because there is a link to big pdf...
transparent is needed in this case?!?

daguer:
I'm not sure at all it's necessary, but the legends are now transparents...

gjn:
I think the width are too big. Don't we have limit of 420px?

I'll merge for deploy today, but this might need adaption.

PR #2184, opened 2016-07-22
Improve coverage owschecker

mariavkourouni:


loicgasser:
jenkins test

chsdi/tests/integration/test_owschecker.py
```diff
@@ -15,6 +17,25 @@ def tearDown(self):
         super(TestOwsChecker, self).tearDown()
         del self.capture
 
+    def _callFUT(self):
+        base_url = 'http://wms.geo.admin.ch'
```
loicgasser: This variable should be red from production.ini file.

mariavkourouni: What do you mean? should I add at production.ini something like: base_url = ${base_url}
or
use here something like this: base_url = "http://" + app.registry.settings['wmshost'] + '/'

loicgasser: Second

PR #2198, opened 2016-08-11
ch.bazl.sachplan-infrastruktur-luftfahrt_anhorung will be empty from …

ltclm:
…2016-08-31


gjn:
Thanks for thinking about this early.

PR #2200, opened 2016-08-16
Fix decimals for gewaesserschutz layers

daguer:


ltclm:
and in the identify result we wont change the precision?

daguer:
@ltclm : For that, we have to modify the data... That's what you propose? 
Is it really worth it? 
I think that it is not necessary...

ltclm:
if @gjn agrees we keep it as it is.

gjn:
To me, change it tooltip only is ok. It's what appears.

The identify result might be seen as technichal. So I'll merge this for today.

PR #2201, opened 2016-08-16
replace zadara with efs

ltclm:
to be merged when infra is ready


gjn:
Test fails because mount does not exist on jenkins instance. Good!

loicgasser:
Addressed in https://github.com/geoadmin/mf-chsdi3/commits/master

gjn:
Sorry @ltclm, forgot about this branch.

PR #2202, opened 2016-08-16
Improve coverage features

mariavkourouni:


mariavkourouni:
Second commit: deleting double code which already exists at https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/lib/validation/features.py#L126

mariavkourouni:
@gjn @loicgasser

chsdi/templates/htmlpopup/cadastralwebmap.mako
```diff
@@ -6,7 +6,7 @@
 <%def name="table_body(c, lang)">
     <%
       from chsdi.models.vector import getScale
-      from chsdi.lib.validation.features import IdentifyServiceValidation
+      from chsdi.lib.validation.identify import IdentifyServiceValidation
```
gjn: Did you test this change? There are differences between the two IdentifyServiceValidation classes. See https://www.diffchecker.com/q6hgmxnq (the one you removed here is on the right)

mariavkourouni: yes, it looks like the removed code was recently adjusted, so should I move this code from the features.py in the identify.py?

gjn: No. Just test that this mako still works as expected. Do not change IdentifyServiceValidation in identify validation as it's already used elsewhere. Only this mako used the old and deprecated version - but it might be that the new version now breaks the mako.

loicgasser: Just tested it works well.

PR #2206, opened 2016-08-19
[WIP] Re-Enable kml temporary download service

gjn:
This is a fix for https://github.com/geoadmin/mf-geoadmin3/issues/3440

The service is needed for IE9 and Safari browsers to make the downloadKML functionality in map.geo.admin.ch work. The bug was introduced with this PR: https://github.com/geoadmin/mf-chsdi3/pull/2126 that was deployed on Wednesday.

[TestLink](https://mf-geoadmin3.dev.bgdi.ch/?api_url=//mf-chsdi3.dev.bgdi.ch/gjn_kml) Only testable in Safari or IE9.


loicgasser:
LGTM, didn't know we used to put kml in print dir.

gjn:
It's a hack for safari/IE as they don't allow the creation of a file via blob directly from browser. It's not used for chrome and Firefox.

In any case, we should even be able to remove it - as the kml exists now on public.geo.admin.ch, from where it should be downloaded with all clients. So, maybe we should not merge this and solve it in mf-geomadin3 - and then remove this service from mf-chsdi3 altogheter.

i'll open an issue in mf-geomadin3.

loicgasser:
Yes I think it's better also. Thank you for investigating.

gjn:
I'll put this to WIP and let you decide together with @oterral next week for the best solution.

oterral:
For me it's a bad regression so we should merge this and redeploy.

loicgasser:
We'll keep this solution for now as we don't have human readable file names otherwise. Redeploying...

PR #2208, opened 2016-08-22
Translation update from geocat

alcapat:
Typo error in the alternative title in German of layer ch.bafu.wasser-vorfluter


oterral:
I take the risk to merge without tests.

PR #2211, opened 2016-08-23
fix translation of luftfahrthindernis tooltip attribute

ltclm:
see [redmine ticket](https://redmine.prod.bgdi.ch/issues/3770#change-37151)


ltclm:
@mariokeusen

PR #2214, opened 2016-08-30
Remove legacy targets

loicgasser:
@gjn Any reason not to? If not please merge as I think it makes no sense anymore.

### DIFF
--- a/deploy/deploy.cfg
+++ b/deploy/deploy.cfg
@@ -21,15 +21,10 @@ content = Include /var/www/vhosts/mf-chsdi3/private/chsdi/apache/*.conf
 [remote_hosts]
 # mf0i
 int = ip-10-220-6-221.eu-west-1.compute.internal,
-      ip-10-220-6-41.eu-west-1.compute.internal,
-      ip-10-220-6-105.eu-west-1.compute.internal,
-      ip-10-220-6-208.eu-west-1.compute.internal
+      ip-10-220-6-41.eu-west-1.compute.internal
 
 
-# mf0 --> vpc_mf0_legacy_prod and vpc_mf0_prod
-prod = ip-10-220-5-86.eu-west-1.compute.internal,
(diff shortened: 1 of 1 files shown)


gjn:
Depends on ogcproxy.

We originally put ogcproxy onto legacy instances to avoid troubles that
it might cause. The standard apache congestion that might happen if a
service uses up all apache connections. If I remember correctly, we had
this a couple of times with ogcproxy as well (hence, we moved it).

Are we all sure it's a good idea to move ogcproxy back to mf instances 
that
contain all our other services?

@procrastinatio @oterral ?

loicgasser:
Ogc proxy won't disappear unless we remove the vhost.
I don't think it s used as much, we can analyze it though. We should finish this separation and not remain half way through.

gjn:
There was a reason for separation of ogcproxy. If this reason is still there, we should not put it on the instance where other important services are run.

The reasons are old and might not apply today - so this I can't tell.

gjn:
Btw: ogcproxy is still pointing to legacy instance. I propose to switch and hope we don't have any congestion anymore.

@procrastinatio Could you have a look at the ogcproxy code and try to avoid 'long' requests? Maybe head request first to determine possible size?

procrastinatio:
@gjn How do you detect long request ? You have no way to know if a server is slow or not.

@loicgasser @gjn Yes, the goal was to switch, but I don't like to push **too many** changes at the same time in varnish. But maybe, it is an effect of old age...

gjn:
We can wait with the varnish switch. ogcproxy runs on the legacies anyway.

ELK could be used to analyse response times.

We could improve the ogcproxy service in that we check for content-length (HEAD request) on the external resource and define some limit (20MB? 50MB?). Then only proxy content which is lower to that limit. This will likely help a ton in avoiding long response times for large resources.

PR #2221, opened 2016-09-02
Remove auth decorator from ogcproxy

procrastinatio:
Since we had double check, at varnish level and application level.

Genereal rule is to check everything at varnish level, except where we need a more fined solution (e.g. removing geometries, which has to be done by the service).


gjn:
Test needs to be adapted as well.

PR #2222, opened 2016-09-08
Improve search ranking

loicgasser:
This PR improves the ranking of the search.

Before we were building queries were some patterns would be repeated several times when

1. we only had digits
2. we only had non-digits

This would sometimes lead to strange ranking results.

We now test if the search pattern contains `digit` and `non-digit` words before building the query.

This is intended to work with https://github.com/geoadmin/service-sphinxsearch/pull/292

(indice have already been rotated in dev)
https://github.com/geoadmin/service-sphinxsearch/pull/292

[Test Link new dev](http://map.geo.admin.ch?api_url=//mf-chsdi3.dev.bgdi.ch/gal_search_ranking)

Note that it is significantly slower on dev.

To compare with previous behaviour on dev 

[Test Link previous dev also](http://map.geo.admin.ch?env=dev)

### DIFF
--- a/chsdi/tests/integration/test_search.py
+++ b/chsdi/tests/integration/test_search.py
@@ -149,6 +149,13 @@ class TestSearchServiceView(TestsBase):
         self.assertEqual(resp.content_type, 'application/json')
         self.assertEqual(resp.json['results'][1]['attrs']['detail'], 'rhonesandstrasse 16a 3900 brig 6002 brig-glis ch vs')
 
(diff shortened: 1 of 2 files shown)


davidoesch:
Ahh really nice 

Before this
image

now this

image


+1
(but i'm sure this si not the last tweak made..)

davidoesch:
issue https://github.com/geoadmin/mf-geoadmin3/issues/3485

Issue #2225, opened 2016-09-09
Bad results in identify


oterral:
This request should return 0 result:
https://mf-chsdi3.dev.bgdi.ch/rest/services/all/MapServer/identify?geometry=533750,136249.99999999994,550250,174249.99999999997&geometryFormat=geojson&geometryType=esriGeometryEnvelope&imageDisplay=1641,867,96&lang=fr&layers=all:ch.bav.haltestellen-oev&mapExtent=249750,-26750,1070250,406750&returnGeometry=true&tolerance=5&where=name+ilike+%27%25dietikon%25%27

This url requests bus stops with  `name=dietikon` and box geometry located around in Lausanne.

ltclm:
there are two resulting queries in the database:
[1] query: all stops with name like 'dietikon' gives 40 results
[2] query: all stops with name like 'dietikon' and bbox intersection gives no result

@loicgasser why do we make two db queries here. this beaviour results in an **or** condition of bbox and name, it should be **and** instead.

[1] 
```sql
  SELECT bav.oev_haltestellen_tooltip.nummer AS bav_oev_haltestellen_tooltip_nummer, ST_AsEWKB(bav.oev_haltestellen_tooltip.the_geom) AS bav_oev_haltestellen_tooltip_the_geom, ST_AsEWKB(bav.oev_haltestellen_tooltip.bgdi_geom_poly) AS bav_oev_haltestellen_tooltip_bgdi_geom_poly, bav.oev_haltestellen_tooltip.name AS bav_oev_haltestellen_tooltip_name, bav.oev_haltestellen_tooltip.tuabkuerzung AS bav_oev_haltestellen_tooltip_tuabkuerzung, bav.oev_haltestellen_tooltip.betriebspunkttyp AS bav_oev_haltestellen_tooltip_betriebspunkttyp, bav.oev_haltestellen_tooltip.verkehrsmittel AS bav_oev_haltestellen_tooltip_verkehrsmittel
        FROM bav.oev_haltestellen_tooltip
        WHERE name ilike '%dietikon%'
         LIMIT 201;
```
[2]
```sql
[...]

PR #2230, opened 2016-09-15
Scale dependent filtering early enough

loicgasser:
Fix for https://github.com/geoadmin/mf-chsdi3/issues/2225

The combination of a spatial filter, attribute filter and scale dependent layer would fail as the models where not filtered early enough .

[Test link](map.geo.admin.ch?api_url=//mf-chsdi3.dev.bgdi.ch/gal_res_management?layers=ch.bav.haltestellen-oev)


gjn:
Was this reviewed?

PR #2238, opened 2016-09-23
Sort by label if order key does not exist

loicgasser:
Fix for https://github.com/geoadmin/mf-geoadmin3/issues/3473

### DIFF
--- a/chsdi/views/catalog.py
+++ b/chsdi/views/catalog.py
@@ -7,7 +7,6 @@ from pyramid.httpexceptions import HTTPInternalServerError, HTTPNotFound
 
 from chsdi.models.bod import Catalog
 from chsdi.lib.validation import MapNameValidation
-from chsdi.lib.sqlalchemy_customs import remove_accents
 from chsdi.lib.filters import filter_by_geodata_staging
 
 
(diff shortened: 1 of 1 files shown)


procrastinatio:
I don't get it: https://mf-geoadmin3.dev.bgdi.ch/?api_url=http:%2F%2Fmf-chsdi3.dev.bgdi.ch%2Fgal_labelsort&lang=fr&topic=ech

loicgasser:
https://mf-geoadmin3.dev.bgdi.ch/?api_url=%2F%2Fmf-chsdi3.dev.bgdi.ch%2Fgal_labelsort&lang=fr&topic=ech
Try this one

procrastinatio:
So, if I do understand, layer's name have to be alphabetically sorted, but only if no key is available ? So something like:  that is not correct ? _Voyage > Cartes > Atlas_

loicgasser:
Yes for some categories we have an order key and for some other we don't. When the order key is not defined, we want to sort it alphabetically. But that s only valid for nodes of the same order. Did you catch anything suspicious ?

procrastinatio:
Looks good, except maybe that **Ä** or **ä** are shown after **z**.

See by yourself: https://mf-chsdi3.dev.bgdi.ch/~ltmom/json_sorting.html, some categories makes absolutely no sense, but as it is the same order across langage, I suppose it is the right order.

loicgasser:
Oh thank you for noticing the ä sorting.

loicgasser:
ok fixed thanks for the review

PR #2240, opened 2016-09-28
Remove empty geotable

loicgasser:
This layer is not empty anymore.
https://s.geo.admin.ch/6e9b8849ec


loicgasser:
Build error is unrelated (some wrz being integrated)

PR #2243, opened 2016-10-03
Remove deprecated demo and add legends target

loicgasser:
This PR removes the deprecated demo target and add the ability to download all the WMS legend images for a given layer id.
It ensures that:
- the backgrounds are transparent
- the images width don't exceed 480px
- the images are optimized with optipng (level 3)

I am aware that this won't necessary work for all the layers as we still manipulate some of the images by hand, but it should be useful for most of them and help to get at least the base images.

Try for instance:

`make legends BODID=ch.blw.erosion-quantitativ`

ping @ltclm , @pfanguin , @AFoletti , @Ariadnigaki ,@Sgachet

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ BRANCH_TO_DELETE :=
 CURRENT_DIRECTORY := $(shell pwd)
 DEPLOYCONFIG ?=
 DEPLOY_TARGET ?=
+BODID ?=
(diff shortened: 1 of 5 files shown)


gjn:
Fantastic!

Should be showcases next Tuesday and made known to all data integrators.

PR #2248, opened 2016-10-05
Doc and translations

loicgasser:


gjn:
jenkins test

loicgasser:
There was also a bad column type mapping. it's fixed

PR #2260, opened 2016-10-11
Fix tests and remove :all test for identify

gjn:
This should fix the jenkinds failures.

It removes the :all tests that interrogates all layers at the same time resulting in huge responses. In addition, it adds an identify test for every layer that has a tooltip.

### DIFF
--- a/chsdi/tests/integration/test_identify_service.py
+++ b/chsdi/tests/integration/test_identify_service.py
@@ -97,7 +97,7 @@ class TestIdentifyService(TestsBase):
 
     def test_identify_valid(self):
         params = {'geometry': '548945.5,147956,549402,148103.5', 'geometryType': 'esriGeometryEnvelope', 'imageDisplay': '500,600,96',
-                  'mapExtent': '548945.5,147956,549402,148103.5', 'tolerance': '1', 'layers': 'all'}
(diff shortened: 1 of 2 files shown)


chsdi/tests/integration/test_identify_service.py
```diff
@@ -133,16 +133,16 @@ def test_invalid_imageDisplay(self):
 
     def test_identify_valid_topic_all(self):
```
loicgasser: Change name

chsdi/tests/integration/test_layers.py
```diff
@@ -109,6 +109,12 @@ def checkLegend(self, layer):
             resp = self.testapp.get(link)
             assert resp.status_int == 200, link
 
+    def checkIdentify(self, layer):
+        link = '/rest/services/all/MapServer/identify?geometry=558945.5,147956,559402,148103.5&geometryType=esriGeometryEnvelope&imageDisplay=500,600,96&mapExtent=558945.5,147956,559402,148103.5&tolerance=1&layers=all:' + layer
+        resp = self.testapp.get(link)
+        assert resp.status_int == 200, link
+        assert resp.content_type == 'application/json'
```
loicgasser: Link as second argument would be nice as well.

loicgasser:
+1, only a minor comment, thanks for the fix

gjn:
Adapted according to your comments. Thanks for the review.

PR #2266, opened 2016-10-13
Optional scale argument to download scale

loicgasser:
It is no possible to specify a SCALE when downloading legends

`make legends BODID=ch.layername WMSSCALELEGEND=1000`

ping @alcapat @ltclm @daguer @pfanguin @Sgachet @rebert


loicgasser:
minor and trivial

PR #2272, opened 2016-10-19
Remove useless model for globalstrahlung meteo

pfanguin:


loicgasser:
templates/htmlpopup/globalstrahlung.mako

pfanguin:
The mako was removed before.

PR #2292, opened 2016-10-26
Change type of column perimeter for smv layers

pfanguin:
Change type "Integer" to "Unicode"
Add releasenotes for bfe layers


oterral:
thx

PR #2297, opened 2016-10-27
add legends for ch.vbs.sachplan-infrastruktur-militaer_anhoerung layer

ltclm:
We forgot to add the legends of this new layer.
@mariokeusen


loicgasser:
Now it's 2 things that are missing in chsdi. It makes sense to redeploy I think @gjn what do you think?

PR #2308, opened 2016-11-02
WMS URL from build env

loicgasser:
@ltclm This PR follows the changes made in wms_metadata table.

Official WMS staging should always follow the app staging and should alwas be https://wms.geo.admin.ch

Now that we are flexible with the `wms_url` permalink parameter in geoadmin, we no longer need to reference it in the layers configuration. (and wmsUrl in layersConfig is already ignored in geoadmin but is still used in JS geoadmin API)


ltclm:
great, thank you.
Will remove the attribute `` wms_url `` from bod_master.re3.view_layers_js
when this code has been deployed to prod.

loicgasser:
+1

PR #2313, opened 2016-11-04
Rename shop layers

daguer:


daguer:
@loicgasser : the last 3 layers to rename: I have made 3 new layers

loicgasser:
Great as those are pointing to the same layer you should define `extend_existing=True`.
Could you prepare a SQL migration script we can apply after the deploy?
Basically a set of DELETE commands in the correct order.

loicgasser:
I meant the same table.

loicgasser:
To test your migration script make sure you use your own DB.

loicgasser:
Now if you could prepare the delete PR post-deploy please?

PR #2320, opened 2016-11-09
Update setuptools on virtualenv install and install Pillow using pip

loicgasser:
Enable Wheel download by updating setuptools.

Wheels are use by default when using 
`setuptools >= 0.8`

``time make cleanall user``

Build time before the first time: 3m43.512s
Build time before the second time: 3m48.863s

Build time after the first time: 3m25.514s
Build time after the second time (using cache): 3m19.680s

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -191,7 +191,7 @@ gdal: .venv
 	then \
 		echo "${GREEN}Installing GDAL...${RESET}"; \
 		mkdir -p $(INSTALL_DIRECTORY)/build && \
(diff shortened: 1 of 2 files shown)


loicgasser:
jenkins test

procrastinatio:
LGTM
My only concern, do we really need the mamoth GDAL library ? Reprojection --> `pyproj`

loicgasser:
Thanks for having a look, Concerning GDAL, yes we could find a better solution. I just don t trust system packages. We ve seen a lot of differences between Dev and int. Maybe an upgrade for Jessie.

Issue #2289, opened 2016-10-25
2 BAZL layer decommission


AFoletti:
the BAZL layers
ch.bazl.heliports-gebirgslandeplaetze and ch.bazl.points-of-interest have to be decommissioned.
I already set their staging to "test"

Still ToDo:

Catalog:
- [x] Remove from googledocs

WMS:
- [x] Remove from wms-bod
- [x] Delete snippets in mapfile_include

CHSDI:
- [x] models
- [x] translations
- [x] makos
- [x] legends

BOD:
- [x] dataset
- [x] geocat_data
- [x] abnahme
- [x] layers_js
- [x] translations

Geometries:
- [x] remove geometry tables

Geocat:
- [x] Remove entry

loicgasser:
`I already set their staging to "test"`

We may have to remove them from the catalog and sphinx (if searchable= for tomorrow's deploy. Otherwise I don't see any potential conflict with this change.

AFoletti:
they are not searchable, so no problem on sphinx.
They are already removed from the catalog spreadsheet, but I have to check the import

AFoletti:
Redmine tickets: 4027 + 4028

AFoletti:
chsdi PR https://github.com/geoadmin/mf-chsdi3/pull/2325

AFoletti:
Everything is done and I asked for the removal of the geocat entries

AFoletti:
closing this since PR is merged

PR #2325, opened 2016-11-14
removed transl and legends for decommissioned BAZL layers

AFoletti:


procrastinatio:
C'est aussi supprimé de tous les autres canaux de publications ? WMS, WMTS, Mapproxy, WFS, print, etc ?

AFoletti:
wms is ok
wmts is ok (tiles to be removed)
mapproxy is still NOK, but the link in the gdoc does not work
WFS not an issue in my opinion. It's not integrated in any WFS service I know of
Print? etc? What do you exatly mean?

loicgasser:
Yes @procrastinatio Checklist is here:
https://github.com/geoadmin/mf-chsdi3/issues/2289

Looks ok, thanks

PR #2326, opened 2016-11-14
Use pk product as ID

loicgasser:
Use pk_product as primary keys.


procrastinatio:
`id`: Integer --> Unicode ?

loicgasser:
I don't see the problem here, we're simply saying a different primary key should be used for these layers.

PR #2332, opened 2016-11-15
Pin current modules

loicgasser:
Several updates broke the build in one way or another.
Let's pin all our libs and update manually.
See also https://github.com/Pylons/pyramid_tm/issues/49

(transactions and flake8)


loicgasser:
CHSDI is down in dev and a test is trying to access a resource:

http://mf-chsdi3.dev.bgdi.ch/examples/bln-style.kmz

this will be fixed once deployed.

PR #2334, opened 2016-11-16
Dynamically change WMTS host

procrastinatio:
* For debug Zweck
* To match requested host. Per default, it is wmts.geo.admin.ch.

### DIFF
--- a/chsdi/views/wmtscapabilities.py
+++ b/chsdi/views/wmtscapabilities.py
@@ -5,9 +5,11 @@ from pyramid.httpexceptions import HTTPBadRequest, HTTPNotFound
 from gatilegrid import GeoadminTileGrid
 
 from chsdi.models.bod import get_wmts_models
-from chsdi.lib.helpers import sanitize_url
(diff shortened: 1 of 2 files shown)


chsdi/views/wmtscapabilities.py
```diff
@@ -5,10 +5,12 @@
 from gatilegrid import GeoadminTileGrid
 
 from chsdi.models.bod import get_wmts_models
-from chsdi.lib.helpers import sanitize_url
+from chsdi.lib.helpers import sanitize_url, is_valid_hostname
 from chsdi.lib.validation import MapNameValidation
 from chsdi.lib.filters import filter_by_geodata_staging, filter_by_map_name
 
+WMTS_DEFAULT_HOST = 'wmts.geo.admin.ch'
```
loicgasser: https://github.com/geoadmin/mf-chsdi3/blob/master/rc_prod#L13

Here we have it already no?

loicgasser:
LGTM, a minor comment, maybe a test also.

PR #2340, opened 2016-11-18
made link to SO portal use LV95

AFoletti:


AFoletti:
Since last week, the SO portal uses LV95. This PR corrects our links to that portal

loicgasser:
and they are not backward compatible ! thanks

PR #2345, opened 2016-11-22
fix OpenLayers 3 documentation url

xarem:
ol3js.org does no longer link to the OpenLayers 3 documentation.

URL: https://api3.geo.admin.ch/api/doc.html


loicgasser:
Thanks for the fix!

PR #2354, opened 2016-11-30
Fixing vd open data mako

gjn:
ping @loicgasser @ltclm 

Calling the htmlpopu without any parameters should not fail. It currently does because we don't have default coordinates and we don't check if the coords parameter is acutally present.


loicgasser:
Thanks

Issue #2333, opened 2016-11-16
Document multi-version javascript API


gjn:
We should mention it in the docs at least. Only then we'll release it officially.

loicgasser:
http://mf-chsdi3.dev.bgdi.ch/api/quickstart.html#include-the-geoadmin-javascript-api

It is in the quick start page.
We could mention it in the examples as well.

oterral:
We need to address polyfill issue.  See feedbacks in google group.

PR #2375, opened 2016-12-07
Make loading of polyfill optional via parameter

gjn:
The polyfill might be in conflict with other libaries. This adds a parameter to the loader to avoid loading our polyfill into the application.

@loicgasser: I think your original idea is best.


loicgasser:
Yes it works well.
Thanks @gjn , errors in tests come from the fact that `bav.schienennetz_segment.id` is not deployed yet in dev. I redeploying
Could we have some doc in the quick start and examples, link to https://github.com/geoadmin/mf-chsdi3/issues/2333

loicgasser:
jenkins test

PR #2377, opened 2016-12-07
update layer ch.bav.schienennetz_*

rebert:
For the update of the schienennetz, the tooltip had to be adapted as well: [try](https://mf-geoadmin3.dev.bgdi.ch/?api_url=%2F%2Fmf-chsdi3.dev.bgdi.ch%2Fltrea_bav_eisenbahn_strecke&lang=de&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege,ch.bav.schienennetz&layers_visibility=false,false,false,false,true&layers_timestamp=18641231,,,,&X=215750.00&Y=644000.00&zoom=5)

### DIFF
--- a/chsdi/models/vector/uvek.py
+++ b/chsdi/models/vector/uvek.py
@@ -41,16 +41,17 @@ class SchienennetzPoint(Base, Vector):
     __bodId__ = 'ch.bav.schienennetz'
     __label__ = 'nom_point'
     __queryable_attributes__ = ['numero', 'nom_point', 'abreviation']
-    id = Column('xtf_id', Integer, primary_key=True)
+    id = Column('id', Integer, primary_key=True)
+    xtf_id = Column('xtf_id', Unicode)
     numero = Column('numero', Integer)
-    nom_point = Column('nom_point', Text)
-    abreviation = Column('abreviation', Text)
-    respdonneesabreviation = Column('respdonneesabreviation', Text)
-    debutvalidite = Column('debutvalidite', Text)
-    finvalidite = Column('finvalidite', Text)
+    nom_point = Column('nom_point', Unicode)
+    abreviation = Column('abreviation', Unicode)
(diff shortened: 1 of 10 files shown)


loicgasser:
rebase me please!

loicgasser:
@rebert could you try to keep the original rows?
Otherwise you're gona break the app when we'll deploy. (as we deploy data first)
Remove the rows after the deploy please.

rebert:
Now I added the attribute id again as primary key - I will change id to id in the model and rebase this branch ...

loicgasser:
Best is to leave the old table as is when the structure changes and create a new one. Thanks for having a look.

rebert:
@loocgasser: if the shop deploy to dev/int still causes problems because of these layers, the old tables (bav.schienennetz_segment, bav.schienennetz_point) could be recovered (from uvek_prod) ... on Monday this issue is going to be my first prio!

chsdi/models/vector/uvek.py
```diff
@@ -61,21 +59,27 @@ class SchienennetzSegment(Base, Vector):
     __label__ = 'nom_segment'
     __queryable_attributes__ = ['nom_segment']
     id = Column('id', Integer, primary_key=True)
-    xtf_id = Column('xtf_id', Integer)
+    xtf_id_tooltip = Column('xtf_id_tooltip', Unicode)
```
loicgasser: Is that the new primary key?

loicgasser:
LGTM, minus a comment on the primary key.

chsdi/models/vector/uvek.py
rebert: No. Now the structure is the same as before with some aditional atributtes

PR #2379, opened 2016-12-08
Do not make apache crash if config are missing

procrastinatio:
If an apache `Include` directory does not have a single `.conf` file it crashes. Use `IncludeOptional` instead. 
Warning apache 2.4 only (Jessie)


loicgasser:
very nice I didn't know this option!

gjn:
+1

Issue #2371, opened 2016-12-06
Missing htpasswd/htgroup on some target


procrastinatio:
Fix it manually or wait for the puppet recipies

```
/var/www/vhosts/mf-chsdi3/private`/htpasswd
/var/www/vhosts/mf-chsdi3/private/htgroup
```

PR #2409, opened 2016-12-19
Basic auth credentials files are managedd by popets now

procrastinatio:
See https://github.com/geoadmin/mf-chsdi3/issues/2407 and https://github.com/geoadmin/mf-chsdi3/issues/2371

And remove _/var/www/vhosts/mf-chsdi3/private/htpasswd_ and _/var/www/vhosts/mf-chsdi3/private/htgroup_ once merged (not in github obviously)


loicgasser:
LGTM thanks

PR #2413, opened 2016-12-21
Initial import fassaden model

ltclm:


AFoletti:
jenkins test

AFoletti:
@ltclm I added the queryable attributes. It works with gwr_egid but not with building_id. Returns a 503
Maybe we should materialize the view?

ltclm:
hello @AFoletti 
the building_id does not exist in the table ``bfe.solarenergie_fassaden``, it has to be generated with a spatial relationship in the fassaden view:
```sql
CREATE OR REPLACE VIEW bfe.view_solarenergie_fassaden_gs AS 
 SELECT f.ff_uuid,
    f.ff_nummer,
    ( SELECT d.building_id
           FROM bfe.solarenergie_daecher d
          WHERE st_intersects(f.the_geom, d.the_geom)
         LIMIT 1) AS building_id,
```
It is enough to replace the spatial relationship in the view ``bfe.view_solarenergie_fassaden_gs`` with an indexed integer column building_id in the table ``bfe.solarenergie_fassaden``. This column has to be filled either with fme or with an update trigger, using the query above.
i would try to avoid the creation of a materialized view, it duplicates the time to deploy and the need of diskspace.

AFoletti:
@ltclm yes, building_id is only present in the view. But the model does indeed use the view and not the table:
https://github.com/geoadmin/mf-chsdi3/pull/2413/files#diff-cfd213600fd54efaa211b4ae7cb7e3b5R84

Ok about not materializing the view. I thought the error was bound to a timeout or somewhat.
I'm probably missing something here

ltclm:
yes we have to physically add this column to the table ``bfe.solarenergie_fassaden`` which is the main source of the view ``bfe.view_solarenergie_fassaden_gs``.

loicgasser:
needs rebase and squash

PR #2429, opened 2017-01-24
Fix build and six requirements

loicgasser:


loicgasser:
Build is fixed errors are related to ch.mobility.standorte and treasurehunt

PR #2450, opened 2017-02-09
Add script to replace scales KML files from public prod bucket

oterral:


scripts/replace-scales.py
```diff
@@ -0,0 +1,118 @@
...
+import base64
+import time
+
+utc=pytz.UTC
+
+
+begin_naive = datetime.datetime(2017,01,26)
+#end_naive = datetime.datetime(2017,01,27)
+
+begin_naive = datetime.datetime(2016,10,25)
```
loicgasser: Declared twice

loicgasser: And not used it seems

scripts/replace-scales.py
```diff
@@ -0,0 +1,118 @@
...
+                file_content = re.sub('<scale>2\.25</scale>', '<scale>1.5</scale>', file_content);
+                file_content = re.sub('<scale>4</scale>', '<scale>2</scale>', file_content);
+                #print file_content
+                #print m.group(0)
+                #print obj.key, obj.last_modified
+              
+                # Zip content
+                f = gzip.open('tmp/kml_replaced/' + obj.key, 'wb')
+                f.write(file_content)
+                f.close();
```
loicgasser: You back them up after modifying them ?

scripts/replace-scales.py
```diff
@@ -0,0 +1,118 @@
...
+                # Send to int bucket
+                f = open('tmp/kml_replaced/' + obj.key, 'rb')
+                f.seek(0)
+                bucket.delete_objects(
+                   Delete={
+                       'Objects': [{
+                            'Key': obj.key
+                        }]
+                   }
+                )
```
loicgasser: Delete is not necessary, put overwrites the object according today the doc

loicgasser:
Build error is related to the changes introduced in https://github.com/geoadmin/service-sphinxsearch/pull/322

Different detail field for gazetter.

scripts/replace-scales.py
oterral: yes

loicgasser: :suspect:

Issue #2456, opened 2017-02-14
Duplicated primary keys (feature IDs) accross tables of the same technical name


loicgasser:
Schema

```python
[DB, BODID, ClassA, ClassB, ClassATable, ClassBTable]
```

```python
['uvek', u'ch.bfe.sachplan-geologie-tiefenlager-thematische-darstellung', <class 'chsdi.models.vector.uvek.SgtPlanningTd'>, <class 'chsdi.models.vector.uvek.SgtPlanningRasterTd'>, 'geologische_tiefenlager', 'geologische_tiefenlager_raste
r']
['bafu', u'ch.bafu.schutzgebiete-aulav_uebrige', <class 'chsdi.models.vector.bafu.schutzgebiete_aulav_uebrige'>, <class 'chsdi.models.vector.bafu.schutzgebiete_aulav_uebrige_general'>, 'uebrige_gebiete', 'uebrige_gebiete_general']
['uvek', u'ch.bazl.sachplan-infrastruktur-luftfahrt_kraft', <class 'chsdi.models.vector.uvek.SilPlanningK'>, <class 'chsdi.models.vector.uvek.SilPlanningRasterK'>, 'sachplan_inf_luft_plmeasures_kraft', 'sachplan_inf_luft_plmeasures_r_kra
ft']
['vbs', u'ch.vbs.sachplan-infrastruktur-militaer_kraft', <class 'chsdi.models.vector.vbs.SimPlanningK'>, <class 'chsdi.models.vector.vbs.SimPlanningRasterK'>, 'sim_pl_kraft', 'sim_pl_r_kraft']
['zeitreihen', u'ch.swisstopo.zeitreihen', <class 'chsdi.models.vector.zeitreihen.Zeitreihen_15'>, <class 'chsdi.models.vector.zeitreihen.Zeitreihen_20'>, 'tooltip_15', 'tooltip_20']
['zeitreihen', u'ch.swisstopo.zeitreihen', <class 'chsdi.models.vector.zeitreihen.Zeitreihen_15'>, <class 'chsdi.models.vector.zeitreihen.Zeitreihen_22'>, 'tooltip_15', 'tooltip_22']
['zeitreihen', u'ch.swisstopo.zeitreihen', <class 'chsdi.models.vector.zeitreihen.Zeitreihen_20'>, <class 'chsdi.models.vector.zeitreihen.Zeitreihen_22'>, 'tooltip_20', 'tooltip_22']
[...]

PR #2457, opened 2017-02-14
Scale ordering rather than filtering for htmlpopup and feature services

loicgasser:
@gjn @procrastinatio please review. (possibly before the 22nd :) )

[Test Link](https://mf-geoadmin3.dev.bgdi.ch/?api_url=%2F%2Fmf-chsdi3.dev.bgdi.ch%2Fgal_scale_html&lang=fr)

A Bod ID and Feature ID is all we should need to retrieve a correct model. 

mapExtent and imageDisplay parameters are provided in geoadmin when requesting:

- a feature
- a htmlpopup content

Note that we alos need to fix (code to generate this list is in this PR):
https://github.com/geoadmin/mf-chsdi3/issues/2456

Right now we filter out models that are not suitable.
With this PR we use the models that are the most suitable first, and stop when we find the first match.

### DIFF
--- a/chsdi/models/__init__.py
+++ b/chsdi/models/__init__.py
@@ -74,10 +74,20 @@ def oereb_models_from_bodid(bodId, scale=None):
     return models
 
 
-def models_from_bodid(bodId, scale=None, resolution=None):
+# Returns None is the bodId is not registered and an empty list if the bodId is registered but
+# the scale filtered all models out. orderScale orders by scale validity
+def models_from_bodid(bodId, scale=None, resolution=None, orderScale=None):
     models = bodmap.get(bodId)
     if models:
-        if scale is not None:
+        if orderScale is not None:
+            mdls = []
+            for m in models:
+                if orderScale < max_scale(m) and orderScale  >= min_scale(m):
+                    mdls.insert(0, m)
+                else:
+                    mdls.append(m)
+            models = mdls
+        elif scale is not None:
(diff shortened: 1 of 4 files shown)


chsdi/models/__init__.py
```diff
@@ -74,10 +74,20 @@ def oereb_models_from_bodid(bodId, scale=None):
...
     models = bodmap.get(bodId)
     if models:
-        if scale is not None:
+        if orderScale is not None:
+            mdls = []
+            for m in models:
+                if orderScale < max_scale(m) and orderScale  >= min_scale(m):
+                    mdls.insert(0, m)
+                else:
+                    mdls.append(m)
```
procrastinatio: I don't get it, in the PR description, you say 'break on first match', and you insert anyway all models ?

procrastinatio:
What is _models_scan.py_ for ?

chsdi/models/__init__.py
loicgasser: yes, on first request match, here I reorder according to the scale (the ones that match the scale first) but I don't filter them anymore.
The break statement is in `_get_feature_db`

loicgasser:
scan db is user:

1. To determine which bodId is associated to which geotable
2. To determine which layer has duplicated primary keys accross the tables

chsdi/models/__init__.py
procrastinatio: OK, I see

loicgasser:
thanks @procrastinatio

PR #2459, opened 2017-02-14
Reflect DB changes and preserve catalog order

loicgasser:
We are not dealing with `None` values in the orderKey field anymore in the catalogs. (It does simplify the logic a bit)
Instead orderKey now defaults to 0.
Instead of relying on the fact that we have None values in the children to order them alphabetically or by order key. We make sure all order keys are different, if not the case we order alphabetically.

[Test Link](mf-chsdi3.dev.bgdi.ch/gal_order/rest/services/inspire/CatalogServer?lang=fr)

### DIFF
--- a/chsdi/views/catalog.py
+++ b/chsdi/views/catalog.py
@@ -24,6 +24,7 @@ def tree_data(G, root, attrs, meta):
         raise nx.NetworkXError('Attribute names are not unique.')
 
     def add_children(n, G):
+        distinct_order_keys = set()
         order_key = lambda x: x['orderKey']
         label_key = lambda x: x['label'].replace(u'Ü', u'U').replace(u'Ä', u'A')
         nbrs = G[n]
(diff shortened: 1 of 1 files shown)


procrastinatio:
Python is OK, but the rest is dead black magic.

loicgasser:
Have a look at:

**https://networkx.github.io/**

useful when dealing with graphs.

procrastinatio:
I don't mean the Python code, but the buisness with the catalog sorting, reordering, etc.

chsdi/views/catalog.py
```diff
@@ -24,6 +24,7 @@ def tree_data(G, root, attrs, meta):
         raise nx.NetworkXError('Attribute names are not unique.')
 
     def add_children(n, G):
+        distinct_order_keys = set()
         order_key = lambda x: x['orderKey']
         label_key = lambda x: x['label'].replace(u'Ü', u'U').replace(u'Ä', u'A')
```
loicgasser: I think I did try something like that, I recall it didn't work out, but feel free to address this weak point.

loicgasser:
Thanks for the review.

PR #2462, opened 2017-02-16
added richtfunkverbindungen

rebert:
[try](https://mf-geoadmin3.int.bgdi.ch/?wms_url=%2F%2Fwms-bgdi.dev.bgdi.ch%2Fltrea_test&api_url=%2F%2Fmf-chsdi3.int.bgdi.ch%2Fltret_bakom_richtfunk&topic=ech&lang=de&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege,ch.bakom.richtfunkverbindungen&layers_visibility=false,false,false,false,true&layers_timestamp=18641231,,,,&X=200439.31&Y=611611.95&zoom=4)


rebert:
@loicgasser : 
1. Seems to be slow when JSON is actvated
2. Tooltip: Profile does not work (and make no sense in this context)

loicgasser:
profile does work, it's a setup issue I think.
https://s.geo.admin.ch/7172a3e37d

loicgasser:
3. is a validation issue, we could fallback on a default style if not available or incorrect.
1. It looks pretty good to me, even if it's probably the biggest geojson we had so far. it's quite fast from here.
Is it still WIP?

loicgasser:
we should also still add a legend

rebert:
@loicgasser can merge -> deploy 22. March

rebert:
@loicgasser the layer is now a regular PostGIS layer ...

loicgasser:
LGTM, yes it's probably for the best with so many lines. thanks

PR #2483, opened 2017-03-07
Making sphinx build break on error

procrastinatio:
* Makefile breaks on doc error
* No warning, only errors

Fix https://github.com/geoadmin/mf-chsdi3/issues/2247


loicgasser:
Great thanks a lot for that.

PR #2486, opened 2017-03-08
Switch to new prod instances

gjn:
Switch from m3.medium to c3.xlarge (which cost less until september)


gjn:
I'll automerge and deploy latest snapshot to these instances.

PR #2494, opened 2017-03-15
Add a css class to htmlpopup specific to a layer

oterral:
Fix  geoadmin/mf-geoadmin3#3700 

[Test](https://map.geo.admin.ch/?topic=ech&lang=de&bgLayer=voidLayer&layers=ch.swisstopo-vd.amtliche-vermessung&X=228418.09&Y=607177.55&zoom=11&api_url=%2F%2Fmf-chsdi3.int.bgdi.ch%2Fteo_fixes)

### DIFF
--- a/chsdi/templates/htmlpopup/base.mako
+++ b/chsdi/templates/htmlpopup/base.mako
@@ -12,6 +12,7 @@
   isExtended = c.get('isExtended')
   isIframe = c.get('isIframe')
   isGridLayer = c.get('isGridLayer')
+  c['htmlpopup_class'] = c['layerBodId'].replace('.', '')
  %>
 
 % if isExtended or isIframe:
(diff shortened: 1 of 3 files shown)


chsdi/templates/htmlpopup/base.mako
```diff
@@ -12,6 +12,7 @@
   isExtended = c.get('isExtended')
   isIframe = c.get('isIframe')
   isGridLayer = c.get('isGridLayer')
+  c['htmlpopup_class'] = c['layerBodId'].replace('.', '')
```
gjn: A little protection in case 'layerBodId' is not defined would be good....

loicgasser: can't really happen it's done much earlier https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/lib/validation/features.py#L50

PR #2508, opened 2017-03-30
Contextual sessions

loicgasser:
Fix for https://github.com/geoadmin/mf-chsdi3/issues/2501

I looked into the reason why we sometimes have an error when deploying to prod, an it trickles down to re-using the same connection too many times (thousands of times). So it's safer to introduce a contextual session. It should solve our problems.


gjn:
Very nice.

I'll see the effect on next deploy. Thanks.

Issue #2473, opened 2017-02-23
Special characters in tooltip


AFoletti:
In some cases special characters ( :registered: in the present case) are not rendered coherently across tooltip attributes and tooltip/extended tooltip.

Example:
[this object](https://map.geo.admin.ch/?ch.bfe.energieforschung=178&lang=de&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.bfe.energieforschung,ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=true,false,false,false,false&layers_timestamp=,18641231,,,&X=276728.24&Y=691406.87&zoom=9) has a :registered: symbol in both Title and Description. It's good in Description but not in Title.
Even more interesting, the very same character for the very same attribute (Description) is rendered incorrectly in the [extended tooltip](https://api3.geo.admin.ch/rest/services/ech/MapServer/ch.bfe.energieforschung/178/extendedHtmlPopup?lang=de).

Any idea?

loicgasser:
Encoding issue probably

AFoletti:
yes, I tend to agree. But why is the very same attribute value rendering once OK and once NOK in the same mako?

loicgasser:
because the characters in the metadata file are different, I'll figure it out.

loicgasser:
who integrated this layer?

AFoletti:
What do you mean by "the caracters in the metadata file"?
About integration: It's very old... first integration was made by me in 2013

loicgasser:
yes I thought it was a kgs layer at first. my bad.

PR #2509, opened 2017-03-30
Escape special character

loicgasser:
fix for https://github.com/geoadmin/mf-chsdi3/issues/2473

image


loicgasser:
trivial self merge

PR #2526, opened 2017-04-19
Always build with the correct branch

loicgasser:
If this variable was previously exported, it is possible to override GIT_BRANCH, this PR makes sure we always build with the good branch.


oterral:
lgtm

Issue #2336, opened 2016-11-16
Generate empty.po automatically on translate


loicgasser:
YAML http://pyyaml.org/wiki/PyYAMLDocumentation
Python built-in https://docs.python.org/2/library/functions.html
os python https://docs.python.org/2/library/os.html
https://docs.python.org/2/library/os.path.html
https://docs.python.org/2/library/__main__.html
https://docs.python.org/2.7/library/contextlib.html

- Add script usage https://docs.python.org/2/library/sys.html or https://docs.python.org/2/library/argparse.html
- Generate empty.po automatically
- Refactor translation script in general
- move it to scripts

Compare:
https://github.com/geoadmin/mf-chsdi3/compare/master...empty_po?expand=1

PR #2527, opened 2017-04-19
Autotranslate without empty.po

loicgasser:
4'000 translations not used in chsdi, some stuff that doesn't make sense as well.

I think there are a couple that are not used. Maybe the chance for a cleanup as well.


loicgasser:
@gjn This PR is ready. Only thing is in the PR description, it introduces a lot of extra translations. I think that we may have a lot of unused translations as well.
These are used in chsdi and wms. @ltclm is there a way to know which ones are in the WMS?

Otherwise review can already be performed as this is technically ready.

gjn:
This LGTM. I think we can live with the extra stuff that is not needed. Should have a great impact on performance and size.

loicgasser:
Ok, I'll update wiki and send an email to the integrators then.

loicgasser:
https://github.com/geoadmin/mf-chsdi3/issues/2336

Issue #1315, opened 2015-03-06
Correct class names in models


AFoletti:
As @procrastinatio pointed out at least a couple of times, our naming scheme for the classes in the models is not consistent.

@kandre and I we will scroll every model and adapt them to the CapWords convention

loicgasser:
CapWords is not best partice for naming tables or views, please don't

* SQL keywords: upper case
* names (identifiers): lower case with underscore

AFoletti:
ok. But then, should we make those models consistent or not? And if yes, which convention should we use?

loicgasser:
Yes would be nice to have naming convention. I support the effort.

AFoletti:
@loicgasser could you please make a suggestion of a naming convention?

loicgasser:
Ok I will.

procrastinatio:
Simply use [PEP8](https://www.python.org/dev/peps/pep-0008/#id34), especially *Naming Styles*

loicgasser:
We use the CapWords convention class names in python.

PR #2537, opened 2017-04-27
Honor caps word convention for class names

loicgasser:
Fix for https://github.com/geoadmin/mf-chsdi3/issues/1315

Now it should be clear for everyone how to name classes when  defining a model.


gjn:
+1

Issue #2538, opened 2017-04-28
[layers] changed alternative titles for several ch.bazl* layers


mariokeusen:
Just imported new alternative titles (mainly new RM translations) from geocat to the bod for almost all ch.bazl layers. Could you please consider that for the next deploy (layer search).
Thanks in advance

loicgasser:
Yes it will be updated automatically at the same time we deploy the BOD.

loicgasser:
Fixed

PR #2539, opened 2017-04-28
fixed units

AFoletti:
hotfix for MM on monday

### DIFF
--- a/chsdi/templates/htmlpopup/statistikwasserkraftanlagen.mako
+++ b/chsdi/templates/htmlpopup/statistikwasserkraftanlagen.mako
@@ -106,7 +106,7 @@
   <tr>
     <td class="cell-meta">${_('ch.bfe.statistik-wasserkraftanlagen.performanceturbinemaximum')}</td>
     % if c['attributes']['performanceturbinemaximum'] != None:
-        <td class="cell-meta">${round(c['attributes']['performanceturbinemaximum'],2) or '-'}&nbsp;GWh</td>
+        <td class="cell-meta">${round(c['attributes']['performanceturbinemaximum'],2) or '-'}&nbsp;MW</td>
     % else:
         <td class="cell-meta">-</td>
     % endif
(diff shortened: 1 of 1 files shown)


gjn:
@loicgasser @AFoletti Please take care of the emergency deploy on monday. As master has been updated since last deploy, I propose to create a branch based on latest release tag and apply this patch and deploy the branch.

Also, mf-geoadmin3 would need to be re-deployed to invalidate the cache (assure a new version is created).

Ok?

AFoletti:
BOD has not changed, so we can use the last snapshot

loicgasser:
@chsdi tooltips are not cached in geoadmin.
No need to deploy geoadmin, if BOD has not changed.

Just chsdi and with the lubis fix, why not?
I am tempted to deploy now.

loicgasser:
@gjn CHSDI on int. Snapshot is: **201704281531**
I'd rather have the lubis fix as well. not much as changed and it will also fix some translations for BAZL.
https://github.com/geoadmin/mf-chsdi3/issues/2538

gjn:
No deploys on friday.

Deploy of geoadmin is needed because we need to change the application version. The application version is used to created cached requests to our api, e.g. https://api3.geo.admin.ch/1704261351/rest/services/ech/MapServer/ch.bfe.statistik-wasserkraftanlagen/202430/htmlPopup

The number after the domain is the application version. All versioned requests are cached 1 year.

loicgasser:
> No deploys on friday.

No I won't

> Deploy of geoadmin is needed

Yes you're right, but Varnish doesn't cache much

> I insisted on being able to re-deploy a snapshot with a new version...

And you got it, DEEP_CLEAN=true ...

gjn:
Thanks for deploying on Monday!

Varnish is not the only thing caching stuff. Other proxies (BIT proxy) and clients/browsers do as well.

We could discuss giving up this mechanism, so things like these don't need deploy of map.geo (I don't really like it myself). Maybe let api3 define it's own version and the map should use that. So before first api request where caching is wanted, the map asks for latest version to use (api3.geo.admin.ch/currentversion) and will use that for all requests. Would be best of both worlds, I guess.

loicgasser:
Yes, that could be an option.
Do you want me to deploy geoadmin with the new cache headers for index.html ?

AFoletti:
As info: MM is at 10:00 monday morning.

PR #2545, opened 2017-05-17
Add feedback layer

loicgasser:
@p1d1d1 Could you ask someone to provide the legends?

geojsons are incorrect, so we can't display them properly at the moment. Please refer to the issue in JIRA.


p1d1d1:
Spec for legends? Dimensions etc. 
Do you have an example?

loicgasser:
[Test Link](https://mf-geoadmin3.dev.bgdi.ch/?api_url=%2F%2Fmf-chsdi3.dev.bgdi.ch%2Fgal_feedbacklayer&layers=ch.swisstopo.meldungen-karten_geodaten&lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&X=156875.00&Y=681750.00&zoom=2)

gjn:
Is there a reason we use `*/test/*` in path to file? Why not use correct path?

p1d1d1:
so was organized/decided by Alain for testing purposes, but we should get rid of this "special" cases, IMHO

gjn:
I would change it now.

p1d1d1:
The standard path seems to me empty. You should check with Andi

rebert:
unrelated build error due to a sphinx index not updated

PR #2555, opened 2017-05-30
Update legend_rm ch.bafu.hydroweb-warnkarte_regional

Sgachet:


AFoletti:
ah sorry, I should have seen that earlier when we did the transparency. The legend is too broad. Max is 480px

Sgachet:
Yep, right !
Now the correct legend is ready 
@loicgasser, you can merge ;)

Sgachet:
@AFoletti It's ok now ?

@loicgasser Then you can merge it :)

loicgasser:
@Sgachet 
There is Makefile target for the legends, that does it for you.

```
"- legends            Downloads and optimizes all WMS legend images (make legends BODID=ch.foo WMSSCALELEGEND=)"
```

I think I sent you an email at the time.

AFoletti:
@loicgasser about the make target doing the legends: what does it do if the GetLegend returns an image with a long text such as this one? Simply resize the picture or is clever enough to make a multiline?

loicgasser:
simple resize, mapserver shall be clever enough. But as this is a geojson (didn't notice at first) and that it is not integrated in the WMS, we can't use it directly.

AFoletti:
No, mapserver is not. At least, not our current configuration. @ltclm maybe we could improve that?

PR #2590, opened 2017-07-05
swisstlm3d strassen, gewaessernetz, eisenbahnnetz, uebrigerverkehr

alcapat:
4 new layers for swisstlm3d with tooltip


alcapat:
Missing legend in png and translations for tooltip exist only in DE and FR for the moment

PR #2593, opened 2017-07-11
Remove unused attr

loicgasser:


gjn:
LGTM.

But topic tests fail.

loicgasser:
Thanks, you re right I need to adapt one test.

PR #2614, opened 2017-07-27
Use pyproj instead of gdal

loicgasser:
As we are using gdal in a single function `transform_coordinate`, it is better to use something else, much much lighter like pyproj.

Build is now shorter. (will tell you exactly how much in a sec)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,7 +106,7 @@ user:
 .PHONY: all
 all: setup chsdi/static/css/extended.min.css templates potomo rss lint fixrights
 
-setup: .venv gdal node_modules .venv/hooks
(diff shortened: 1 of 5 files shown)


loicgasser:
Build time now:
2m12.980s

Build time before:
2m13.256s

not much of a change but cleaner install and code for the same result.

gjn:
Both GDal and pyproj are using the same transformation?

Which?

gjn:
I think it's grid based transformation.

loicgasser:
yes grid based, test returns exactly the same result

gjn:
+1

Unfortunately, now a conflict because of the other PR. But once solved, please merge this.

PR #2621, opened 2017-08-09
Fetch ids in integration tests dynamically

loicgasser:


chsdi/tests/integration/__init__.py
```diff
@@ -32,6 +35,19 @@ def tearDown(self):
...
+    def getSession(self):
+        session = scoped_session(sessionmaker())
+        yield session
+        session.close()
+
+    def getRandomFeatureId(self, bodId):
+        models = models_from_bodid(bodId)
+        with self.getSession() as session:
+            query = session.query(models[0]).limit(1)
+            reslt = query.one()
```
gjn: Is this truly random?

loicgasser: it's random from a user point of view, as in I don't care just give me some id. You'd need ORDER BY RANDOM() to really get a different value every time.

gjn: Yes, I had that once. Works fine with small datasets, explodes with big datasets :=), that's why I asked.

Truly random would be better, but can live with this. Maybe there's a way to randomize the first 1000 only?

loicgasser:
```sql
SELECT sub.id FROM (SELECT id FROM table LIMIT 1000) AS sub ORDER BY RANDOM() LIMIT 1
```

The first 1000 will always be the same, and we choose randomly one out of a 1000

gjn: Could be worth it. But as said, not major IMO.

loicgasser: Sure I can do it, we need to change all our tests where we have hardcoded ids in the integration tests in the future as well. Let's make it 500, build is long enough already.

loicgasser:
I'll megre for Jenkins tonight and change that tomorrow.

chsdi/tests/integration/__init__.py
gjn: +1

PR #2624, opened 2017-08-10
Randomly select a feature

loicgasser:


loicgasser:
1 build error related to ongoing work with treasure hunt layer.

Issue #2627, opened 2017-08-11
Prepare for ToD changes (WMTS GetCap)


gjn:
https://github.com/geoadmin/tool-wms/pull/91

loicgasser:
https://github.com/geoadmin/mf-chsdi3/pull/2635

PR #2628, opened 2017-08-11
Work with new ToD based backend

gjn:
This is for https://github.com/geoadmin/mf-chsdi3/issues/2627, see discussion here https://github.com/geoadmin/tool-wms/pull/91

It also needs small extension to gatilegrid: https://github.com/geoadmin/lib-gatilegrid/pull/15 which needs to be merged and uploaded for this branch to work.
ping @loicgasser 

Also, it doesn't work for all layers. `ch.swisstopo.swisstlm3d-karte-grau.3d` and `ch.swisstopo.swisstlm3d-karte-farbe.3d` are now part of the getCap views in the bod. I think that this is wrong as we don't publish this officially now - and don't want to in the future. Can we remove them from the this view?
ping @ltclm 

Because of the above, branch can't be deployed currently - but it can be tested [here](http://mf-chsdi3.dev.bgdi.ch/ltjeg) under my private account.

We should also verify the content of the layersConfig service - which I didn't do yet.


gjn:
Here's the diff of current dev wmtsgetcap and with the new tod bod:
https://gist.github.com/gjn/d38977a5b0c9043217bcb34c5d9173f0/revisions

`ch.bazl.projektierungszonen-flughafenanlagen` has been removed.

[Zeitreihen layers changed from jpgpng to png only](https://gist.github.com/gjn/d38977a5b0c9043217bcb34c5d9173f0/revisions#diff-a9505927dd897cd4292416edbf089f2fR7837)

`ch.bafu.wasser-teileinzugsgebiete_2` has different [max resolution and different timestamps](https://gist.github.com/gjn/d38977a5b0c9043217bcb34c5d9173f0/revisions#diff-a9505927dd897cd4292416edbf089f2fR8303)

Otherwise, it looks the same.

ltclm:
great thank you!
everything in tileset table will be visible in the GetCap of the supported SRS.
We could add a boolean column ``publish`` or ``visible`` to the tileset table for the published getcap Documents, if set to false the layer won't appear in any SRS GetCap but it will still be available for ToD, if set to true it will appear in ALL SRS GetCap Documents.
I recommend to not introducte a perLayer/SRS/publish/(True|False) configuration now.

gjn:
I agree with the new column and that it applies to all srs. I think the .3d layer is mainly needed in the layersConfig (which I'm comparing now).

What about the other differences?

ltclm:
**ch.bazl.projektierungszonen-flughafenanlagen** is one of the layers which has been switched to wms and which has not been accessed anymore for the last 3 months, according to https://github.com/geoadmin/tool-wms/pull/91

**ch.bafu.wasser-teileinzugsgebiete_2** This is an issue in our getcap query, multiple time dimensions with different max resolutions are not supported. Due to the change from max_zoomlevel to max_resolution the result of the query is different. the old query returns the 21781_25 entry 20121231, the new query returns the 21781_26 entry with timestamp 20120701.

I have removed the old timestamp in the migration script.
the results should now be identical.

gjn:
Very good. Thanks.

ltclm:
i have added a published column to the tileset table, the layers ch.swisstopo.swisstlm3d-karte-grau.3d and ch.swisstopo.swisstlm3d-karte-farbe.3d are now hidden.
https://github.com/geoadmin/tool-wms/pull/91/commits/4068463adc875fb91f036f1301a5e91bf6e56b87

chsdi/models/bod.py
```diff
@@ -206,12 +205,16 @@ class GetCap(object):
     datenherr = Column('datenherr', Unicode)
     wms_kontakt_abkuerzung = Column('wms_kontakt_abkuerzung', Unicode)
     wms_kontakt_name = Column('wms_kontakt_name', Unicode)
-    zoomlevel_min = Column('zoomlevel_min', Integer)
-    zoomlevel_max = Column('zoomlevel_max', Integer)
```
loicgasser: never ever do that

ltclm: the bod migration script pr has been open for over a month...

loicgasser:
It will fail as new structure is not deployed yet. waiting on @ltclm before deploying the code to int.

chsdi/models/bod.py
loicgasser: we can remove from the code but not from db

ltclm:
i wont be able to fix the views before 830pm today...
i will put back zoomlevel_min zoomlevel_max and tilematrixsetid.

loicgasser:
np there no rush

ltclm:
@loicgasser i have added the legacy columns
* zoomlevel_min
* zoomlevel_max
* tile_matrix_set_id

to the following views:

* bod_master.re3.view_tileset_concatenated
* bod_master.re3.view_bod_wmts_getcapabilities_de
* bod_master.re3.view_bod_wmts_getcapabilities_fr
* bod_master.re3.view_bod_wmts_getcapabilities_it
* bod_master.re3.view_bod_wmts_getcapabilities_rm
* bod_master.re3.view_bod_wmts_getcapabilities_en

They will be removed from these views as soon as https://github.com/geoadmin/mf-chsdi3/issues/2627 has been merged.

loicgasser:
And deployed. Sorry but I really want to avoid down time on this service. We ll test it well tomorrow. I don t think we need the resolution mapping in db anymore, it the transformation of degrees to meters in handled in gatilgrid.

gjn:
yes, 2-step deployment.

PR #2642, opened 2017-09-05
Run flake8 tests on Python 2 and Python 3

cclauss:
Helps to find syntax errors and undefined names in the code.  Runs the [flake8](http://flake8.readthedocs.io) tests on Python 3 in __allow_failures__ mode.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,21 @@
+language: python
+cache: pip
+python:
+    - 2.7
+    - 3.6
+matrix:
+    allow_failures:
+        - python: 3.6
+install:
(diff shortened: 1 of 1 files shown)


loicgasser:
we're using Jenkins, we have a pre-commit hook in the project that won't allow a commit unless our flake8 rules are respected.
Do you want to make this project py3 compatible?
Therefore, I would avoid adding travis on top of it.

cclauss:
Given that Python 2 end-of-life is in 245 days, would it make sense to add a Python 3 run to Jenkins?

Issue #2638, opened 2017-08-30
Error 500 on ch.swisstopo.geologie-geologischer_atlas


cedricmoullet:
http://api3.geo.admin.ch/rest/services/api/MapServer/ch.swisstopo.geologie-geologischer_atlas/GA25-PK-140/htmlPopup

loicgasser:
thank you for the heads up.
shop side effect, the shop model was also registered..
This request should not be triggered on this layer as Id for this layer is an integer.
Now such an error will return a 400 error instead.

PR #2644, opened 2017-09-19
update GA25 related to https://github.com/geoadmin/mf-chsdi3/issues/2638

rebert:


ltclm:
@rebert why a new view view_grid_ga25? 
we could have added a new column bgdi_id to the existing view view_grid_ga25_shop instead?

rebert:
@ltclm When I had to add bgdi_id related to the change of CHSDI I would have had to change about 7 Views...

PR #2660, opened 2017-10-16
decomission bundesinv_anhoerung of several bafu layers

rebert:


loicgasser:
for when is it supposed to be decommissioned @rebert

rebert:
can be merged already now

PR #2669, opened 2017-10-25
Allow to remove branch on integration

oterral:


scripts/delete_branch.sh
```diff
@@ -1,44 +1,64 @@
...
-if [ "$#" -ne 1 ]; then
-        echo "No branch name given"
-        echo "Possible values:"
-        for b in ${BRANCHES};
-            do
-               echo ${b}
-            done
-        exit 0
+if [ $# -ne 2 ]; then
+    echo "No branch name given, use parameter BRANCH_TO_DELETE=your_branch_name"
```
procrastinatio: Message is misleading: yo want two parameters

procrastinatio: And I would also test the number of args at the very top

oterral: I can't do it at the very top I need all the previous variables to display the list of branches available

oterral:
@procrastinatio doc improved

scripts/delete_branch.sh
```diff
@@ -1,44 +1,71 @@
 #!/bin/bash
```
loicgasser: we could add set -e

PR #2671, opened 2017-10-26
Add new version for ol3 api 4.3.2(LV03) and 4.42(LV95)

oterral:
Remove also the loading of ol.css because it is already imported in ga.css


gjn:
+1!

I think we should update the documentation, add the new versions and specifically say that the new versions are for LV95.

oterral:
Yes right.

PR #2676, opened 2017-10-31
Use LV03 for zeitreihen

oterral:
@ltclm  could it be the solution without modifying the db ?


loicgasser:
Closing since DB has been changed accordingly thanks @ltclm

ltclm:
would have been an option, thanks.
good to know that it is possible to define/override the source srid somewhere

PR #2680, opened 2017-10-31
Fix getcap pointing to TOD

loicgasser:
does this makes any sense to anybody? Why do we have that?
Fix for https://github.com/geoadmin/service-proxywms/issues/150


procrastinatio:
@loicgasser mapproxy was used to provide a WMTS from a WMS.

loicgasser:
yes I guess this makes sense from a MapProxy point of view.

loicgasser:
But this means that it probably didn't work in the JS API as well.

gjn:
+1

PR #2681, opened 2017-10-31
Remove mapproxy ref, everything points to wmts(1-9)

loicgasser:
After LV95 or before?

ping @gjn @procrastinatio

This assumes that public wmts will always be wmts[20-24]


gjn:
I don't understand this. Why everything wmts20?

I wouldn't change the doc at all for the moment. We will consolidate the WMTS addresses at a later time.

gjn:
I think our official domain for the getcapabilities should be: wmts.geo.admin.ch. This should probably also be the wmts url in the getcapabilities document.

Then for the tiles, we offer ranges wmts1-wmts9.

loicgasser:
yes, it's maybe better to keep those instead. Anyway we should always use the same range since TOD is in the place. We'll have to change the codepens to be consistent, as well as a lot of those are still pointing to 20-24.

gjn:
Yes, we need to get rid of all the old ones. Cleaning up will be big early 2018.

loicgasser:
adapted according to your comments.

loicgasser:
@gjn @procrastinatio please review

Issue #2665, opened 2017-10-22
Layer attributes doesn't work for mobility


gjn:
https://api3.geo.admin.ch/rest/services/api/MapServer/ch.mobility.standorte

Results in 500 error. This shouldn't happen.

gjn:
I think the problem is that this layer as multiple different geometry field definitions. And the rest of the code relies on having exactly 1 geometry - the_geom.

See model.

PR #2683, opened 2017-11-01
Fix attrname service

loicgasser:
Fix for https://github.com/geoadmin/mf-chsdi3/issues/2665


gjn:
This looks easy :=)

Wasn't able to find it. Thanks!

PR #2691, opened 2017-11-08
New regbl addresses

loicgasser:
Works with https://github.com/geoadmin/service-sphinxsearch/pull/343/files

This is an early version of the integration of  the new addresses in the search.

For now we only have  the addresses in the `locations_preview` service, it is based on the very first gdb created by Elias which I imported in `kogis_master.bfs.gwr_addresses` and I created the view `bfs.adressen_sphinx_preview` for sphinx.

That makes the search language specific which is somewhat incompatible with what we have at the moment.

In geoadmin for instance we always provide the `lang` parameter which will determine the language of the label and not apply filters like it is requested for the regBL project.

Therefore we have to introduce a new parameter to support regBL use case and still be backward compatible with geoadmin.
I propose we use the parameter `searchLang` to support both use cases.

Language rules will be applied as follow:
`lang=de`: current behaviour. search is performed in all languages and labels are returned in german
`searchLang=de`: we search only in the german fields and we return the labels in german

As explained earlier I still needed to add a generic detail field (search text) to keep the current behavior of the search as we want to search in all the addresses all the time.

[1]
We currently have overall 18'112 addresses for which we have multiple languages.

The only combinations we have are:

de-fr: 17'020
de-rm: 1'092

-> 17'020 + 1'092 = 18'112 -> CQFD

So the general detail field (the search text) would hold one of the 4 languages or a combination of at most 2 languages.
So at the moment we only have one language in the generic detail field.
(if de -> de elif fr -> fr elif it -> it else -> rm) 
Combining different address langs in the detail field may have undesired side effects on the search ranking.

[1]
```sql
SELECT count(a.sum_of_notnulls) as nb_multiple_langs
FROM
(SELECT (
 CASE WHEN stn_strname_de is NULL THEN 0 ELSE 1 END +
 CASE WHEN stn_strname_fr is NULL THEN 0 ELSE 1 END +
 CASE WHEN stn_strname_it is NULL THEN 0 ELSE 1 END +
 CASE WHEN stn_strname_rm is NULL THEN 0 ELSE 1 END
) as sum_of_notnulls
FROM bfs.gwr_addresses) a
WHERE a.sum_of_notnulls > 1
```
I will also have to introduce some redundancy in the other indices that are involved in the locations search as I can't combine distributed indices that have a different structure. (thus indexing each detail field 5 extra times, but I may be able to tweak it)

De only
[Test Link Seftigenstrasse De](http://mf-chsdi3.dev.bgdi.ch/gal_regbl/rest/services/api/SearchServer?searchText=seftigenstrasse&type=locations_preview&searchLang=de)
[Test Link Seftigenstrasse Fr](http://mf-chsdi3.dev.bgdi.ch/gal_regbl/rest/services/api/SearchServer?searchText=seftigenstrasse&type=locations_preview&searchLang=fr)

De and Fr (entry in both french and german see feature_id):
[Test Link boujean Fr](http://mf-chsdi3.dev.bgdi.ch/gal_regbl/rest/services/api/SearchServer?searchText=boujean&type=locations_preview&searchLang=fr)
[Test Link bözingenstrasse De](http://mf-chsdi3.dev.bgdi.ch/gal_regbl/rest/services/api/SearchServer?searchText=bözingenstrasse&type=locations_preview&searchLang=de)

ping @gjn @cedricmoullet @ltclm

### DIFF
--- a/chsdi/lib/validation/search.py
+++ b/chsdi/lib/validation/search.py
@@ -2,7 +2,7 @@
 
 from pyramid.httpexceptions import HTTPBadRequest
 
-from chsdi.lib.helpers import float_raise_nan
+from chsdi.lib.helpers import float_raise_nan, shift_to
 from chsdi.lib.validation import MapNameValidation
 
 MAX_SPHINX_INDEX_LENGTH = 63
@@ -11,8 +11,9 @@ MAX_SEARCH_TERMS = 10
 
 class SearchValidation(MapNameValidation):
 
-    def __init__(self):
+    def __init__(self, request):
         super(SearchValidation, self).__init__()
+        self.availableLangs = request.registry.settings['available_languages'].split(' ')
         self.locationTypes = [u'locations', u'locations_preview']
         self.layerTypes = [u'layers']
         self.featureTypes = [u'featuresearch']
@@ -29,6 +30,7 @@ class SearchValidation(MapNameValidation):
         self._origins = None
         self._typeInfo = None
         self._limit = None
+        self._searchLang = None
 
     @property
     def searchText(self):
@@ -74,6 +76,10 @@ class SearchValidation(MapNameValidation):
     def limit(self):
         return self._limit
 
+    @property
+    def searchLang(self):
+        return self._searchLang
+
     @featureIndexes.setter
     def featureIndexes(self, value):
         if value is not None and value != '':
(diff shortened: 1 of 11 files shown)


loicgasser:
Just went to Kibana, it seems that http://geodata4edu.ethz.ch/ are still using `locations_preview` after 2+ years. ping @gjn . should I use a different string? the only one I can see in the logs. Maybe it won t affect them though. Can't tell you need to login.

loicgasser:
mmmm, can't connect to sphinx server anymore. avoid cleaning up stuff we use, after Jenkins, now search....

PR #2693, opened 2017-11-13
update legends 4 skitouren and schneeschuhrouten

rebert:


oterral:
PNGs should stay transparent.

rebert:
@oterral I only delivered as ordered...

rebert:
@AFoletti : beside of the transparency nothing changed, shall we keep the old transparent legends?

loicgasser:
Why would you order non transparent png? Sounds odd. We have standards for legends, I would like to keep those.

AFoletti:
Not every customer is aware of the transparency rule (or remembers it). We remind it if we catch the issue but this slipped past.
I'd say to keep the old ones if they are the same, as suggested by @rebert

rebert:
done

loicgasser:
When should it be deployed?

AFoletti:
22.11

Issue #2698, opened 2017-11-15
Service 'releases' broken for LUBIS


procrastinatio:
Working for `ch.swisstopo.zeitreihen`
```
curl -X POST "http://api3.geo.admin.ch/rest/services/all/MapServer/ch.swisstopo.zeitreihen/releases?geometry=660850.0%2C180300.0&sr=21781&imageDisplay=2851%2C1884%2C256.0&geometryType=esriGeometryPoint&mapExtent=590118.055556%2C133556.944444%2C731581.944444%2C227043.0"
{"results": ["18641231", "18691231", "18711231", "18731231", "18881231", "18941231", "19001231", "19031231", "19051231", "19121231", "19231231", "19321231", "19481231", "19541231", "19601231", "19651231", "19691231", "19761231", "19801231", "19861231", "19931231", "19991231", "20071231", "20131231"]}3d [root@ip-10-220-5-68:~] # 
```
But not for all LUBIS layers
```
curl -X POST "http://api3.geo.admin.ch/rest/services/all/MapServer/ch.swisstopo.lubis-luftbilder_farbe/releases?geometry=660850.0%2C180300.0&sr=21781&imageDisplay=2851%2C1884%2C256.0&geometryType=esriGeometryPoint&mapExtent=590118.055556%2C133556.944444%2C731581.944444%2C227043.0"
```
Error log

```
13793 [Tue Nov 14 21:28:36.391619 2017] [wsgi:error] [pid 6298] [remote 10.220.4.223:30983] AttributeError: type object 'Bildstreifen' has no attribute 'bgdi_order'
13820 [Tue Nov 14 21:28:36.417088 2017] [wsgi:error] [pid 6298] [remote 10.220.4.223:29959] AttributeError: type object 'LuftbilderSwisstopoFarbe' has no attribute 'bgdi_order'             
 ```

(https://github.com/geoadmin/mf-chsdi3/blob/3a15824ec59cedacaba27c48f958f294633bdffa/chsdi/views/features.py#L785

ltclm:
that might have never worked for time-enabled layers without bgdi_order attribute in the model.
Thanks for the fix!

PR #2699, opened 2017-11-15
LUBIS release

procrastinatio:
Testing for missing attributes. Years list is still empty

See https://github.com/geoadmin/mf-chsdi3/issues/2698

### DIFF
--- a/chsdi/tests/integration/test_features_service.py
+++ b/chsdi/tests/integration/test_features_service.py
@@ -319,7 +319,9 @@ class TestFeaturesView(TestsBase):
         resp.mustcontain('cb_({')
 
     def test_feature_big_but_good(self):
-        resp = self.testapp.get('/rest/services/all/MapServer/ch.swisstopo.geologie-geocover/1080284', params={'geometryFormat': 'geojson'}, status=200)
+        bodId = 'ch.swisstopo.geologie-geocover'
+        featureId = self.getRandomFeatureId(bodId)
+        resp = self.testapp.get('/rest/services/all/MapServer/%s/%s' % (bodId, featureId), params={'geometryFormat': 'geojson'}, status=200)
         self.assertEqual(resp.content_type, 'application/json')
         self.assertIn('geometry', resp.json['feature'])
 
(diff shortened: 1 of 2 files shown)


ltclm:
years list is probably empty because of the lv95 migration.
@loicgasser has the releases service been migrated to lv95 to?
the zeitreihen data for releases is still lv03, thats why it is still working. the lubis data is now in lv95.

loicgasser:
no releases is a zeitreihen service only.
Time series in print was also only designed for zeitreihen.
People forget so much...

layers config right here tells what years there are. On top of it lubis are points, and release service is intersecting with a point and not a bbox.

@procrastinatio I propose that you pass in the timestamps to the print and that you raise a 400 if the release service is used with something else than zeitreihen.


```
ch.swisstopo.lubis-bildstreifen: {
shop: [
"mapsheet"
],
wmsUrl: "https://wms.geo.admin.ch",
background: false,
searchable: true,
serverLayerName: "ch.swisstopo.lubis-bildstreifen",
highlightable: true,
timeEnabled: true,
attributionUrl: "https://www.swisstopo.admin.ch/fr/home.html",
tooltip: true,
label: "Bandes d'images aériennes swisstopo",
singleTile: false,
timestamps: [
"2017",
"2016",
"2015",
"2014",
"2013",
"2012",
"2011",
"2010",
"2009",
"2008",
"2007",
"2006",
"2005"
],
type: "wms",
topics: "api,ech,luftbilder,swisstopo,wms-bgdi",
attribution: "swisstopo",
format: "png",
chargeable: false,
wmsLayers: "ch.swisstopo.lubis-bildstreifen",
gutter: 20,
timeBehaviour: "all",
hasLegend: true,
queryableAttributes: [
"id"
]
}
```

procrastinatio:
@loicgasser @ltclm 
* `releases` service works fine in both lv95 and lv03
* See https://github.com/geoadmin/mf-geoadmin3/issues/2099. If we want to print timeserie for LUBIS layer, we need to know wich years to print in a given extent. It is a **new** feature. Nothing broken here.

procrastinatio:
I guess it is a complexe issue, because the `flugjahr` is not what you see on the map at `published year`, it was usually 3 years later.

loicgasser:
I fixed the build and added the test for zeitreihen.

PR #2703, opened 2017-11-15
geologie geocover metadata new attribute in shop tooltip

alcapat:


loicgasser:
when is it supposed to be deployed?

AFoletti:
22.11

2017-11-17 15:10 GMT+01:00 Gasser Loïc <notifications@github.com>:

Issue #2687, opened 2017-11-01
Remove model and mako (+unused translations -> edit ms_item and make translate)


loicgasser:
ch.swisstopo-vd.geometa-standav

@mariokeusen @AFoletti 

do you know if this layer is supposed to have a tooltip? Because we have a model in CHSDI but it is marked without tooltip in the layers config.

loicgasser:
ping @mariokeusen @AFoletti BMD should have this info.

AFoletti:
MuM is not (yet) a reliable source for tooltip info.
But the JIRA ticket comes to our rescue!!

`Foletti Ambrogio added a comment - 28/Sep/12 1:56 PM 
Hey Damien,
We have to implement some changes:
no more tooltip on this layer`

Then: no tooltip is right, model is no more needed

loicgasser:
@daguer please proceed.

PR #2725, opened 2017-11-30
removed standav mako and model

AFoletti:
fixes https://github.com/geoadmin/mf-chsdi3/issues/2687


loicgasser:
ah but i shouldn't wait, :x

Issue #2726, opened 2017-11-30
Support abbrevation for addresses


davidoesch:
In Swisssearch, abbreviations like 
Seftigenstr.
Bundespl.
Av.
etc
Are not recognized anymore. ( I'm quite sure this worked in the past) 

Can you dig out the old behavior?

ltclm:
duplicate of https://github.com/geoadmin/mf-geoadmin3/issues/4036

PR #2734, opened 2017-12-01
Remove dot from location search

loicgasser:
I analyzed locations content, we should be able to remove the dots from the location search.
Fix for https://github.com/geoadmin/mf-chsdi3/issues/2726 and https://github.com/geoadmin/mf-geoadmin3/issues/4036
[Test Link](https://mf-chsdi3.int.bgdi.ch/search_locations_special_charaters/shorten/773d811bd8)

### DIFF
--- a/chsdi/lib/helpers.py
+++ b/chsdi/lib/helpers.py
@@ -127,6 +127,13 @@ def format_search_text(input_str):
     )
 
 
+def format_locations_search_text(input_str):
+    if input_str is None:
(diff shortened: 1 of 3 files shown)


loicgasser:
ping @davidoesch good for you?

davidoesch:
Tested with pl. Av. And str. Hope we don't break sth else with it 

+1 go

gjn:
Merci!

Issue #2737, opened 2017-12-01
Date formatting error in htmlpopup for ch.bav.sachplan-infrastruktur-schiene_ausgangslage


loicgasser:
layer id: ch.bav.sachplan-infrastruktur-schiene_ausgangslage
feature id: 71-W-06-001

try a html popup request these values.
result:
relevant traceback
```
  File "/var/lib/jenkins/workspace/PR_chsdi3/chsdi/templates/htmlpopup/base.mako", line 58, in render_body
    ${self.table_body(c, lang)}
  File "/var/lib/jenkins/workspace/PR_chsdi3/chsdi/templates/htmlpopup/sis_angaben.mako", line 4, in render_table_body
    <%
  File "/usr/lib/python2.7/_strptime.py", line 325, in _strptime
    (data_string, format))
ValueError: time data '' does not match format '%Y%m%d'

```

Do you guys now what's going on?

It says it doesn't exist but we have the models.
http://mf-chsdi3.dev.bgdi.ch/rest/services/ech/MapServer/ch.bav.sachplan-infrastruktur-schiene_ausgangslage/71-W-06-001/htmlPopup

PR #2738, opened 2017-12-01
date attribute can have null values

ltclm:
fixes https://github.com/geoadmin/mf-chsdi3/issues/2737
new data delivery can have NULL values in the attribute ``valid_from``

### DIFF
--- a/chsdi/templates/htmlpopup/sis_angaben.mako
+++ b/chsdi/templates/htmlpopup/sis_angaben.mako
@@ -9,7 +9,10 @@
     description = 'description_%s' % lang
     facility_kind = 'fackind_text_%s' % lang
     facility_status = 'facstatus_text_%s' % lang
-    datefrom = datetime.datetime.strptime(c['attributes']['valid_from'].strip(), "%Y%m%d").strftime("%d.%m.%Y")
+    try:
+        datefrom = datetime.datetime.strptime(c['attributes']['valid_from'].strip(), "%Y%m%d").strftime("%d.%m.%Y")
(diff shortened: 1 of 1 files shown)


chsdi/templates/htmlpopup/sis_angaben.mako
```diff
@@ -9,7 +9,10 @@
     description = 'description_%s' % lang
     facility_kind = 'fackind_text_%s' % lang
     facility_status = 'facstatus_text_%s' % lang
-    datefrom = datetime.datetime.strptime(c['attributes']['valid_from'].strip(), "%Y%m%d").strftime("%d.%m.%Y")
+    try:
+        datefrom = datetime.datetime.strptime(c['attributes']['valid_from'].strip(), "%Y%m%d").strftime("%d.%m.%Y")
+    except:
+        datefrom = '-'
```
loicgasser: Thanks but better test if this attr is None or not. Would be a
Slightly cleaner

ltclm: but less robust. db input for this function is a text column which can contain anything
```python
datefrom = datetime.datetime.strptime('anything_but_knickerbocker'.strip(), "%Y%m%d").strftime("%d.%m.%Y")
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/_strptime.py", line 325, in _strptime
    (data_string, format))
ValueError: time data 'anything_but_knickerbocker' does not match format '%Y%m%d'

```

PR #2739, opened 2017-12-04
fix model ch.bafu.bundesinventare-moorlandschaften

nadine-piveteau:


nadine-piveteau:
there is still an error, but as far as I can see it is due to the layer `ch.bav.sachplan-infrastruktur-schiene_ausgangslage`

PR #2742, opened 2017-12-06
BFS model update

loicgasser:


gjn:
As said, I would like @davidoesch and @cedricmoullet to validate the new data, inlcuding the new search.

Can you provide a mf-geoadmin3 testlink for them?

davidoesch:
Happy to Test... Need a link

loicgasser:
no change in location search. featuresearch only and there is no change either, we just added the `searchLang` parameter for BFS. it has been validated by BFS.

Issue #2750, opened 2017-12-13
update geocat links


ltclm:
after the migration to geocat 3.4 this week  we have to update the links in layer info toolbox.
s

```
Im Infofenster von map.geo.admin.ch, als Link zu geocat.ch wird immer noch der sehr alte Dienst gebraucht von geocat.ch 2. Inzwischen gab es geocat.ch 3.0, und jetzt geocat.ch 3.4.

Bei der Ablösung von geocat 2 zu geocat 3.0 hatten wir kommuniziert, dass dieser alten Diesnt noch eine Weile funktionnieren wird als Übergangslösung, dass der neue Dienst von geocat.ch gebraucht werden sollte. 
Ich dachte, dass alle Partner dieses Umschalten gemacht hatten, jetzt wurde aber festgestellt, dass es bei map.geo.admin.ch nicht der Fall ist.

Ich habe heute morgen C2C gefragt, ob man der alte Deist wieder aufschalten kann. Antwort habe ich noch keine.
Besser wäre aber, dass man bei der Generierung dieser URL in map.geo.admin.ch den Code wechselt. Die URL wird ja aus der UUID automatisch bei euch generiert.

Hier ein Beispiel:
Der alte Dienst: 
http://www.geocat.ch/geonetwork/srv/fra/metadata.show?uuid=9e574967-a470-4f33-b635-b7badb0b30a3&currTab=simple

der neue Dienst:
https://www.geocat.ch/geonetwork/srv/fre/md.viewer#/full_view/9e574967-a470-4f33-b635-b7badb0b30a3
oder sonst dieser wäre auch möglich
https://www.geocat.ch/geonetwork/srv/fre/catalog.search#/metadata/9e574967-a470-4f33-b635-b7badb0b30a3
```
please ask andre schneider for the new language codes

ltclm:
the new language codes are:
``ger/fre/eng/ita``

PR #2752, opened 2017-12-14
Update links to geocat

Sgachet:


Sgachet:
To fix #2750

loicgasser:
Thanks, we need to look into the geocat to bod code  as well.

Issue #2637, opened 2017-08-30
Generate dynamically tilematrix set defs for 4326 and 3857


loicgasser:
At the moment we don't honour min and max resolutions for 4326 and 3857.
We have only one tilemartixset definition.
We have to generate these templates dynamically, the same way we do it for 2056 and 21781.

Of course we need to adapt the zoom levels, but you can use getClosestZoom with a metric resolution to get this info and generate the appropriate tilematrixsets.

Issue #2636, opened 2017-08-30
Add tilematrixset identifier tests


loicgasser:
Still has not been done

```xml
<TileMatrixSetLink>
<TileMatrixSet>4326</TileMatrixSet>
</TileMatrixSetLink>
```

at the layer level should be found in

```xml
<TileMatrixSet>
<ows:Identifier>4326</ows:Identifier>
<ows:SupportedCRS>EPSG:4326</ows:SupportedCRS>
....
```

Make sure that for every `TileMatrixSet` there is an `ows:Identifier`.

https://mf-chsdi3.int.bgdi.ch/1.0.0/WMTSCapabilities.EPSG.4326.xml

oterral:
look into the possibility of generating one getcap that contains all the SRIDs.
just look at it, and we'll discuss it

PR #2759, opened 2017-12-15
Dynamic generation of tilematrixset in GetCapabilities

nadine-piveteau:
To solve issue #2637 and #2636 
Tilesetmatrix are generated automatically for all SRIDs. 
Unittest has been accordingly modified.

[Testlink ](https://mf-chsdi3.dev.bgdi.ch/ltpin_getCap/rest/services/api/1.0.0/WMTSCapabilities.xml?epsg=4326) for EPSG:4236


oterral:
Funny just saw we have a strange  ServiceMetadatURL at the end of the GetCap:
```
<ServiceMetadataURL xlink:href="http://www.opengis.uab.es/SITiled/world/1.0.0/WMTSCapabilities.xml"/>
```
It's not related to this PR, it's already on producton.

oterral:
@nadine-piveteau trst link doesn't work anymore, deploy a branch please. `make deploybranch`

chsdi/templates/wmtscapabilities/wmtscapabilities.mako
```diff
@@ -7,8 +7,10 @@
   onlineressource = pageargs['onlineressource']
   tilematrixset = pageargs['tilematrixset']
   tmsDefs = pageargs['tilematrixsetDefs']
+  zoomLevels = pageargs['zoomlevels']
   epsg = tilematrixset
   TileMatrixSet_epsg = "TileMatrixSet_%s.mako" % epsg
+  topLeftCorners = {'2056':[2420000.0, 1350000.0], '21781':[420000.0, 350000.0], '3857':[-20037508.3428, 20037508.3428], '4326':[90, -180]}
```
loicgasser: You should pass in the correct topLeftCorner directly from the view.

chsdi/views/wmtscapabilities.py
```diff
@@ -26,6 +26,15 @@ def getDefaultTileMatrixSet(tileMatrixSet):
     return tilematrixSet
```
loicgasser: You could return the topLeftCorner directly from here. gargrid object has a `MAXX`, ` MAXY` attributes.

nadine-piveteau:
Here is the [testlink](http://mf-chsdi3.dev.bgdi.ch/ltpin_getCap/rest/services/api/1.0.0/WMTSCapabilities.xml?epsg=4326).

loicgasser:
`4326_11`
We have
ch.swisstopo.fixpunkte-hfp1
ch.swisstopo.fixpunkte-hfp2
ch.swisstopo.fixpunkte-lfp1
ch.swisstopo.fixpunkte-lfp2

@ltclm seems weird to me that we can't zoom more on these layers.
In geoadmin these layers are WMS. To me this incorrect but I may be wrong.

I tested and it seems to work well thanks.



`

chsdi/views/wmtscapabilities.py
nadine-piveteau: Ok nice! Shouldn't it be `MAXY` and `MINX`?

loicgasser: yes :D

ltclm:
good find @loicgasser 
It is correct to have wms  layers in map.geo.admin.ch because they're updated daily.
I think we should remove them from the wmts getcap, i think it is better to have no data than outdated/wrong data. 
@mariokeusen @gjn what do you think?

loicgasser:
Yes if we don t manage cache update at WMTS level there s no point. +1

loicgasser:
You won't merge this for today's deploy even if it already looks very good.

loicgasser:
```
ch.swisstopo.fixpunkte-hfp1 | 193
ch.swisstopo.fixpunkte-hfp2 | 122
ch.swisstopo.fixpunkte-lfp1 | 118
ch.swisstopo.fixpunkte-lfp2 | 100
```

These are the number of WMTS requests for these layer for the past 7 days.

All of which have been generated either in the tests either by me testing those out, so I removed these from the GetCap.

loicgasser:
@nadine-piveteau is gona have a look at the suspisous ones.

ltclm:
thanks @loicgasser  for removing them

gjn:
If in wmts, they should be managed accordingly. Meaning when data is updated, caches (S3, CF) need to be invalidated.

I think in the future we will have use cases of often updated WMTS layers, automatically updated WMTS layers, etc. That's why the ToD maintenance services are so central. They need to be stable and they need to scale well to enable us in this direction.

For now, I would remove those layers from the WMTS.

loicgasser:
LGTM, please remove the templates we don't use anymore.

PR #2764, opened 2017-12-19
ch.sem.sachplan-asyl_kraft will not be empty after the deploy

Sgachet:
but ch.sem.sachplan-asyl_anhoerung remains empty


loicgasser:
This is only for the tests. I ll merge after tomorrow s deploy.

PR #2775, opened 2018-01-03
Meteo Layers First integration

loicgasser:
[Test Link](https://mf-chsdi3.dev.bgdi.ch/meteo_temp/shorten/79702f7dc4)

**Before merge**

- revert bod_master target
- remove hack in chsdi/models/bod.py


scripts/createlegends.js
```diff
@@ -39,16 +39,18 @@ let getStylesByResolution = function(data, styleType) {
...
     let style = data[styleType][i];
-    if (style.minResolution) {
-      resolutionHash += '-min-' + style.minResolution.toString();
-    }
-    if (style.maxResolution) {
-      resolutionHash += '-max-' + style.maxResolution.toString();
-    }
-    if (!styles.hasOwnProperty(resolutionHash)) {
-      styles[resolutionHash] = [];
+    if (String(style.vectorOptions.src).indexOf('nodata') == -1) {
```
oterral: It's cleaner to use RegExp to compare string content and you can't be sure vectorOptions exists so do:
```
if (style.vectorOptions && /nodata/.test(style.vectorOptions.src))
```
Anyway it seems really hackish,  the function stores styles by resolutions, I don't think the url of the image is a good reason to not store a styles in this array

Can you ask meteo guys why they don't want the no data image in there legend. It seems better idea to add it to the legend.

scripts/createlegends.js
```diff
@@ -273,7 +275,9 @@ let createLegendRamp = function(data, labels, lang, cb) {
   let styles = getStylesByResolution(data, styleType);
   let keys = Object.keys(styles);
   // availale res
-  let legendKey = keys.indexOf('res-max-50') !== -1 ? 'res-max-50' : keys[0];
+  //let legendKey = keys.indexOf('res-max-50') !== -1 ? 'res-max-50' : keys[0];
+  let legendKey = keys.indexOf('res-max-100') !== -1 ? 'res-max-100' : keys[0];
```
oterral: We need to know why this `if ` exists otherwise we risk to break other styles. @loicgasser ?

rebert: I verified all legends and it seems to work

scripts/createlegends.js
rebert: @oterral 
1. thanks for the RegExp
2. the hack is, that the all nodata images are called nodataxx.png
3. the script is a oneshot in the context of generating meteoschweiz legends - based on a spreadseet taking the order of the entries into account...

rebert:
Done:
    revert bod_master target
    remove hack in chsdi/models/bod.py

rebert:
@ltclm @procrastinatio @oterral after merging this, jenkins is going to be happy again

rebert:
@oterral thanks for your feedback: now I used optipng -o3 as chsdi uses for the other legends

procrastinatio:
LGTM.

PR #2784, opened 2018-01-16
added type to tooltip

rebert:
Added attribute type to tooltip of layer gemeinden: [try](https://mf-chsdi3.dev.bgdi.ch/ltrea_boarders3d/shorten/7829275793)

### DIFF
--- a/chsdi/models/vector/stopo.py
+++ b/chsdi/models/vector/stopo.py
@@ -491,9 +491,10 @@ class SwissboundariesGemeinde(Base, Vector):
     __label__ = 'gemname'
     id = Column('id', Integer, primary_key=True)
     gemname = Column('gemname', Unicode)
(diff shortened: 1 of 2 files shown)


chsdi/models/vector/stopo.py
```diff
@@ -494,6 +494,7 @@ class SwissboundariesGemeinde(Base, Vector):
     gemflaeche = Column('gemflaeche', Numeric)
     perimeter = Column('perimeter', Numeric)
     kanton = Column('kanton', Unicode)
+    objektart = Column('objektart', Numeric)
```
loicgasser: Please take the chance to replace Numeric, prefer float or integer. Only use it if really needed

rebert:
@loicgasser thanks for your feedback, which has been applied

PR #2794, opened 2018-01-23
Add new neophyten layers

daguer:


loicgasser:
Is that for tomorrow?

daguer:
No...

loicgasser:
@daguer this seems to be a duplicate of https://github.com/geoadmin/mf-chsdi3/pull/2796/files

Can you confirm and close if this is the case?

loicgasser:
ah no too many neophyten layers. confusing stuff.

Issue #2745, opened 2017-12-07
Unify GWR import


loicgasser:
2 tables at the moment

1. bfs.adr
2. bfs.gwr_addresses

Unify into one import 

primary key should be egid_edid

we don't need gde_gdename_${lang} as it seems there is always only one language anyway.

but we need stn_strname_${lang}

ltclm:
on-the-fly concatenation of egid_edid has been statified and indexed in db.

loicgasser:
Thks

ltclm:
will be fixed with:
changes in db: https://github.com/geoadmin/tool-wms/pull/120
automated import: https://github.com/geoadmin/bgdi-scripts/pull/259

ltclm:
see here for a short overview of the attributes in v01 and v02 tables and what has been done
https://docs.google.com/spreadsheets/d/1cVuuk2M3U2LMXyDMXsTuGv94ek78FM9L_ipqdQr7pHo/edit?usp=sharing

ltclm:
will be applied 2018-01-22 after the next regular auto-update

gjn:
Can we close this? @loicgasser @ltclm

ltclm:
yes, everything is done.

PR #2795, opened 2018-01-23
Unify GWR

loicgasser:
Fix for https://github.com/geoadmin/mf-chsdi3/issues/2745

Now the table has all the indices we can/need to point to one table.


loicgasser:
@ltclm can you confirm that it is correct?

PR #2801, opened 2018-01-24
Test top left

loicgasser:
Avoid the issue we had today.

### DIFF
--- a/chsdi/tests/integration/test_wmtscapabilities.py
+++ b/chsdi/tests/integration/test_wmtscapabilities.py
@@ -76,14 +76,32 @@ class TestWmtsCapabilitiesView(TestsBase):
                     if s not in used_matrices:
                         used_matrices.append(s)
             # Get all TileMatrixSets which are defined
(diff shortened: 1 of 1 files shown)


chsdi/tests/integration/test_wmtscapabilities.py
```diff
@@ -76,14 +76,32 @@ def test_tilematrixsets_are_defined(self):
...
             # Get all TileMatrixSets which are defined
-            tilematrixsets = root.findall('.//{http://www.opengis.net/wmts/1.0}TileMatrixSet/{http://www.opengis.net/ows/1.1}Identifier')
-
-        for tilematrixset in tilematrixsets:
-            t = tilematrixset.text
-            if t not in defined_matrices:
-                defined_matrices.append(t)
-
-        self.assertTrue(set(used_matrices).issubset(defined_matrices))
+            tile_matrixs = root.findall('.//{http://www.opengis.net/wmts/1.0}TileMatrixSet/{http://www.opengis.net/wmts/1.0}TileMatrix')
```
procrastinatio: plural of _matrix_ is _matrixes_ or _matrices_ (which is the correct latin flexion) 🤣

loicgasser: indeed

PR #2803, opened 2018-01-25
[WIP] Route method control

loicgasser:
As discussed earlier, we should be more accurate and restrictive about the HTTP methods we accept for certain routes.

- Restrict HTTP methods
- Update pyramid
- Remove checker_dev

https://docs.pylonsproject.org/projects/pyramid/en/latest/changes.html

@procrastinatio I count on you for a full review!

[Test Link](https://mf-chsdi3.dev.bgdi.ch/route_method_control)

### DIFF
--- a/chsdi/__init__.py
+++ b/chsdi/__init__.py
@@ -22,17 +22,23 @@ def db(request):
     return session
 
 
-def add_cors_route(config, pattern, service, headers=None):
+def add_cors_route(config, pattern, service, headers=None, methods=[]):
+    allowed_methods = ['OPTIONS', 'HEAD']
+
     def view(request):  # pragma: no cover
         response = request.response
         response.cache_control.no_cache = True
         response.cache_control.max_age = 0
+        response.headers['Access-Control-Allow-Origin'] = '*'
+        response.headers['Access-Control-Allow-Methods'] = ','.join(
+            allowed_methods + methods)
         if headers is not None:
-            response.headers = headers
+            for k, v in headers.iteritems():
+                response.headers[k] = v
         return response
 
     name = service + '_options'
-    config.add_route(name, pattern, request_method=('OPTIONS', 'HEAD'))
+    config.add_route(name, pattern, request_method=tuple(allowed_methods))
     config.add_view(view, route_name=name)
 
 
@@ -52,7 +58,10 @@ def main(global_config, **settings):
     # renderers
     config.add_mako_renderer('.html')
     config.add_mako_renderer('.js')
-    config.add_renderer('jsonp', JSONP(param_name='callback', indent=None, separators=(',', ':')))
+    config.add_renderer('jsonp',
+                        JSONP(param_name='callback',
+                              indent=None,
+                              separators=(',', ':')))
     config.add_renderer('geojson', GeoJSON(jsonp_param_name='callback'))
     config.add_renderer('esrijson', EsriJSON(jsonp_param_name='callback'))
     config.add_renderer('csv', CSVRenderer)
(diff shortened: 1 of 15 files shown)


chsdi/__init__.py
```diff
@@ -22,6 +22,27 @@ def cleanup(request):
...
 
+def add_cors_route(config, pattern, service, headers=None, methods=[]):
+    # The use of "GET" also implies that the view will respond to "HEAD".
+    allowed_methods = ['OPTIONS', 'HEAD']
+
+    def view(request):  # pragma: no cover
+        response = request.response
+        response.cache_control.no_cache = True
+        response.cache_control.max_age = 0
+        response.headers['Access-Control-Allow-Origin'] = '*'
```
loicgasser: I think Origin must be in header, but anyway it's overriden at apache level and we put * everywhere.

To handle CORS properly we should maybe remove some of the stuff here:
https://github.com/geoadmin/mf-chsdi3/blob/master/apache/wsgi.conf.in#L93

and add a subscriber on new response something like:
https://gist.github.com/kamalgill/b1f682dbdc6d6df4d052

chsdi/__init__.py
```diff
@@ -49,49 +73,276 @@ def main(global_config, **settings):
...
+    config.add_route('mapservice', '/rest/services/{map}/MapServer',
+                     request_method='GET')
+    add_cors_route(config,
+                   '/rest/services/{map}/MapServer',
+                   'mapservice',
+                   methods=['GET'])
+    # Returns the technical configuration of the layers in geoadmin
+    config.add_route('layersConfig',
+                     '/rest/services/{map}/MapServer/layersConfig',
+                     request_method='GET')
```
gjn: Sometimes `['GET']`, sometimes `'GET'`.

chsdi/__init__.py
```diff
@@ -49,49 +73,276 @@ def main(global_config, **settings):
...
+                   'checker',
+                   methods=['GET'])
+    # Download a KML created in geoadmin
+    config.add_route('downloadkml',
+                     '/downloadkml',
+                     request_method='POST')
+    add_cors_route(config,
+                   '/downloadkml',
+                   'downloadkml',
+                   methods=['POST'])
```
gjn: Why only POST?

oterral:
Does this PR forbid HEAD and OPTIONS requests on all services ?

loicgasser:
no you always have `OPTIONS` and `HEAD`

now at the moment you can make a `POST` request to perform an identify request for instance.
This PR is here to address that. But if people don't care, I don't either.

chsdi/views/downloadkml.py
```diff
@@ -47,9 +46,6 @@ def __init__(self, request):
     @view_config(route_name='downloadkml', renderer='jsonp')
     def downloadkml(self):
 
-        if self.request.method == 'OPTIONS':
-            return Response(status=200)
-
```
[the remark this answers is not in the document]
loicgasser: because of a misunderstanding on how pyramid works.

chsdi/__init__.py
loicgasser: it's the same, It'll fix to be consistent.

chsdi/__init__.py
loicgasser: because it works only with a body and a post request.
See:
https://github.com/geoadmin/mf-chsdi3/blob/42c8b7199a8f59fb10253d1727e8fcb38b82d19f/chsdi/views/downloadkml.py#L54

chsdi/__init__.py
loicgasser: actually `request_method` is always a string, except when we have several methods.

`methods` defines a custom route for `HEAD` and `OPTIONS` requests.

this is the way to do it in pyramid. As much as it is ugly....
c2c geoportal does something similar.

chsdi/views/downloadkml.py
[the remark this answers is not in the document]
loicgasser: mapgeoadmin validates CORS on this service.
It's a POST request that expects a body. But we should  have used a more generic approach for these since the beginning.

chsdi/__init__.py
```diff
@@ -49,49 +73,276 @@ def main(global_config, **settings):
...
+                   '/files',
+                   'files_collection',
+                   headers={
+                       'Access-Control-Allow-Credentials': 'true'
+                   },
+                   methods=['GET', 'POST', 'DELETE'])
+    # KML files handling, after the first one was created
+    config.add_route('files',
+                     '/files/{id}',
+                     request_method=('GET', 'POST', 'DELETE'))
```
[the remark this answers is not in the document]
loicgasser: lol, there is an endpoint for delete
https://github.com/geoadmin/mf-chsdi3/pull/2803/files#diff-5959b32d91ef71ebd584d90f1b47acbfR235

chsdi/__init__.py
loicgasser: access control allow methods should be removed from apache configuration as it is incorrect now.
It always sets all the methods.

gjn:
ok, let's go forward with this.

danduk82:
@boecklic and @procrastinatio , let's close some old tickets. Please first rebase on current `master`

boecklic:
Since PR is pretty old and parts of the code (eg. https://github.com/geoadmin/mf-chsdi3/pull/2803/files#diff-27266d2ca2d2f25f364b0ddf450144d7R25) have been introduced into master via different PRs at a later stage and in a improved manner (as far as I can tell) (https://github.com/geoadmin/mf-chsdi3/pull/2850) I'll go through the changes bit by bit and pack the relevant parts into a new PR based on current master.

danduk82:
ok, please create an issue to track it down, and reference this old PR and the new one. But please first finish the `mvt` stuff, as we will add restrictions also on `/gl-styles` as well.

PR #2811, opened 2018-02-05
added a mention of sr parameter to OeREB API doc

AFoletti:


AFoletti:
it would be good to have this one merged for tomorrow's deploy. Is just some doc 😉

loicgasser:
Yes it works if you supply lv95 coordinates but the response is always in LV03.

Remember what we did with these layers.
I don't know what priority this should have but, we would need to refactor some of our solution to also have LV95 coordinates in the response.

AFoletti:
no, the answer is not in LV03 but reads the XML as is stored in the database.
Since late November, we store (because of the new model) both LV03 and LV95 in the XML.
Thus, the service answers with the complete object (both geometries).

No changes are needed in the service. It works already as it should

loicgasser:
Ok I red lv03 coordinates that s why I was a bit surprised. Lgtm

PR #2774, opened 2017-12-20
Add Jenkinsfile

loicgasser:
We now use instance roles for aws resources and integrate a Jenkinsfile for the tests.

PR #2821, opened 2018-02-19
Htmlpopup test for CadastralWebMap failed

nadine-piveteau:
The test regarding the htmlpopup of cadastralwebmap failed because it couldn't find the ID. I also couldn't find it in the DB `os_realestate`. Therefore, I assumed that the ID in the REST call of `test_htmlpopup_cadastralwebmap` in `tests/integration/test_features_service.py` had to be updated.


loicgasser:
https://github.com/geoadmin/mf-chsdi3/pull/2774
I fixed in  the PR without having to hard code anything.
I am waiting for Andrea to apply instance roles.....

loicgasser:
Fixed in other PR

PR #2839, opened 2018-02-28
[WIP] Use instance role use

loicgasser:
Because of weird cross-account policies we get access denied server encryption.

Not detectable in chsdi tests nor in geoadmin until you reload the page....
Setting the acl at the object level didn't do the trick.
It's hard to get rid of these credentials....

see:
https://stackoverflow.com/questions/47654527/aws-s3-server-side-encryption-access-denied-error

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,9 +1,9 @@
 #!/usr/bin/env groovy
 
 node(label: "jenkins-slave") {
+
   try {
     stage("Checkout") {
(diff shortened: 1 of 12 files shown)


danduk82:
@nadine-piveteau ,
the new buckets for the service will be:
- public-dev-bgdi-ch
- public-int-bgdi-ch
- public-prod-bgdi-ch

PR #2840, opened 2018-03-01
remove cadastral timestamp workaround

ltclm:
will fix timestamp issues with ToD
https://github.com/geoadmin/service-proxywms/pull/209

### DIFF
--- a/chsdi/templates/wmtscapabilities/wmtscapabilities.mako
+++ b/chsdi/templates/wmtscapabilities/wmtscapabilities.mako
@@ -47,8 +47,6 @@
   ## Main loop
    % for layer in layers:
 <%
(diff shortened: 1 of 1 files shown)


ltclm:
we can revert the service-proxywms hack afterwards

loicgasser:
what about the hack just after can we get rid of it?

loicgasser:
All AWS related tests will fail because we couldn't use instance roles. Some cross account bucket policies issues with KML bucket...

loicgasser:
I am deploying with the hack. It's ready.
I'll remove after.
Otherwise we'll have to redeploy bod as well.

ltclm:
your're right, i have removed the zeitreihen workaround too.

ltclm:
thank you @loicgasser

PR #2850, opened 2018-03-14
Bucket Migration: Commit A 

nadine-piveteau:
This is the first step for the bucket migration where we add the newly created KML-files to both, the old bucket and the new bucket. 

[Testlink](https://mf-chsdi3.dev.bgdi.ch/ltpin_commitA/shorten/79501a5c90)

You can verify if it works by running the following commands: 
- check if added to new bucket: `aws s3 ls s3://public-dev-bgdi-ch/{file_id}`
- check if added to old bucket: `aws s3 ls s3://public.dev.bgdi.ch/{file_id}`


nadine-piveteau:
on its way

loicgasser:
We need a new variable PUBLIC_BUCKET_HOST as now bucket name is not the same as the host name anymore. Also let s squash all the commits so that we can easily revert in case we need to. Squash then rebase.

PR #2851, opened 2018-03-14
Bucket Migration: Commit B

nadine-piveteau:
PR for code after bucket migration, i.e. the KML-files are only writen to the new buckets.

[Testlink](https://mf-chsdi3.dev.bgdi.ch/ltpin_commitB/shorten/795098e343)
`aws s3 ls s3://public-dev-bgdi-ch/{file_id}` should return one entry
`aws s3 ls s3://public.dev.bgdi.ch/{file_id}` should return no entry


loicgasser:
https://github.com/geoadmin/mf-chsdi3/compare/ltpin_commitA...ltpin_commitB?expand=1

for comparison

loicgasser:
please rebase and squash as well

PR #2868, opened 2018-03-29
fix swissimage hist translation

ltclm:


loicgasser:
commit of the year

rebert:
the most colorful / colourful

PR #2891, opened 2018-04-30
by default use pypi.org

danduk82:
as with geoadmin, we should use pypi.org by default


danduk82:
seems to be working also with the CI

PR #2917, opened 2018-06-05
News layers bakom.notruf

Sgachet:
[Test link](https://mf-chsdi3.int.bgdi.ch/gas_bakom_notruf/shorten/7b0b4991ae)


alcapat:
Do not forget to translate all attributes of query tool (layername.attributename)

PR #2927, opened 2018-06-15
change bucket for grid based data

danduk82:
migrate the bucket for grid based data from the vectortiles bucket to a new dedicated one


danduk82:
the data copy is ongoing, it might take some time

rebert:
@danduk82 has been rebased

PR #2953, opened 2018-07-13
Received an alert that pillow < 3.3.2 had a moderate security vulnera…

ltkum:
…bility. https://www.cvedetails.com/vulnerability-list/vendor_id-10210/product_id-27460/Python-Pillow.html for references. Thought it would be a good idea to update that before we have some problems.

We should verify that builds are still valid after the update.


rebert:
@ltkum thanks! seems to work

PR #2955, opened 2018-07-16
updated translations

AFoletti:
Some updated geology translations for next deploy


rebert:
@AFoletti will be retranslated anyway

PR #2958, opened 2018-07-17
added atom to doc

rebert:


davidoesch:
Don't forget: Christina has to as well document it on www.geo.admin.ch under services... As we did for Wms

PR #2961, opened 2018-07-20
[WIP] api resource are elsewhere

procrastinatio:


gjn:
Wow, you are fast! 

- loader is 30KB gzipped, that's ok.
- true, but see above
- yes, but newer versions are available via parameter
- not too terribel, but because it's versioned, I think it's ok.
- yes, but we have others. we could create yet another bucket?

Few other points:
- Can you also make sure that `public, max-age=86400` header is set on the versioned resources?

(I assume and it looks liek all are gzipped).

I would be great if we could deploy this this week. What do you think?

gjn:
I'll close this as we use the other approache discussed previously.

procrastinatio:
Superseeded by https://github.com/geoadmin/mf-chsdi3/pull/2963

PR #2963, opened 2018-07-25
Adding vip option for api loader

gjn:
This add a new parameter to the loader.js API loader: `vip`

If `vip=true`, then the reources will be loaded from cloudfront backed URL (`public-cdn.*`) and it will load a special javascript file which will use the VIP wmts service range.

Note: the CF distributions and the vip files are not yet ready.


gjn:
@procrastinatio Please review. I think we can deploy this PR this afternoon (probably off of a release tag).

procrastinatio:
Demo: https://mf-chsdi3.int.bgdi.ch/gjn_vip_api_loader/loader.js?version=4.4.2&vip=true

procrastinatio:
Too much undocumented black magic. But looks OK.

gjn:
Ok, I will deploy this now.

PR #2972, opened 2018-08-30
Reduce deploy size

procrastinatio:
Only keep the current branch to save space on dev, int


danduk82:
will be done by separating static content

PR #2978, opened 2018-09-13
support linefeeds in ilike statements

ltclm:
you can test the new ilike behaviour here:
https://mf-chsdi3.dev.bgdi.ch/gas_bakom_notruf_fix/shorten/7cfd376a13

### DIFF
--- a/chsdi/lib/helpers.py
+++ b/chsdi/lib/helpers.py
@@ -177,7 +177,7 @@ def format_query(model, value):
     '''
     def escapeSQL(value):
         if u'ilike' in value:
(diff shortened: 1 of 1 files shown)


AFoletti:
Functionally it works very good and solves the problem 100%
I'm however leaving to the technicians to merge it because I did not test thoroughly for side-effects

ltclm:
integration tests are all passing... i don't think that we break something when we merge this.
Since this feature still is BETA and mostly unknown.

AFoletti:
Then let's go

PR #2991, opened 2018-10-03
Fix tooltip display too much info

alcapat:


alcapat:
Rq: Do not forget to deploy uvek_master on prod during next deploy, the tooltip use a 2 new views.

alcapat:
@ltclm : Perfect, thanks

PR #3001, opened 2018-10-23
new legend 4 schutzgebiete

rebert:


rebert:
will recreate them bigger

PR #3011, opened 2018-11-05
update legends sicherheitszonenplan

rebert:


chsdi/templates/htmlpopup/sicherheitszoneneplan.mako
```diff
@@ -16,10 +16,10 @@
 <%def name="extended_info(c, lang)"> 
 
 <%
-    import datetime
+    import dateutil.parser as date_parser
     lang = lang if lang in ('fr','it') else 'de'
     format = 'formate_%s' % lang
-    date = datetime.datetime.strptime(c['attributes']['approval_date'].strip(), "%d-%m-%Y").strftime("%d.%m.%Y")
+    date = date_parser.parse(c['attributes']['approval_date']).strftime('%d.%m.%Y')
```
procrastinatio: What if error occur?

rebert: @procrastinatio when error then error. but there is no field without a valid date in the table. Any proposition?

Issue #2720, opened 2017-11-24
"Feature ID <blabla>: is too large" is not a good solution


AFoletti:
When an object has too large a geometry, the identify service currently answers this:
`{
  "status": "error",
  "code": 509,
  "detail": "Feature ID 814241: is too large"
}`

This is (maybe???? But I doubt it) not bad for services but absolutely bad for map.geo.admin.ch, where the user gets nothing. Apart the impression that the layer is broken.
See [here for example](https://s.geo.admin.ch/771999112c)

**My suggestion:**
when the condition for this error is found true, the identify correctly returns the object propreties EXCEPT the geometry.

Assigned @gjn and @davidoesch for evaluation 😄

gjn:
The reason for this is service protection. large geometries result in timeouts and can lead to congestions that then affect other services. As you said, this is almost a necessity from the service point of view.

You can get what you want from the service perspectiv: use the returnGeometry=false parameter as indicated in the documentation and you'll get all other results.

From the application point of view, I agree that it's bad. User should have feedback. Not sure what the best approach would be - in this case, I think that it would make sense to have highlight geometries. We already have a couple of layers with highlight geometries - which are used for highlighting on front-end (and are usually a simplified version of the real geometry).

Could this be possible for this layer?

AFoletti:
Possible is everything. My proposal was a more general one.
Geometry too big=the service answers with attributes but no geometry. We will avoid the 509 error (this giving service-users something useful about the queryed object) while keeping the service protection capability.

gjn:
Might be a possibility - but then again it will confuse the user of the API that he suddenly gets a result but with no geometry - and he doesn't know why.

hen-egg.

We should have geometries in our services that are transferable in http services...

AFoletti:
As I said, it is open to you for evaluation. Concocting a geometry suited for highlight is not a problem if we want to go that way

oterral:
highlight geometry seems a good approach.

procrastinatio:
it makes absolutely no sense to highlight the clicked feature in this example: do you really want to highlight the whole quaternary layer?

AFoletti:
No @procrastinatio  I don't want to highlight a whole quaternary layer. But if you click on it I **want** to become a tooltip, not a hidden error thrown by the service.
Because in the actual state the user is getting nothing --> the layer is for the user broken!!

procrastinatio:
So solved. No highlight, but tooltip.

PR #3019, opened 2018-11-08
Return attributes even if geom is too big (fixe #2720)

procrastinatio:
If geometry is too big (WKB >1e6):
- Attributes are returned
- Only bbox is calculated
- geometry is null

[Demo (click on marker)](https://map.geo.admin.ch/?api_url=%2F%2Fmf-chsdi3.dev.bgdi.ch%2Fmom_too_big_geom&lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.geologie-geologischer_atlas&layers_opacity=0.75&E=2556761&N=1162056&zoom=9&crosshair=marker) _Moraine rhodanienne_ is the big geometry

Use http://mf-chsdi3.dev.bgdi.ch/mom_too_big_geom/

For instance:

https://mf-chsdi3.dev.bgdi.ch/mom_too_big_geom/rest/services/all/MapServer/identify?geometry=2556868,1162263&geometryFormat=geojson&geometryType=esriGeometryPoint&imageDisplay=900,981,96&lang=fr&layers=all:ch.swisstopo.geologie-geologischer_atlas&limit=10&mapExtent=2555861,1161075,2557661,1163037&returnGeometry=true&sr=2056&tolerance=10

From the service point of view this is not a good solution

See https://github.com/geoadmin/mf-chsdi3/issues/2720

### DIFF
--- a/chsdi/models/vector/__init__.py
+++ b/chsdi/models/vector/__init__.py
@@ -79,16 +79,16 @@ class Vector(object):
                     id = val
                 elif (isinstance(col.type, GeometryChsdi) and
                       col.name == self.geometry_column_to_return().name):
-                    if hasattr(self, '_shape'):
+                    if hasattr(self, '_shape') and \
+                            len(self._shape) < MAX_FEATURE_GEOMETRY_SIZE:
                         geom = self._shape
-                    elif val is not None:
+                    elif val is not None and \
+                            len(val.data) < MAX_FEATURE_GEOMETRY_SIZE:
                         geom = to_shape(val)
                     try:
                         bbox = geom.bounds
                     except:
                         pass
-                    if len(val.data) > MAX_FEATURE_GEOMETRY_SIZE:
-                        geom = None
                 elif (not col.foreign_keys and
                       not isinstance(col.type, GeometryChsdi)):
                     properties[p.key] = val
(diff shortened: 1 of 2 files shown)


AFoletti:
Great! Many thanks!

chsdi/models/vector/__init__.py
```diff
@@ -79,19 +82,17 @@ def __read__(self):
                     if hasattr(self, '_shape'):
                         geom = self._shape
                     elif val is not None:
-                        if len(val.data) > 1000000:
-                            raise HTTPBandwidthLimited(
-                                'Feature ID %s: is too large' % self.id)
                         geom = to_shape(val)
```
ltkum: Wouldn't it be better to put it after the length test ? Seems to me like it would transform the data into a shape, then verify if the data isn't too big to put into a shape. As to_shape(val) is the expected return value, it would be definitiely be better to do a 
```
if len(val.data) > MAX_FEATURE_GEOMETRY_SIZE:
  geom = None
else: 
  geom = to_shape(val)
```

chsdi/models/vector/__init__.py
```diff
@@ -65,6 +67,7 @@ class Vector(object):
     def __read__(self):
         id = None
         geom = None
+        bbox = None
```
ltkum: yeah, it is a good idea to have all attributes declaration at the same place.

procrastinatio:
@ltkum Did also add a test.

ltkum:
seems good to me.

Issue #2879, opened 2018-04-17
offset param not documented


oterral:
@davidoesch @gjn @loicgasser do you remember why the offset param (paging stuff) of identify service is not documented?

davidoesch:
Assumption:  back in the days when gwr/addresses request results were limited to 50 ?
-> you can document it.

loicgasser:
Yes we tried to avoid data scraping.

procrastinatio:
@danduk82  @ltkum Basically, where do we document _undocumented_ features?

davidoesch:
Yeah I'm interested in that one as well

loicgasser:
I think it was to avoid data scraping. You put the extent of switzerland and then you add more and more offset.

PR #3022, opened 2018-11-09
Documenting "undocumented" params

procrastinatio:
http://mf-chsdi3.dev.bgdi.ch/fix_2879/services/sdiservices.html#identify-features

See https://github.com/geoadmin/mf-chsdi3/issues/2879


ltkum:
seems good to me

PR #3039, opened 2018-11-21
Add new layers biomasse

Brunim28:


Brunim28:
[test link](https://mf-chsdi3.dev.bgdi.ch/brg_biomasse/shorten/7ec30b9e84)

Sgachet:
I'm wondering if the second attribute name (in the tooltip) could not be put on 2 lines rather than 3, what do you think @ltclm . But I do not know if it's feasible?

PR #3047, opened 2018-11-30
Limit search bbox area

procrastinatio:


danduk82:
@procrastinatio , could you please reslove the conflicts before @ltclm reviews it? Would be great to include in next deploy.

danduk82:
@ltclm, is this solved on sphinx level?

ltclm:
yes, the memory issues have been solved on sphinx level. index/dictionary types have been changed to legacy type crc.
memory usage is stable now  https://grafana.bgdi.ch/d/FU8IPh8iz/sphinx-prod?refresh=1m&orgId=5

danduk82:
after discussion, the problem is fixed on sphinx side. So I close this as `won't do`

PR #3059, opened 2018-12-14
Update sdiservices.rst

davidoesch:
remove the example... we even add it to the exmaple section and we will add in the cms as well.


chsdi/static/doc/source/services/sdiservices.rst
```diff
@@ -1087,7 +1087,7 @@ Example
 * A `Vector tile <https://vectortiles.geo.admin.ch/mbtiles/ch.swisstopo.swissnames3d/v004/7/67/44.pbf>`_
 * A `Style <https://vectortiles.geo.admin.ch/gl-styles/ch.swisstopo.leichte-basiskarte.vt/v004/style.json>`_
 * A `Tileset <https://vectortiles.geo.admin.ch/mbtiles/ch.swisstopo.swissnames3d/v004.json>`_
-* A `Mapbox-gl application using our light base map <https://vectortiles.geo.admin.ch/gl-styles/ch.swisstopo.leichte-basiskarte.vt/v004/?vector#7/47/8/0/6>`_
```
danduk82: @davidoesch , why not? I think it is good if people see also a working mapbox example. This is the vectortiles service, and using a style with mapbox is easier than with OL...

davidoesch: We do communicate via CMS ....

PR #3060, opened 2018-12-17
Davidoesch patch 1

danduk82:


chsdi/static/doc/source/services/sdiservices.rst
```diff
@@ -1039,6 +1039,55 @@ Example
...
+URL
+***
+
+- vector tiles: https://vectortiles.geo.admin.ch/mbtiles/
+- gl-styles: https://vectortiles.geo.admin.ch/gl-styles/
+
+Metadata Service
+****************
+
+Each tileset has a corresponding metatda `json` file that describes the available set of tiles.
```
oterral: missing 'a'

loicgasser:
We should give examples with om and mapboxgl.

davidoesch:
we will provide it in www.geo.admin.ch produkt page including isntructions  to how to use it in QGIS

davidoesch:
This One 

We will provide data links in CMS

davidoesch:
hi @danduk82 can't find it on todays deploy. since we would like to go live soon: when will this be available on api3.geo.admin.ch?

danduk82:
@davidoesch , when we decide for an offical go-live for the service vectortiles. Anyway after the service is stable on S3.

davidoesch:
well the idea is, as metioend last monday, to be ready to define the go live next monday, but for this the description needs to be online.

no worry if it is already online silent... we just loose time if we wait.

Issue #3071, opened 2019-01-14
create a new clean branch based on master and pick changes from `route_method_control`


boecklic:
The old PR #2803 was made against a fairly different state of master. A new PR should be created that contains only the still relevant changes.

procrastinatio:
Work is done, but we will never merge it because it will break things on the user side.

PR #3072, opened 2019-01-14
Route method control and CORS

boecklic:
This PR resolves issue #3071 and replaces the old PR #2803. PR #2803 has a lot of unresolved discussions. Currently the PR just copy-paste changes as is in the old PR, so when reviewing this, have a quick look at the discussions in PR #2803. 

Edit by  @procrastinatio 
:bomb: Please carefully test, it may break many things. Most of the content of this PR is however useless, as most headers are rewritten by apache.

Test link = http://mf-chsdi3.int.bgdi.ch/route_method_control_clean/

https://map.geo.admin.ch/?api_url=//mf-chsdi3.int.bgdi.ch/route_method_control_clean

### DIFF
--- a/chsdi/__init__.py
+++ b/chsdi/__init__.py
@@ -39,6 +39,7 @@ def add_cors_route(config, pattern, service, headers=None, methods=[]):
         return response
 
     name = service + '_options'
+    # TODO: should add_route not also consider 'methods' from args?
     config.add_route(name, pattern, request_method=allowed_methods)
     config.add_view(view, route_name=name)
 
@@ -59,7 +60,10 @@ def main(global_config, **settings):
     # renderers
     config.add_mako_renderer('.html')
     config.add_mako_renderer('.js')
-    config.add_renderer('jsonp', JSONP(param_name='callback', indent=None, separators=(',', ':')))
+    config.add_renderer('jsonp',
+                        JSONP(param_name='callback',
+                              indent=None,
+                              separators=(',', ':')))
     config.add_renderer('geojson', GeoJSON(jsonp_param_name='callback'))
     config.add_renderer('esrijson', EsriJSON(jsonp_param_name='callback'))
     config.add_renderer('csv', CSVRenderer)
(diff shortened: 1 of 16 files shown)


chsdi/__init__.py
```diff
@@ -70,61 +74,277 @@ def main(global_config, **settings):
...
-    config.add_route('qrcodegenerator', '/qrcodegenerator')
-    config.add_route('sitemap', '/sitemap')
-    config.add_route('luftbilder', '/luftbilder/viewer.html')
-    config.add_route('historicalmaps', '/historicalmaps/viewer.html')
-    config.add_route('checker', '/checker')
-    config.add_route('checker_dev', '/checker_dev')
-    config.add_route('downloadkml', '/downloadkml')
-
-    config.add_route('files_collection', '/files', request_method=('GET', 'POST', 'DELETE'))
+    add_cors_route(config, '/rest/services', 'topics', methods=['GET'])
```
boecklic: Adding the CORS routes imho adds a lot of duplicated code, maybe it's worth packing the two into a generic method that adds normal and cors routes and makes it easier to see which routes are defined how?

requirements.txt
```diff
@@ -32,7 +32,7 @@ pycodestyle==2.0.0
 pyflakes==1.2.3
 Pygments==2.1.3
 pyproj==1.9.5.1
-pyramid==1.8.3
+pyramid==1.9.1
```
boecklic: Is this upgrade save?

procrastinatio: Use my PR https://github.com/geoadmin/mf-chsdi3/pull/3086 (Pyramid 1.10)

tests/integration/test_admin.py
```diff
@@ -8,3 +8,7 @@ class TestAdminkml(TestsBase):
     def test_admin_kml(self):
         resp = self.testapp.get('/admin/kml', status=200)
         self.assertEqual(resp.content_type, 'text/html')
+
+    def test_admin_kml_post(self):
+        resp = self.testapp.post('/admin/kml', '', status=404)
+        self.assertEqual(resp.content_type, 'text/plain')
```
procrastinatio: Is it OK? Normally, all errors are returned as JSON

chsdi/__init__.py
procrastinatio: Hmm, https://github.com/geoadmin/mf-chsdi3/blob/d1ee3ff291d5b974ae6cb1ff7a8f0dbcd4240af8/chsdi/__init__.py#L25

pakb:
@procrastinatio would this be ready for Wednesday? Or do we add this for next milestone?

pakb:
what's the status on this?

procrastinatio:
This is a bit cleaner, but I really don't like this PR, because we have now routing and CORS at three different levels: `varnish`, `apache` and `pyramid`, and this is not easy to clean up as `route` are so entangled.

danduk82:
@davidoesch, we could add some kind of "deprecated warning" list on API docu before merging this, what do you think?

davidoesch:
users don't read the API docu.
So: use KIBANA and analyse which hosts /Domains / Applications are affected / we will break with this fix. 
Then we will inform them and give them a 3 month notice time to  react on changes

danduk82:
this will break a lot of clients...

ltshb:
@boecklic can we close this ?

procrastinatio:
I guess it has been superceded by https://github.com/geoadmin/mf-chsdi3/pull/3987 and may be deleted.
Any concern @ltshb @boecklic  @hansmannj @ltclm @rebert ?

boecklic:
Jup

PR #3073, opened 2019-01-15
default search is divided into exact search and wildcard search

ltclm:
@davidoesch @gjn @danduk82 
the standard sphinx query will be converted into a sphinx multi-query [1].
It's still one network request, but there are two queries executed on sphinx side.
The standard query has been splitted into a
* exact match query, returning first result of each origin only
* wildcard query (same configuration as before)

triggered by https://github.com/geoadmin/mf-geoadmin3/issues/4287

[Testlink](https://mf-chsdi3.dev.bgdi.ch/dev_ltclm_exact_search/shorten/7f7c3a7288)

The results of the exact match query (**one per origin**) will be prepended to the results of the wildcard query.
p.e search for 
```
Entle
Dom
Rotten
Ricken
```
The results of the exact match search will be on top of the list now.

**question:**
* how should we merge the results of the two searches?

a) search1 (exact) results on top of search2 (wildcard) results **current implementation in this pr**
p.e.
```
search1->gg25
search1->gazetteer
search2->gg25
search2->gazetteer
search2->gazetteer
search2->gazetteer
search2->address
search2->address
search2->address
search2->address
```
b) search1 (exact) results on top of each search2 (wildcard) results origin
p.e
```
search1->gg25
search2->gg25
search1->gazetteer
search2->gazetteer
search2->gazetteer
search2->gazetteer
search2->address
search2->address
search2->address
search2->address
```

We will also have to do some yandex load tests an monitor the impact on sphinx of this change.

**TODO:**

- [x] Take a decision how to merge the results (exact on top of wildcard) how many of the exact search results should be used (currently one per origin)
> "Take a decision how to merge the results (exact on top of wildcard) how many of the exact search results should be used (currently one per origin)"
-> All exact results are shown in top. User gets what he expects
- [x] load tests
- [x] customer feedback (chmobil, others?)

[1] http://sphinxsearch.com/docs/current/multi-queries.html

### DIFF
--- a/chsdi/views/search.py
+++ b/chsdi/views/search.py
@@ -131,14 +131,14 @@ class Search(SearchValidation):
 
         if len(searchList) != 0:
             try:
-                # wildcard search
+                # standard wildcard search
                 self.sphinx.AddQuery(searchTextFinal, index='swisssearch')
-                # exact search
-                self.sphinx.SetRankingMode(sphinxapi.SPH_RANK_PROXIMITY_BM25)
-                self.sphinx.SetGroupBy('origin', sphinxapi.SPH_GROUPBY_ATTR, 'rank ASC, @weight DESC, num ASC')
+                # exact search, first 10 results
+                self.sphinx.SetLimits(0, 10)
                 searchText = '@detail ^%s' % ' '.join(self.searchText)
                 self.sphinx.AddQuery(searchText, index='swisssearch')
-                self.sphinx.ResetGroupBy()
+                # reset settings
+                self.sphinx.SetLimits(0, limit)
 
                 temp = self.sphinx.RunQueries()
 
--- a/chsdi/views/search.py
+++ b/chsdi/views/search.py
@@ -131,10 +131,32 @@ class Search(SearchValidation):
 
         if len(searchList) != 0:
             try:
-                temp = self.sphinx.Query(searchTextFinal, index='swisssearch')
+                # wildcard search
+                self.sphinx.AddQuery(searchTextFinal, index='swisssearch')
+                # exact search
+                self.sphinx.SetRankingMode(sphinxapi.SPH_RANK_PROXIMITY_BM25)
+                self.sphinx.SetGroupBy('origin', sphinxapi.SPH_GROUPBY_ATTR, 'rank ASC, @weight DESC, num ASC')
+                searchText = '@detail ^%s' % ' '.join(self.searchText)
+                self.sphinx.AddQuery(searchText, index='swisssearch')
+                self.sphinx.ResetGroupBy()
+
+                temp = self.sphinx.RunQueries()
+
             except IOError:  # pragma: no cover
                 raise exc.HTTPGatewayTimeout()
-            temp = temp['matches'] if temp is not None else temp
+
+            temp_merged = temp[1]['matches'] + temp[0]['matches'] if len(temp) == 2 else temp['matches']
+
+            # remove duplicate results, exact search results have priority over wildcard search results
+            temp = []
+            seen = []
+            for d in temp_merged:
+                if d['id'] not in seen:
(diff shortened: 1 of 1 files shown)


davidoesch:
Needs to be approved by SCHWEIZ Mobil as well

ltclm:
First @gjn @davidoesch have to decide how the results are merged.
Then, i will modify the code and do some load tests.
Then we can send the testlink.

davidoesch:
"Take a decision how to merge the results (exact on top of wildcard) how many of the exact search results should be used (currently one per origin)"
-> All exact results are shown in top. User gets what he expects

ltclm:
ok, so for the merge it will be option a).

ltclm:
@danduk82 @davidoesch 
i did some load and performance tests with yandex:

the new multi-query requests are two times slower than the old requests (85ms -> 160ms)
the capacity stays the same, first 500 errors occur with 20 requests/s

see here https://docs.google.com/document/d/1qO3Si9xWCKIaca0pvhoZw7IJZGVPapbCG8wDkkoC73o/edit?usp=sharing

davidoesch:
ouch... performance goes a long way.
if we reduce the exact results to max of 10 ?

ltclm:
the exact search query already returns a maximum of one object per rank/origin, thus 7 objects (zipcode,gg25,district,kantone,gazetteer,address,parcel) 
``self.sphinx.SetGroupBy('origin', sphinxapi.SPH_GROUPBY_ATTR, '@group desc')``.
We have to scan the index twice instead of once, that explains pretty much the loss in performance i guess.

ltclm:
so the performances with a longer test benchmark period are quite equal, see results of test ``5 rps / 300s`` in the spreadsheet. Also the 2nd run in the test ``5 rps / 60s`` more or less the same performance.

davidoesch:
Will do a 1:1 with u tomorrow

danduk82:
@davidoesch can you check with CHmobil?

ltclm:
> @ltclm , where do you take the numbers `85ms -> 160ms`?
> from what I see in your test results, median times and 90percentile are slower, ok, but not that significantly. More 30% than 200% slower.

from 1st run from 5 rps / 60s test
```
new has  mean : 169.65505723905721
old has mean: 85.58455555555555
```

danduk82:
I don't trust mean values, median and percentiles are much more reliable.

danduk82:
and test 2 shows exactly the opposite behavior, do you know why? some sort of caching?

ltclm:
fyi i have changed the search behaviour to:
* exact search: first 10 results will be prepended to 
* wildcard search (unchanged)

this will be easier to explain and to tweak than a groupby

davidoesch:
LGTM we have a go from schweizmobil

gjn:
Almost 2 weeks live with this now (released on 22nd of February). And here's the reality:

image

Almost no detectable/significant change in average, 50th percentile and 90th percentile response times on the location search. I think that the 2 requests are executed in parallel and sphinx can handle it quite well.

I guess the change could lead to a lower ceiling (number of total concurrent requests) we can handle, so a spike will lead to sphinx to break earlier, but standard traffic, it handles equally well.

Thoughts? @danduk82 @ltclm

ltclm:
> @danduk82 @davidoesch
i did some load and performance tests with yandex:
the new multi-query requests are two times slower than the old requests (85ms -> 160ms)
the capacity stays the same, first 500 errors occur with 20 requests/s
see here https://docs.google.com/document/d/1qO3Si9xWCKIaca0pvhoZw7IJZGVPapbCG8wDkkoC73o/edit?usp=sharing

We did some load tests with yandex before putting this change to production. the max number of requests/sec stays the same : ``20 req/sec``.

davidoesch:
so far no negative feedback from customers

PR #3076, opened 2019-01-16
remove temporary transition mode locations_preview

ltclm:
this will remove the transition search type locations_preview, which is not working anymore and is mostly unused now (see kibana).

### DIFF
--- a/chsdi/lib/validation/search.py
+++ b/chsdi/lib/validation/search.py
@@ -14,7 +14,7 @@ class SearchValidation(MapNameValidation):
     def __init__(self, request):
         super(SearchValidation, self).__init__()
         self.availableLangs = request.registry.settings['available_languages'].split(' ')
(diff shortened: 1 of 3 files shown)


ltclm:
the referened sphinxsearch index swisssearch_preview is not existing anymore since dec 2017 
``ff5e1b0 loicgasser      Mon Nov 20 17:39:54 2017 +0100  Search lang in feature search not location search``
locations_preview searches have always been using fuzzy search since then.

ltclm:
@danduk82 its also planned for next week's deploy, just remove a legacy search mode which is not working anymore for a long time.

danduk82:
@ltclm , I might be wrong, but this is all, there's nothing to do on service-search side, is it?

ltclm:
yes, that's all. the corresponding index has been removed from service in dec 2017 already.

Issue #3042, opened 2018-11-22
Switch default ol3 version from 3.6.0 to 4.4.2 (LV95)


procrastinatio:
Switch to the latest available API version

danduk82:
@procrastinatio , is this related to some PR already?

procrastinatio:
No. Waiting decision...

procrastinatio:
PR #3078

danduk82:
won't do. Too many clients implied

PR #3078, opened 2019-01-16
Api default version to 4.4.2

procrastinatio:
This is a big change. **Please be sure to notify the users before merging**.

Examples (http://api3.geo.admin.ch/api/examples.html) are already using `4.4.2`

User https://mf-chsdi3.int.bgdi.ch/fix_3042/loader.js

Fixes https://github.com/geoadmin/mf-chsdi3/issues/3042


procrastinatio:
In the doc, it is explicitly stated older versions are not supported and may even not work. And by the way, nobody is suposed to use **LV03**  now.

Full list (kibana): https://s.geo.admin.ch/7fab3fdad0

davidoesch:
@procrastinatio : can you add in kibana as well the sites? so I can point the user to an URL

danduk82:
Won't fix, this would break too many client applications...

Issue #3085, opened 2019-01-22
Upgrade pyramid


procrastinatio:
The currently used version `pyramid==1.8.3` is not maintained anymore. Use `1.10`

PR #3086, opened 2019-01-22
Upgrading pyramid

procrastinatio:
to the latest available version
Fixing https://github.com/geoadmin/mf-chsdi3/issues/3085

### DIFF
--- a/tests/integration/test_identify_service.py
+++ b/tests/integration/test_identify_service.py
@@ -225,7 +225,7 @@ class TestIdentifyService(TestsBase):
         self.assertTrue(resp.json['detail'].startswith('Please provide the parameter imageDisplay in a comma separated list of 3 arguments (width,height,dpi)'))
 
         resp = self.testapp.get('/rest/services/ech/MapServer/identify', params=params, headers={'Accept': 'text/html'}, status=400)
-        self.assertEqual(resp.content_type, 'text/html')
+        self.assertEqual(resp.content_type, 'application/json')  # All errors are json
 
     def test_identify_valid_topic(self):
         params = {'geometry': '548945.5,147956,549402,148103.5',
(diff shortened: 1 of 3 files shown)


tests/integration/test_identify_service.py
```diff
@@ -225,7 +225,7 @@ def test_invalid_imageDisplay(self):
         self.assertTrue(resp.json['detail'].startswith('Please provide the parameter imageDisplay in a comma separated list of 3 arguments (width,height,dpi)'))
 
         resp = self.testapp.get('/rest/services/ech/MapServer/identify', params=params, headers={'Accept': 'text/html'}, status=400)
-        self.assertEqual(resp.content_type, 'text/html')
+        self.assertEqual(resp.content_type, 'application/json')  # All errors are json
```
boecklic: "All errors are json" is this defined globally/on Pyramid level? In the code I cannot find the place where content-type is changed to json when request header has `'accept':'text/html'`...

danduk82:
@procrastinatio and @boecklic  is this mergeable for next deploy?

procrastinatio:
```
 curl  -H "Accept: text/html" "http://localhost:9001/rest/services/api/MapServer/find?layer=ch.bfs.geude_wohnungs_register&searchText=1231641&searchField=egid&returnGeometry=false&contains=false"
<html>
 <head>
  <title>400 Bad Request</title>
 </head>
...
```

```
 curl  -H "Accept: */*" "http://localhost:9001/rest/services/api/MapServer/find?layer=ch.bfs.geude_wohnungs_register&searchText=1231641&searchField=egid&returnGeometry=false&contains=false"
{"status":"error","code":400,"detail":"No Vector Table was found for ch.bfs.geude_wohnungs_register for searchField egid"}
```

tests/integration/test_identify_service.py
procrastinatio: No, the render was set in the decorator of the view `@exception_view_config(context=HTTPException)`. I had to remove it, because only one decorator may be set per view. So returning a string and setting the correct mime-type should be enough.

procrastinatio:
Rebased

PR #3102, opened 2019-02-07
Missing translations

procrastinatio:
Cannot deploy. Why?


pakb:
You stole my translations :-)

PR #3108, opened 2019-02-13
Add models and mako for Geomol layers

Sgachet:
[Test link](https://mf-chsdi3.int.bgdi.ch/gas_geomol/shorten/80bc213543)


Sgachet:
@ltclm I need a deep review here too, thanks in advance ;-)

ltclm:
@AFoletti, @davidoesch  i think we should not configure tooltips anymore on layers like these, there's a lot of integration overhead (model, mako, etc.) and no added value / information, what do you think?
* ch.swisstopo.geologie-geomol-isotherme_60
* ch.swisstopo.geologie-geomol-isotherme_100
* ch.swisstopo.geologie-geomol-isotherme_150

PR #3115, opened 2019-02-25
Update tooltip of layer ch.are.wohnungsinventar-zweitwohnungsanteil

alcapat:
Update model and mako to display new attributes. 
Get translations from bod (not in data table)


pakb:
Is this intended to be deployed next week?

PR #3123, opened 2019-03-07
identify support for EsriGeometryPoint

procrastinatio:
See https://github.com/geoadmin/mf-chsdi3/issues/3122


ltkum:
Seems good to me.

PR #3141, opened 2019-03-21
OEREB get egrid from service

rebert:
@p1d1d1 [try](https://mf-chsdi3.dev.bgdi.ch/oereb_get_egrid_from_service/shorten/80cc5287e3)

Geneva not working:
https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSMissingAllowOrigin

### DIFF
--- a/chsdi/templates/htmlpopup/oerebkataster.mako
+++ b/chsdi/templates/htmlpopup/oerebkataster.mako
@@ -14,10 +14,10 @@ def get_xml(path):
        xmlstring = re.sub(r'\sxmlns[^"]+"[^"]+"', '', response.text)
        root = et.fromstring(xmlstring)
        #egrid = root.find('{http://schemas.geo.admin.ch/V_D/OeREB/1.0/Extract}egrid').text
-       egrid = root.findall('.//egrid')[0].text # findall, as the cantons have different structures
+       list_egrid = root.findall('.//egrid') # findall, as the cantons have different structures
     except:
       pass
-    return egrid
+    return list_egrid
 
 %>
 <%def name="table_body(c,lang)">
@@ -34,7 +34,7 @@ is_oereb_service = c['attributes']['oereb_webservice'] != None and c['attributes
 
 if is_oereb_service:
   url_get_egrid = "{}{}{}".format(c['attributes']['oereb_webservice'], path_xml, coord)
-  egrid = get_xml(url_get_egrid)
+  list_egrid = get_xml(url_get_egrid)
 %>
     <tr><td class="cell-left">${_('kanton')}</td>    <td>${c['attributes']['kanton'] or '-'}</td></tr>
     <tr><td class="cell-left">${_('gemgemeinde')}</td>    <td>${c['attributes']['gemeindename'] or '-'}</td></tr>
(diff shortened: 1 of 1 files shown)


p1d1d1:
Does it apply to Geneva only?

Il giorno giovedì 21 marzo 2019, Tobias Reber <notifications@github.com> ha
scritto:

p1d1d1:
Why it doesn’t affect the extract request on prod

Il giorno giovedì 21 marzo 2019, Pasquale Di Donato <
pasquale.didonato@gmail.com> ha scritto:

chsdi/templates/htmlpopup/oerebkataster.mako
```diff
@@ -52,7 +54,28 @@ if not 'oereb_webservice' in c['attributes'].keys():
     % if c['attributes']['oereb_webservice'] != None and c['attributes']['bgdi_status'] == 0:
         <tr>
             <td class="cell-left">${_('ch.swisstopo-vd.stand-oerebkataster.oereb_webservice')}</td>
-            <td><a target="_blank" href="${c['attributes']['oereb_webservice']}${path}${c['attributes']['egris_egrid']}">PDF</a></td>
+            <td id="${c['featureId']}">Loading link to PDF ...</td>
         </tr>
+       <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.1.1/jquery.min.js"></script>
+       <script src="https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.16.0/moment.min.js"></script>
```
loicgasser: This is a use case for an iframe tooltip since you load javascript :see_no_evil:

rebert: I will! thanks Loic

rebert:
@p1d1d1 now without cross-site problems.

But, almost each canton has a different xml structure... phhh

p1d1d1:
Should not be. Will check after holidays

Il ven 22 mar 2019, 13:29 Tobias Reber <notifications@github.com> ha
scritto:

p1d1d1:
@rebert this is not yet in prod, right?
PS: I've just discovered that the getegrid request may return more than one egrid, e.g. https://www.oereb2.apps.be.ch/OerbverSVC.svc/getegrid/?XY=2606296.3,1215309.4
Do you have an idea on how to deal with that? Do we show a PDF link for each egrid?

rebert:
@p1d1d1 right now this has not been considered yet... to be discussed

p1d1d1:
@rebert got OK from data owner

rebert:
@procrastinatio we are talking about Geneva ;-)

p1d1d1:
@rebert is not working for Basel-Landschaft and Schwyz

p1d1d1:
@rebert could you please debug this?

rebert:
@p1d1d1 Yes, I will, but:
- You had tested this solution and approved for prod!
- I don't like the fact that we have to adapt to 26 different XMLs ... maintanance is expensive (and boring)

p1d1d1:
@rebert I've just asked you to debug in order to identify where the problem is! Then we can discuss for a solution
> You had tested this solution and approved for prod!

Yes, so I did a mistake. Then what? Do we continue to have on prod partially working solutions?
> I don't like the fact that we have to adapt to 26 different XMLs ... maintanance is expensive (and boring)

I'd like to avoid this also. If you want you can use the service from canton BL as a reference, since it is the only one conform to the XML schema. If the consequence will be that on map.geo.admin.ch we'll only have this canton, then we can  discuss and explain the situation with the data owner.

PR #3142, opened 2019-03-22
ch.swisstopo.vec25-gewaessernetz_referenz

Brunim28:
change in the model + new mako


Brunim28:
@procrastinatio, this layer goes finally on prod on the 29 avril (milestone was not updated, sorry). Since it's a first integration, merging this PR is not a problem but the staging in the BOD will stay on "integration" and not on prod.

PR #3143, opened 2019-03-26
New layer ch.bfe.kehrichtverbrennungsanlagen

Sgachet:
[Test link](https://mf-chsdi3.dev.bgdi.ch/gas_kehrichtverbrennungsanlagen/shorten/80f556df0c)


chsdi/models/vector/uvek.py
```diff
@@ -2162,6 +2162,29 @@ class Kernkraftwerke(Base, Vector):
...
+    id = Column('bgdi_id', Integer, primary_key=True)
+    wasteincinerationplantr = Column('wasteincinerationplantr', Unicode)
+    xtf_id = Column('xtf_id', Unicode)
+    number_id = Column('number_id', Integer)
+    name = Column('name', Unicode)
+    place = Column('place', Unicode)
+    beginningofoperation = Column('beginningofoperation', Integer)
+    web = Column('web', Unicode)
+    year = Column('year', Integer)
+    recycledwaste = Column('recycledwaste', Float)
```
ltclm: do we really need float? the data has only integer values for ``recycledwaste`` and ``electricity``?

ltclm: maybe the data gets rounded somewhere in the fme workbench?

Sgachet: I think that in the future, some data would be in float. I propose to keep like this

PR #3148, opened 2019-03-31
show all exact search results before wildcard search

ltclm:
@davidoesch @gjn
**new search behaviour**
instead of 10 results of exact prefix match search, all the results of exact prefix search will be prepended to the wildcard search results.

[Testlink](https://mf-chsdi3.dev.bgdi.ch/dev_ltclm_esel/shorten/810052d6fc)
search for 
* ``esel``
* ``loch``
* ``quartier``

### DIFF
--- a/chsdi/views/search.py
+++ b/chsdi/views/search.py
@@ -133,12 +133,10 @@ class Search(SearchValidation):
             try:
                 # standard wildcard search
                 self.sphinx.AddQuery(searchTextFinal, index='swisssearch')
-                # exact search, first 10 results
-                self.sphinx.SetLimits(0, 10)
+                # exact search with priority
                 searchText = '@detail ^%s' % ' '.join(self.searchText)
                 self.sphinx.AddQuery(searchText, index='swisssearch')
                 # reset settings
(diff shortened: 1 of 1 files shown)


davidoesch:
Just wondering "Bahnhofstrasse 23"
The sorting of the places... Is based on what? Arbitrary?

gjn:
Yes, there is no explicit ordering when we have many results for locations. We could use population size in addresses...(I think there is an ordering in the tlm datasets which could be taken. But that's no easy task with current setup).

I think this PR is good. Thanks @ltclm

ltclm:
there was an initial sort of the addresses, but the sphinx view got changed somehow and this sort got lost.
i have fixed the view, the initial sort of the address data is as follows:
Unbenannt

will be applied in a few hours, have to rebuild sphinx indexes.

danduk82:
I think it's good. What do we break if we apply this like this?  Before changing last time we asked our clients for feedback and waited more than one month before going live. @gjn what do you think, *let's go like this* in next deploy?

@ltclm I don't think that the performance will be impacted, do you? Maybe can you test it with the same yandex setup as last time just to be sure just once?

PR #3165, opened 2019-04-24
Limiting featureIds number pro request

procrastinatio:
for `features service`, limit number of id in a single request to something

Too many
http://mf-chsdi3.dev.bgdi.ch/mom_max_ids/rest/services/api/MapServer/ch.bafu.nabelstationen/1,2,3,4,5,6,7,8,9,1,2,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2

OK (two features)
http://mf-chsdi3.dev.bgdi.ch/mom_max_ids/rest/services/api/MapServer/ch.bafu.nabelstationen/RIG,LAU

### DIFF
--- a/chsdi/lib/validation/features.py
+++ b/chsdi/lib/validation/features.py
@@ -59,6 +59,8 @@ class HtmlPopupServiceValidation(BaseFeaturesValidation):
     def featureIds(self, value):
         if value is not None:
             self._featureIds = value.split(',')
+            if len(self._featureIds) > int(self.request.registry.settings['max_featureids_request']):
+                raise HTTPBadRequest('Too many featureIds')
         else:
             raise HTTPBadRequest('Please provide featureIds')  # pragma: no cover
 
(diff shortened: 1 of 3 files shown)


pakb:
Is it used by `/identify`?
If so, shouldn't we adapt also how mf-geoadmin3 is requesting feature infos? It is easy to do a rectangle selection with a dense layer to have that much features in one request.

boecklic:
geoadmin makes calls to `MapServer/identify?geometry=` with geometry being an extent or a point, and subsequently to `MapServer/ch.astra.unfaelle-personenschaeden_alle/3AE1822AABFC0024E0530A8394275CF2/htmlPopup?coord=2664550,1232200`, so no direct calls to feature service, `/identify` is not imho not directly affected by this limitation.

ltclm:
we could not find any requests of that type within the last 6 months in kibana, it seems to be a hidden feature which is not used directly.

PR #3167, opened 2019-04-24
Features reprojection

procrastinatio:
This PR 
* SQLAlchemy and Geoalchemy update (security, https://github.com/geoadmin/mf-chsdi3/network/alert/requirements.txt/SQLAlchemy/open)
* All feature/geom transformations outside the DB (hopefully fixing #3172 and  #2965)

[Demo](https://mf-chsdi3.int.bgdi.ch/sqlalchemy_update)

Identify in **WGS1984**...
https://mf-chsdi3.int.bgdi.ch/sqlalchemy_update/rest/services/api/MapServer/ch.bafu.nabelstationen/RIG,LAU?sr=21781&geometryFormat=geojson
...and **LV03**
https://mf-chsdi3.int.bgdi.ch/sqlalchemy_update/rest/services/api/MapServer/ch.bafu.nabelstationen/RIG,LAU?sr=4326&geometryFormat=geojson

PR #3186, opened 2019-05-16
Using a single worker

procrastinatio:
Clearly not the best solution, but fixing the SRID mess is not easy.

[Demo](http://mf-chsdi3.dev.bgdi.ch/mom_wsgi/)

### DIFF
--- a/production.ini.in
+++ b/production.ini.in
@@ -8,6 +8,8 @@ use = egg:chsdi
 app_version = ${app_version}
 entry_path = ${apache_entry_path}
 available_languages = de fr it en rm
+# Using a unique worker (issues with models shared between workers #3172)
+workers = 1
 
 pyramid.reload_templates = false
 pyramid.debug_authorization = false


loicgasser:
yes this is a problem. Try to update geoalchemy2. Geoalchemy is now able to read the `EWKB` data so you could read it from
https://github.com/geoalchemy/geoalchemy2/blob/master/geoalchemy2/elements.py#L183

instead of using srid_out for the json response
https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/models/vector/__init__.py#L109

so `geom.srid`

procrastinatio:
Thanks for the input @loicgasser 
But the main issue we have, requests are setting the `srid_out` on the models, to reproject the geometries, which is not a good idea as the models are global objects. Some servers may cope with that, but not apache with our current settings.

loicgasser:
yes I know, you can't have several workers per process. But it seems that the geometries are re-projected correctly, the srid in the JSON output is incorrect. Reading it directly from the geometry could help mitigate the issue.

loicgasser:
A solution to this problem could be to dynamically create as many instances of the same model with the different srid outputs instead and select the appropriated one in `models_from_bodid` instead of mutating models state. you need to make sure you work with `extend_existing=True`.
Not sure what is best

procrastinatio:
@loicgasser I moved the reprojection outside the DB/models, as it has already done for `search`. Maybe not the best approach, but at least it is manageable

See https://github.com/geoadmin/mf-chsdi3/pull/3167

procrastinatio:
Closing as we are now reprojection the `Shapely` features.

PR #3192, opened 2019-05-23
fix ch.bfe.energiestaedte-2000watt-areale

Brunim28:
Fix mako + model


Brunim28:
[test link](https://mf-chsdi3.dev.bgdi.ch/brg_fix_2000watt-areale/shorten/821084d0cb)

ltclm:
@AFoletti do we have to keep the attribute ``kategorie`` in the identify result and in the model?
It is not referenced in htmlpopup anymore.

AFoletti:
With the code we ensure m2m compatibility. I would keep it

PR #3195, opened 2019-05-27
[WIP] we could run the tests in parallel, couldnt we?

rebert:
make fasttest runs the tests in parallel on 4 processes:

make test: 612 sec.
make fasttest: 273 sec.

https://nose.readthedocs.io/en/latest/plugins/multiprocess.html

What do you think about it?


rebert:
how would Jenkins like to run the tests in parallel on 4CPUs?

pakb:
replaced by https://github.com/geoadmin/mf-chsdi3/pull/3200

PR #3211, opened 2019-06-17
fixed the tests, but dont know if it is ok

rebert:
- fixed tests, but don't know it it is ok
- Makefile does not call nosetest paralell any more, because it does not return all error

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -190,11 +190,11 @@ shell:
 
 .PHONY: test
 test:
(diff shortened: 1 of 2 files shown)


rebert:
quess that it is related to:
https://github.com/geoadmin/mf-chsdi3/commit/3a11e34856f39365df4647649368f3922b757949

rebert:
test

rebert:
https://github.com/geoadmin/mf-chsdi3/pull/3216

Issue #1808, opened 2015-12-10
Store last access for shortened URL's


gjn:
...to eventually be able to delete urls that werene't accessed in x years.

@davidoesch @cedricmoullet Agree to add this feature?

cedricmoullet:
+1

loicgasser:
That's the only way. so +1

davidoesch:
+1

for X years is 4

procrastinatio:
* Impact on performance...
* We need to split de DB into 3, as we only have a prod DB.

procrastinatio:
@danduk82   Maybe with https://jira.swisstopo.ch/browse/BGDIINF_SB-445

procrastinatio:
We never ever close/shutdown/delete/remove anything.

procrastinatio:
Small recap

- `POST/PU/DELTE` via service (chsdi3)
- `GET` directly via varnish and s3

Considering that

- We have to reroute `GET` on the service, which impacts performance and stability
- We never ever delete anything

Does it make sense to have the last access date if we never delete anything?

davidoesch:
Last access date. Keep it. Whenever we need to actually delete it we can proceed according to our TERMS AND CONDITIONS stated in Impressum

procrastinatio:
Decision?
Do we re-varnish on the service? Yes/no
For now, we only have a single DynamoDB for all environments. Should we extend? Yes/No

davidoesch:
@chris

PR #3215, opened 2019-06-21
[WIP] Updating last_access timestamp when getting a file

procrastinatio:
Currently, files are served directly by varnish/S3.
In order to update the last_access timestamp, we have to use the service.

See #1808


danduk82:
I don't think it is worth to use chsdi for GET accesses. Currently we use directly varnish-S3, and this is way more efficient.

procrastinatio:
Another great example of _let's do it first, we'll think about it later_

Closing as we are moving to boto3

Issue #3218, opened 2019-06-25
Sphinx paging for spatial queries sometimes too low


AFoletti:
See customer email in webgis: 23. Juni 2019 um 10:42 (Betreff: Grundstück Nummer finden)

**Situation:**
Sometimes a search query with a searchText **AND** a bbox does not return any result. The example is proposed with a parcel search.

**Steps to reproduce:**
Based on two locations
1. [here](https://map.geo.admin.ch/?lang=en&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.kantone.cadastralwebmap-farbe&E=2538177&N=1165702&zoom=11&crosshair=marker) (OK)

2. [here](https://map.geo.admin.ch/?lang=en&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.kantone.cadastralwebmap-farbe&E=2538525&N=1165844&zoom=11&crosshair=marker) (NOK)

- Open [this search query](http://api3.geo.admin.ch/rest/services/api/SearchServer?searchText=Echallens&origins=parcel&type=locations&bbox=2538177,1165702,2538177,1165702&sr=2056) corresponding to point 1. A result is returned
- Open [this second search query](http://api3.geo.admin.ch/rest/services/api/SearchServer?searchText=Echallens&origins=parcel&type=locations&bbox=2538525,1165844,2538525,1165844&sr=2056) corresponding to point 2. No result is returned even if the point finds itself in the same parcel as the first one

**Expected result:**
- The second query returns exactly the same result as the first one.

**Initial analysis:**
Today @ltclm already had a look at the issue. Here the first findings:
- Both points are in the same parcel. See

```SQL
-- identify
SELECT * FROM vd.view_os_realestate_cwm WHERE st_intersects(the_geom,st_setsrid(st_point(2538525,1165844),2056)) -- 92951298
SELECT * FROM vd.view_os_realestate_cwm WHERE st_intersects(the_geom,st_setsrid(st_point(2538177,1165702),2056)) -- 92951298

-- search
SELECT * FROM vd.parzellen_sphinx WHERE id = 92951298
"BOX(2536789.688 1164864.794,2539019.757 1166924.984)"
```
thus both requests should return the same result
- Sphinx passes only 50 results to chsdi from the text search to evaluate against the geometry. If the match is not among those first 50 results, nothing is found

**Solution ideas:**
[...]

PR #3222, opened 2019-06-27
increase number of sphinx results in bbox search

ltclm:
there is a risk that the object we're looking for is not within the first 50 results of the sphinx search when doing a bbox search.

this pr will increase the number of results in the initial spinx query to 150 in case of bbox search.
the quadindex matching is done in a post-processing on the sphinx query results in the function ``parse_location_results()``. 

[First Query](http://mf-chsdi3.dev.bgdi.ch/dev_ltclm_bbox_search/rest/services/api/SearchServer?searchText=Echallens&origins=parcel&type=locations&bbox=2538177,1165702,2538177,1165702&sr=2056)
[Second Query](http://mf-chsdi3.dev.bgdi.ch/dev_ltclm_bbox_search/rest/services/api/SearchServer?searchText=Echallens&origins=parcel&type=locations&bbox=2538525,1165844,2538525,1165844&sr=2056)

fix for https://github.com/geoadmin/mf-chsdi3/issues/3218

### DIFF
--- a/chsdi/views/search.py
+++ b/chsdi/views/search.py
@@ -98,16 +98,18 @@ class Search(SearchValidation):
 
     def _swiss_search(self):
         limit = self.limit if self.limit and self.limit <= self.LOCATION_LIMIT else self.LOCATION_LIMIT
-        self.sphinx.SetLimits(0, limit)
 
         # Define ranking mode
         if self.bbox is not None and self.sortbbox:
             coords = self._get_geoanchor_from_bbox()
             self.sphinx.SetGeoAnchor('lat', 'lon', coords[1], coords[0])
             self.sphinx.SetSortMode(sphinxapi.SPH_SORT_EXTENDED, '@geodist ASC')
+            limit = 150
         else:
             self.sphinx.SetRankingMode(sphinxapi.SPH_RANK_WORDCOUNT)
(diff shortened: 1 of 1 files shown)


AFoletti:
This will lower the risk but is not a real solution.
Nevertheless, I am totally fine with it. The risk should now be really really low and I do not think that the issue merits more time than already spent on it

chsdi/views/search.py
```diff
@@ -98,17 +98,18 @@ def _fuzzy_search(self, searchTextFinal):
...
     def _swiss_search(self):
         limit = self.limit if self.limit and self.limit <= self.LOCATION_LIMIT else self.LOCATION_LIMIT
-        self.sphinx.SetLimits(0, limit)
-
         # Define ranking mode
         if self.bbox is not None and self.sortbbox:
             coords = self._get_geoanchor_from_bbox()
             self.sphinx.SetGeoAnchor('lat', 'lon', coords[1], coords[0])
             self.sphinx.SetSortMode(sphinxapi.SPH_SORT_EXTENDED, '@geodist ASC')
+            limit = 150
```
procrastinatio: I would set this variable at the class level, like the other one

boecklic: done

PR #3235, opened 2019-07-29
updated translations

AFoletti:


rebert:
@AFoletti we are going to re-translate anyway

PR #3247, opened 2019-08-27
new attributes in model and identify template

ltclm:
https://jira.swisstopo.ch/browse/BGDIDI_KB-2024?


ltclm:
@mariokeusen has confirmed that this can be published with the next deploy 2019-09-10

PR #3251, opened 2019-09-02
[WIP] create a makefile target to update db indexes of queryable_attributes

ltclm:


procrastinatio:
@ltclm  What is the status of this? It has to work on python3, if we want to merge it.

PR #3262, opened 2019-09-06
Shop item images : adding a max constraint instead of fixed width and height

pakb:
fixes https://github.com/geoadmin/mf-chsdi3/issues/3261


pakb:
test link here : https://mf-chsdi3.int.bgdi.ch/ltbtp_fix_toposhop_image_size/shorten/84310b60e5
@pauloamado could you check that it works as you intended? (it changes all shop items image size, is there other items than maps that are shown this way?)

PR #3266, opened 2019-09-17
Translation service

procrastinatio:
[Demo](https://mf-chsdi3.int.bgdi.ch/mom_translations/rest/services/translations?lang=en) or
[Demo 2](https://mf-chsdi3.int.bgdi.ch/mom_translations/configs/de/translations.json)


ltclm:
that's a great new service, thank you @procrastinatio !

@procrastinatio @danduk82 @boecklic 
What do you think of replacing the po/mo based translations with this service inside chsdi too? this would ease the automated publication of new metadata (without code deploy of po/mo files) but
there will be an impact on performance of the affected services (p.e. identify).

PR #3272, opened 2019-09-23
Escaping incoming params (in mako template) before passing them to JS

pakb:
impact : LUBIS and Zeitreihen related tooltips


gjn:
Can we have a testlinks that demonstrates that it fixes the issue?

procrastinatio:
What is issue is this supposed to fix ?

PR #3283, opened 2019-10-03
Python2 update

procrastinatio:
Outaded modules won't run with Python 3.7

[Demo](https://mf-chsdi3.int.bgdi.ch/python2_update)

The following package will not be updated
```
Package Version Latest Type 
------- ------- ------ -----
pyproj  1.9.6   2.2.2  wheel
```

Starting with `pyproj2` a version of `libproj6` is needed.


chsdi/static/doc/source/conf.py
```diff
@@ -50,7 +50,7 @@
 
 # General information about the project.
 project = u'GeoAdmin API'
-copyright = u'2017, COGIS'
+copyright = u'2013-2019, COGIS'
```
pakb: what about something like `u'2013-' + datetime.datetime.now().year`? So we don't have to change it ever again

chsdi/static/doc/source/conf.py
```diff
@@ -127,7 +127,7 @@
 
 # The name of an image file (relative to this directory) to place at the top
 # of the sidebar.
-html_logo = "_static/bg_header_logo.png"
+# html_logo = "_static/bg_header_logo.png"
```
pakb: Do we need to keep this?

chsdi/static/doc/source/conf.py
procrastinatio: Yes, nice idea.

PR #3287, opened 2019-10-07
CHSDI rohstoffe_ziegel_verarbeitung

Simon1193:
[testlink](https://mf-chsdi3.int.bgdi.ch/ltbei_rohstoffe_ziegel_verarbeitung/shorten/84f4e5a2a1)


chsdi/models/vector/stopo.py
```diff
@@ -1878,6 +1878,26 @@ class GeologieRohstoffeZementAbbauVerarbeitung(Base, Vector):
 register('ch.swisstopo.geologie-rohstoffe-zement_abbau_verarbeitung', GeologieRohstoffeZementAbbauVerarbeitung)
 
 
+class GeologieRohstoffeZiegelVerarbeitung(Base, Vector):
+    __tablename__ = 'rohstoffe_ziegel_verarbeitung'
+    __table_args__ = ({'schema': 'geol', 'autoload': False})
+    __bodId__ = 'ch.swisstopo.geologie-rohstoffe-ziegel_verarbeitung'
+    __template__ = 'templates/htmlpopup/rohstoffe_ziegel_verarbeitung.mako'
+    __queryable_attributes__ = ['obname', 'pckind']
```
ltclm: these attributes are supposed to be searchable, so they have to be part of the detail column in sphinx, you have to remove the __queryable_attributes__ property, this is for the configuration of  the object information tool

image

chsdi/models/vector/stopo.py
```diff
@@ -1878,6 +1878,26 @@ class GeologieRohstoffeZementAbbauVerarbeitung(Base, Vector):
...
+    __label__ = 'obname'
+    id = Column('bgdi_id', Integer, primary_key=True)
```
ltclm: there is a unique and not null attribute ``id_objekt`` in the table, if this attribute is provided by the data owner please use this attribute as primary key instead, but please verifiy the source of this attribute first. if it is only a meaningless counter, we keep bgdi_id.

chsdi/models/vector/stopo.py
ltclm: therefore you should prepare a sphinx pr with this layer.

AFoletti:
I allowed myself to use this PR for a trivial geology legend fix 😉

chsdi/models/vector/stopo.py
Sgachet: Already done, see [sphinx](https://github.com/geoadmin/service-sphinxsearch/pull/409)

chsdi/models/vector/stopo.py
Sgachet: For each Rohstoffe layer, the primary key is put on the bgdi_id, I think we can leave like that here too

ltclm: yes you're right

PR #3288, opened 2019-10-07
Add new layer ch.bfe.erneuerbarheizen

Sgachet:
[Test link](https://mf-chsdi3.int.bgdi.ch/gas_erneuerbarheizen/shorten/84f9ea7888)


chsdi/models/vector/uvek.py
```diff
@@ -807,6 +807,47 @@ class StatistikwasserkraftanlagenNew(Base, Vector):
...
 
 
+class Erneuerbarheizen(Base, Vector):
+    __tablename__ = 'erneuerbarheizen'
+    __table_args__ = ({'schema': 'bfe', 'autoload': False})
+    __template__ = 'templates/htmlpopup/erneuerbarheizen.mako'
+    __bodId__ = 'ch.bfe.erneuerbarheizen'
+    __queryable_attributes__ = ['company']
+    __label__ = 'company'
+    id = Column('bgdi_id', Integer, primary_key=True)
```
ltclm: i would use xtf_id as id / primary key in the model and also create an pk constraint / Index in the table.

chsdi/models/vector/uvek.py
```diff
@@ -807,6 +807,47 @@ class StatistikwasserkraftanlagenNew(Base, Vector):
...
+    street = Column('street', Unicode)
+    streetnumber = Column('streetnumber', Unicode)
+    street_streetnumber = Column('street_streetnumber', Unicode)
+    postalcode = Column('postalcode', Integer)
+    place = Column('place', Unicode)
+    pc_place = Column('pc_place', Unicode)
+    website = Column('website', Unicode)
+    privatecontrol = Column('privatecontrol', Unicode)
+    additionalinformation = Column('additionalinformation', Unicode)
+    consulting_type_de = Column('consulting_type_de', Unicode)
```
ltclm: do we really need these attributes in the model / identify result? they're not referenced in the mako.

Sgachet: yes, the bfe need all the attributes in the API...

chsdi/models/vector/uvek.py
Sgachet: Make sense, i do the change

PR #3296, opened 2019-10-14
Add model/mako ch.bfe.komo-projekte

Sgachet:
[Test link](https://mf-chsdi3.int.bgdi.ch/gas_komo_proj/shorten/84f9cb81cc)


chsdi/models/vector/uvek.py
```diff
@@ -748,6 +748,62 @@ class FernWaermeAngebot(Base, Vector):
...
+    web_de = Column('web_de', Unicode)
+    web_fr = Column('web_fr', Unicode)
+    web_it = Column('web_it', Unicode)
+    web_en = Column('web_en', Unicode)
+    mail = Column('mail', Unicode)
+    aim_de = Column('aim_de', Unicode)
+    aim_fr = Column('aim_fr', Unicode)
+    aim_it = Column('aim_it', Unicode)
+    aim_en = Column('aim_en', Unicode)
+    measure_de = Column('measure_de', Unicode)
```
ltclm: attribute measure_* is not referenced in mako, do we really need it in the model?

chsdi/models/vector/uvek.py
```diff
@@ -748,6 +748,62 @@ class FernWaermeAngebot(Base, Vector):
...
 
 
+class KomoProjekte(Base, Vector):
+    __tablename__ = 'komo_projekte'
+    __table_args__ = ({'schema': 'bfe', 'autoload': False})
+    __template__ = 'templates/htmlpopup/komo.mako'
+    __bodId__ = 'ch.bfe.komo-projekte'
+    __queryable_attributes__ = ['project_title_de', 'project_title_fr', 'project_title_it', 'project_title_en']
+    __label__ = 'id'
+    id = Column('bgdi_id', Integer, primary_key=True)
```
ltclm: please use xtf_id as primary key

chsdi/models/vector/uvek.py
AFoletti: As for erneuerbarheizen, those attributes are needed in the API for external applications. This is an explicit request from the DH

PR #3299, opened 2019-10-22
Fixing tests

procrastinatio:
* ini file
* removing nose parallel


rebert:
@procrastinatio merci!

PR #3304, opened 2019-10-30
[WIP] do not filter out large geometries with shop layers

ltclm:
a geom size filter has been introduced with:
https://github.com/geoadmin/mf-chsdi3/pull/1621
and modified with:
https://github.com/geoadmin/mf-chsdi3/pull/3019

this pr adds an exception for shop layers.
we have complex geometries and were not allowed to modify (generalize, thin out, ...) because they are used for prize calculations.

### DIFF
--- a/chsdi/models/vector/__init__.py
+++ b/chsdi/models/vector/__init__.py
@@ -157,7 +157,7 @@ class Vector(object):
         return cls.__mapper__.columns[geom_column_name]
 
     def ignore_max_feature_geometry_size_column(cls):
-        return cls.__mapper__columns[cls.__ignore_max_feature_geometry_size__] if hasattr(cls, '__ignore_max_feature_geometry_size__') else False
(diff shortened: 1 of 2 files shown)


rebert:
@boecklic what about this switch?

chsdi/models/vector/__init__.py
```diff
@@ -86,7 +86,8 @@ def __read__(self):
                             len(self._shape) < MAX_FEATURE_GEOMETRY_SIZE:
                         geom = self._shape
                     elif val is not None and \
-                            len(val.data) < MAX_FEATURE_GEOMETRY_SIZE:
+                            (len(val.data) < MAX_FEATURE_GEOMETRY_SIZE or
+                            self.__ignore_max_feature_geometry_size__):
```
procrastinatio: add `__ignore_max_feature_geometry_size__` to the base Vector model. If not defined, you get an error.

PR #3308, opened 2019-11-05
Releasenotes for Nov. 13th

rebert:


rebert:
@Sgachet @Simon1193 @AFoletti @ltclm @alcapat @daguer here the release notes for next deploy

PR #3317, opened 2019-11-20
stop mf-chsdi3 error log flooding with invalid search text

ltclm:
this will stop the flooding of mf-chsdi3 error logs when invalid search queries are sent to search service like:
```
http://localhost:9076/rest/services/api/SearchServer?sr=2056&searchText="Arbon&lang=fr&type=locations
```

it is also fixing query strings by removing double quotes from the input parameter

[Testlink](http://mf-chsdi3.dev.bgdi.ch/dev_ltclm_fix_search_errors/rest/services/api/SearchServer?sr=2056&searchText=%22Arbon&lang=fr&type=locations)

triggered by https://jira.swisstopo.ch/browse/BGDIINF_SB-895

### DIFF
--- a/chsdi/lib/helpers.py
+++ b/chsdi/lib/helpers.py
@@ -153,6 +153,8 @@ def format_locations_search_text(input_str):
         return input_str
     # only remove trailing and leading dots
     input_str = ' '.join([w.strip('.') for w in input_str.split()])
+    # remove double quotation marks
+    input_str = input_str.replace('"', '')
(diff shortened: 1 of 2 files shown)


ltclm:
the second error message from jira is due to an overcharged instance in our current cluster.
the server error message is:
```
ValueError: no results from sphinx service (temporary searchd error: server maxed out, retry in a second)
```
with the last commit of this pr, this information will be printed to the vhost error log.

PR #3324, opened 2019-12-02
Update script createall.sh for weather legends

pakb:
updating script so that it functions with node > 10


rebert:
https://github.com/geoadmin/mf-chsdi3/commit/f9cd67e5fb604de7a3a016e8cc5e6ad9d68ea9c2

PR #3325, opened 2019-12-02
working with a more recent node version

rebert:


rebert:
thanks to @pakb https://github.com/geoadmin/mf-chsdi3/pull/3324

pakb:
you should also include changes made on `package.json` as we've adapted the node script for `node-canvas` 2.6.1

PR #3327, opened 2019-12-04
Removing uneeded module import

procrastinatio:
* Not used
* breaking with `python2-dev`
* Renamed in python3


ltkum:
> I would also remove `configparser==4.0.2` from requirements.txt, but those are the only two uses of ConfigParser in the whole project, so everything else looks good to me.

I did it, if checks pass, you can go ahead and merge it

procrastinatio:
Cool

PR #3329, opened 2019-12-04
Check if doc i generated

procrastinatio:
Better check, as it is the landing page for http://api3 .geo.admin.ch


procrastinatio:
@ltkum Thx

PR #3345, opened 2019-12-16
releasenotes 4 deploy 20191218

rebert:


rebert:
Hej @AFoletti @ltclm @Simon1193 @daguer @alcapat 

Have fun with the last releasenotes of the year.

PR #3346, opened 2019-12-18
zero as well

rebert:


tests/integration/test_identify_service.py
```diff
@@ -80,6 +80,17 @@ def test_identify_without_mapextent(self):
         self.assertEqual(resp.json['code'], 400)
         self.assertEqual(resp.json['status'], 'error')
 
+    def test_identify_with_0_0_0_in_imagedisplay(self):
```
procrastinatio: No. please explain what an image with height and width of zero is.
It generates also many ZeroDivisionException

rebert: Imagine that you have zero cookies and you split them evenly among zero friends. How many cookies does each person get? See? It doesn’t make sense. And Cookie Monster is sad that there are no cookies, and you are sad that you have no friends.

Issue #3348, opened 2019-12-18
imageDisplay and tolerance in identify


procrastinatio:
See https://jira.swisstopo.ch/browse/BGDIINF_SB-944

rebert:
https://github.com/geoadmin/mf-chsdi3/pull/3349

ltkum:
The issue is closed in Jira and the PR solving it has been merged. I close the Issue here too

PR #3349, opened 2019-12-18
adaption doc not using imageDisplay=0,0,0 any more

rebert:


rebert:
@procrastinatio thanks for investigating in #3348

Issue #3351, opened 2019-12-18
Layer search service and geometryFormat


fredj:
Hi all,

According to the documentation and the examples in #3268 description, it should be possible to search for layers with the `geometryFormat` as `geojson`.

Is it the case? I'm receiving HTTP 500 errors with:
https://api3.geo.admin.ch/rest/services/api/SearchServer?type=layers&searchText=aerial&geometryFormat=geojson

davidoesch:
@procrastinatio  is looking into that

PR #3353, opened 2019-12-18
Search 'layers' returns GeoJSON

procrastinatio:
Returning a valid `geojson`

The returned `bbox` and `Polygon` do not have a real meaning (it's the default bbox), because we don't have the `layer` extent stored in the `sphinxapi` indices. Feel free to open an issue in https://github.com/geoadmin/service-sphinxsearch/issues if you like.

**Demo**
http://mf-chsdi3.int.bgdi.ch/fix_3351/rest/services/api/SearchServer?searchText=aerial&type=layers&geometryFormat=geojson&sr=3857

Fix https://github.com/geoadmin/mf-chsdi3/issues/3351

### DIFF
--- a/chsdi/lib/mortonspacekey.py
+++ b/chsdi/lib/mortonspacekey.py
@@ -20,7 +20,7 @@ class BBox:
         self.maxy = float(maxy)
 
     def __repr__(self):
-        return 'BBox(%s,%s,%s,%s)' % (self.minx, self.miny, self.maxx, self.maxy)
+        return 'BBox({}, {}, {}, {})'.format(*self.bounds)
 
     def __eq__(self, other):
         if (self.minx == other.minx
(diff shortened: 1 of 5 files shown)


chsdi/lib/mortonspacekey.py
```diff
@@ -20,7 +20,7 @@ def __init__(self, minx, miny, maxx, maxy):
         self.maxy = float(maxy)
 
     def __repr__(self):
-        return 'BBox(%s,%s,%s,%s)' % (self.minx, self.miny, self.maxx, self.maxy)
+        return 'BBox({},{},{},{})'.format(*self.bounds)
```
ltkum: this is better

fredj:
There's an issue with the `bbox` value in this branch but only for the locations results:
https://mf-chsdi3.int.bgdi.ch/fix_3351/rest/services/api/SearchServer?limit=1&searchText=lausanne&geometryFormat=geojson&sr=4326&type=locations

The bbox value is:
```
"bbox": [
  -0.163847,
  44.890639,
  -0.163846,
  44.89064
]
```

It only applies to the location results, the layers results are correct

procrastinatio:
@fredj The was simply not pushed to the server (the shortcomings of a still largely manual CI/CD)

PR #3356, opened 2020-01-06
[WIP] CHSDI laermbelastung_eisenbahn_zulaessige_immissionen (tag und nacht)

Simon1193:


chsdi/templates/htmlpopup/laerm_eisenbahn_zulaessige_immissionen_nacht.mako
```diff
@@ -0,0 +1,18 @@
...
+<%def name="table_body(c, lang)">
+
+<%
+  lang = lang if lang in ('fr','it') else 'de'
+  lang = lang if lang != 'it' else 'fr'
+  es = '%s_es' % lang
+  pointofdetermination_t = '%s_pointofdetermination_t' % lang
+%>
+
+<% c['stable_id'] = True %>
```
ltclm: you should not use stable_id TRUE if the primary key is an db sequence. This should option should only be used if the primary key is part of the data delivery and declared as stable by data owner.

ltclm:
since jenkins is down now, i have run the integration tests locally, they are passing without errors.

chsdi/templates/htmlpopup/laerm_eisenbahn_zulaessige_immissionen_nacht.mako
Simon1193: > you should not use stable_id TRUE if the primary key is an db sequence. This should option should only be used if the primary key is part of the data delivery and declared as stable by data owner.

OK thanks for your feedback. I will adapt it then

PR #3359, opened 2020-01-13
 Makefile adaptation - change version setuptools to 44.0.0

Simon1193:


Simon1193:
Since the actual version of setuptools requires => python 3.5, we change the version of setuptools in the Makefile to an anterior one (version 44.0.0) in order to be able to run the script

Makefile
```diff
@@ -487,16 +487,16 @@ requirements.txt:
...
 		${PIP_CMD} install --index-url ${PYPI_URL}  -e .
 else
 .venv: requirements.txt
 	@echo "${GREEN}Setting up virtual environement...${RESET}";
 	@if [ ! -d $(INSTALL_DIRECTORY) ]; \
 	then \
 		virtualenv -p /usr/bin/python2  $(INSTALL_DIRECTORY); \
-		${PIP_CMD} install --upgrade pip==19.2.3 setuptools --index-url ${PYPI_URL} ; \
-		${PIP_CMD} install --upgrade pip==19.2.3 enum34==1.1.6 setuptools --index-url ${PYPI_URL} ; \
+		${PIP_CMD} install --upgrade pip==19.2.3 setuptools==44.0.0 --index-url ${PYPI_URL} ; \
```
procrastinatio: No need to install twice

Makefile
```diff
@@ -487,16 +487,16 @@ requirements.txt:
 ifeq ($(USE_PYTHON3), 1)
 .venv: requirements.txt
 		test -d "$(INSTALL_DIRECTORY)" || local/bin/python3.6 -m venv $(INSTALL_DIRECTORY); \
-		${PIP_CMD} install --upgrade pip==19.2.3 setuptools --index-url ${PYPI_URL} ; 
+		${PIP_CMD} install --upgrade pip==19.2.3 setuptools==44.0.0 --index-url ${PYPI_URL} ;
```
procrastinatio: This is specifically target for python3.6, I would leave as ease, as it is not used in the production line (for now)

procrastinatio:
OK thanks

PR #3361, opened 2020-01-15
setting a bod that serves diemo int 

ltkum:
For the diemo conference next week, we are going to use the integration diemo instance, so I'll deploy a diemo-int branch to serve it.


ltkum:
@AFoletti 
https://mf-chsdi3.int.bgdi.ch/ltkum_demo_diemo/shorten/86d56c5150 
if there is anything wrong, tell me ^^

ltkum:
we no longer need this.

PR #3366, opened 2020-01-17
CHSDI update ch.bafu.wasser-vorfluter

Simon1193:
[testlink](https://mf-chsdi3.dev.bgdi.ch/ltbei_wasser_vorfluter/shorten/86ed238715)


ltclm:
@procrastinatio  do we have to be concerned if some of the codebuild tests are not working?

procrastinatio:
The codebuild for python2 is missing some files in your branch. I did update it, should work now.

procrastinatio:
But for now we don't have any integration tests on codebuild.

chsdi/models/vector/bafu.py
```diff
@@ -1098,19 +1098,23 @@ class Gewaesserschutzbereiche (Base, Vector):
...
-    __tablename__ = 'vorfluter'
+    __tablename__ = 'effluent'
     __table_args__ = ({'schema': 'wasser', 'autoload': False})
     __bodId__ = 'ch.bafu.wasser-vorfluter'
     __template__ = 'templates/htmlpopup/vorfluter.mako'
-    __label__ = 'teilezgnr'
+    __label__ = 'id'
     id = Column('bgdi_id', Integer, primary_key=True)
-    teilezgnr = Column('teilezgnr', Integer)
+    id_2 = Column('id', Integer)
```
ltclm: why do you need to expose both ids bgdi_id and id (from the data , which is primary key). i would only expose the id which comes with the data

Simon1193: ID is an attribute from the table delivered by bafu and must apparear in the tooltip as specified in the data integration form.

However as they didn't specifiy a stabile id in the data integration form to be used, I created one with the table (the bgdi_id)

Simon1193: Ok I will use these id. thanks for the feedback

PR #3367, opened 2020-01-17
Allow polling on deploydev

procrastinatio:
For cronjobs, check for changes on GitHub and deploy to dev if necessary. A snapshot is also created.


ltkum:
This seems all right to me, but I'd get a second opinion for the Makefiles. @ltclm , does it look good to you too ?

PR #3389, opened 2020-02-12
Add parent constructor to CadastralWebMapParams

ltkum:
calling the parent's constructor to initialize variables, solving root cause for issue 3386

I did not check if other classes within makos could have the same problem yet.


ltkum:
After a small discussion with Marc, we need to have a much more deep cleansing of makos in general

PR #3395, opened 2020-02-17
updating doc regarding profile and height services for the next service-alti deploy

ltkum:
This refers to https://github.com/geoadmin/service-alti/pull/56


ltclm:
this should have been merged in the data branch i think. the master branch is for hotfixing and will be deployed at any time.

Issue #3394, opened 2020-02-16
layerDefs in Find/Identify can not handle non-ascii characters


eraymann:
Find-request returns 400 or 500 if layerDefs contains any non-ascii character.

e.g:

```
{
    "searchText": "contrôle",
    "searchField": "label",
    "layerDefs": {
        "ch.swisstopo.amtliches-strassenverzeichnis": "gdenr=371"
    }
}
```

works fine, whereas

```
{
    "searchText": 371,
    "searchField": "gdenr",
    "layerDefs": {
        "ch.swisstopo.amtliches-strassenverzeichnis": "label like '%contrôle%'"
    }
}
```

crashes with 400 or 500, depending on encoding/quoting.

Reproduction with py3:
```
import json
import urllib.parse

import requests

url = r"https://api3.geo.admin.ch/rest/services/api/MapServer/find"

# with special char in SeachText but not in LayerDefs, without quoting => 200
params_1 = {
    "layer": "ch.swisstopo.amtliches-strassenverzeichnis",
    "returnGeometry": "false",
    "searchText": "contrôle",
    "searchField": "label",
    "layerDefs": json.dumps({
        "ch.swisstopo.amtliches-strassenverzeichnis": "gdenr=371"
    })
}

# with special char in LayerDefs and unquoting => 500
params_2 = {
    "layer": "ch.swisstopo.amtliches-strassenverzeichnis",
    "returnGeometry": "false",
    "searchText": 371,
    "searchField": "gdenr",
    "layerDefs": urllib.parse.unquote(json.dumps({
        "ch.swisstopo.amtliches-strassenverzeichnis": "label like '%contrôle%'"
    }))
}

# with special char in LayerDefs and quoting => 400
params_3 = {
    "layer": "ch.swisstopo.amtliches-strassenverzeichnis",
    "returnGeometry": "false",
    "searchText": 371,
    "searchField": "gdenr",
    "layerDefs": urllib.parse.quote(json.dumps({
        "ch.swisstopo.amtliches-strassenverzeichnis": "label like '%contrôle%'"
    }))
}

for params in (params_1, params_2, params_3):
    response = requests.get(url=url, params=params)
    print(response.url)
    print(response.content)
```
If it's not a bug, please update your documentations with examples to handle special chars  
here https://api3.geo.admin.ch/services/sdiservices.html?highlight=layerdefs#correct-encoding  
[...]

PR #3400, opened 2020-02-24
Fix Issue 3394

ltkum:
This will sanitize inputs and remove the need for accents in searches while keeping it possible to do so. 
in Issue #3394 , 
case with param_2 work, but param_3 still does not. The issue is different in this case, as it seems we are not unquoting those parameters server side. 

The error message I receive back is : 
```
'{"status":"error","code":400,"detail":"Cannot parse \'layerDefs\' %7B%22ch.swisstopo.amtliches-strassenverzeichnis%22%3A%20%22label%20like%20%27%25Controle%25%27%22%7D"}'
```
and if you decode the url (for example with https://meyerweb.com/eric/tools/dencoder/), I can tell that this is a correct URL encoding. so, I'll be investigating this now.
- [x] add a test for queries with accents
- [x] investigate url encoding problem
- [x] give a testlink in the issue

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -59,4 +59,4 @@ WebOb==1.8.5
 WebTest==2.0.33
 zope.deprecation==4.4.0
 zope.interface==4.6.0
-unidecode
+unidecode==1.1.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -59,3 +59,4 @@ WebOb==1.8.5
 WebTest==2.0.33
 zope.deprecation==4.4.0
 zope.interface==4.6.0
+unidecode
--- a/chsdi/lib/helpers.py
+++ b/chsdi/lib/helpers.py
@@ -6,6 +6,7 @@ import requests
 import datetime
 import gzip
 import six
+import unidecode
 from decimal import Decimal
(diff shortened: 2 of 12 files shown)


ltkum:
@eraymann 
https://mf-chsdi3.dev.bgdi.ch/ltkum_dat_issue/ if you want to run some tests.

eraymann:
It's working now, thank you very much. One remark: It's a small disadvantage that accents are not included when searching with layerDefs. But that's not too bad, because they are considered in the conventional search with searchText. I can live with it.

requirements.txt
```diff
@@ -59,3 +59,4 @@ WebOb==1.8.5
 WebTest==2.0.33
 zope.deprecation==4.4.0
 zope.interface==4.6.0
+unidecode
```
procrastinatio: Please, pin the version (1.1.1)

procrastinatio: Nice

procrastinatio:
@eraymann
Deploying the last version
https://mf-chsdi3.dev.bgdi.ch/ltkum_dat_issue/

PR #3393, opened 2020-02-13
[POC] Caching some BOD queries

procrastinatio:

Issue #3282, opened 2019-10-03
Dynamic translation from DB


procrastinatio:
Do not use PO files anymore

PR #3401, opened 2020-02-25
translation class

ltkum:
Concern : Issue #3282 
Based upon the POC from @procrastinatio  in PR #3393 

For now, it is only a translation class. 

The max size of the cache for each language is 10'000, the current number of translations for each language is around 6000. 

I also created a simple method to empty the cache. It might not be used, but I tought it would be best to have one on hand if we want to quickly test if a translation is updated.

To Do List: 
- [x] apply translations to the whole code with this class rather than with the po files and the gettext method
- [x] ensure the cache is thread safe for writing
- [x] cleanup of the code still referring to the old translation method
- [x] cleanup of the translation scripts
- [x] cleanup of the translation files
- [x] cleanup of the makefile
- [x] codebuild (and jenkins) modifications 
Added by momo
- [x] cleanup old tests
- [x] added new tests
- [x] create a test link : [geoadmin test link](https://map.geo.admin.ch/?api_url=%2F%2Fmf-chsdi3.dev.bgdi.ch%2Fltkum_dynamic_translation&lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege&layers_visibility=false,false,false,false&layers_timestamp=18641231,,,), [api test link](http://mf-chsdi3.dev.bgdi.ch/ltkum_dynamic_translation/rest/services/translations?lang=fr)  
- [x] test this out


chsdi/lib/translator.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+class Translator:
+
+    _translations = None
+    _supported_languages = ['en', 'fr', 'de', 'rm', 'it']
+
+    @staticmethod
+    def get_db_connection():
+        return psycopg2.connect(host=os.environ.get('DBHOST'), dbname='bod_master', user='www-data')
```
ltclm: please use the env variable
https://github.com/geoadmin/mf-chsdi3/blob/ltkum_dynamic_translation/rc_dev#L7
for the database staging

ltkum: It has been done.

chsdi/lib/translator.py
```diff
@@ -0,0 +1,67 @@
...
+
+    _translations = None
+    _supported_languages = ['en', 'fr', 'de', 'rm', 'it']
+
+    """
+    Database transation methods
+    """
+    @staticmethod
+    def get_db_connection():
+        return psycopg2.connect(host=os.environ.get('DBHOST'), dbname=os.environ.get('DBSTAGING'), user='www-data')
```
ltclm: thank you but the db name is missing now:
```
'bod_{}'.format(os.environ.get('DBSTAGING'))
```

ltkum: ha yeah, didn't think of that. I correct that

procrastinatio:
So the idea is to get rid of the whole `gettext` business? nice.
But why aren't you using `SQLAlchemy`, but `psychopg2` directly?

ltkum:
Hey @procrastinatio , 
I based off the class from the translation script that was used to create the po files. ( long story short : Something was there and working, I'm using that ^^.) 
I could create a model and use sql alchemy directly.

procrastinatio:
`BeforeRender` and `NewRequest` are used for all sort of black magic. And `add_localizer` is responsible for locale negotiation if I'm not mistaken.

ltkum:
Yeah, I noticed for the black magic ^^. I removed the old translation renderer from subscribers (I FINALLY FOUND IT) and the translation factory. Me might get errors from that, but that would mean cleanup is not completely over

procrastinatio:
@ltkum 
You should have in your `.git/hooks/pre-commit`
```
#!/bin/bash                                                                                                                                                                              
scripts/pre-commit.sh
```

procrastinatio:
You cannot get your connection with env variables.
Put your translation class into something similar to `models.bod.Bod`

procrastinatio:
`lint` finally taming the Martin :-)

ltkum:
> `lint` finally taming the Martin :-)

Do not celebrate victory too soon ^^.

procrastinatio:
@ltkum Just in case, https://github.com/geoadmin/mf-chsdi3/blob/data/chsdi/views/translations.py and https://github.com/geoadmin/mf-chsdi3/blob/data/chsdi/models/bod.py#L491-L501

procrastinatio:
@ltkum @boecklic 

Having a more dynamic translation system is a great thing (because of automatic updates), but I'm still wondering if completely replacing the current system, by a completely new one is the best approach.

The current system is based on `babel` and `gettext`, which are file-based, which is not ideal for our need.
Instead of building something new, I think we may extend the current one and override the [`make_localizer`](https://github.com/Pylons/pyramid/blob/48cf45b0a1ac04e701101fa18778164825edd429/src/pyramid/i18n.py#L169-L174) (`pyramid.i18n.Localizer`), to use translation strings from the DB, either with a caching system or triggered by an API.
Or use the `merge` function of the `Translations` class (same file), to dynamically update the translations with the one from the BOD.

What's you take?

ltkum:
I don't know. I might not want to do an overload approach for one main reason : 
I already did most of the work replacing everything, and I, somehow, don't like to re-do the work.


I am not fond of overloading frameworks functions. It reduce codes readability (If I'm familiar with pyramid and see a "make_localizer" call, I'll expect it to be a standard make localizer call, not an overload.) and make the solution dependent on the used framework. Should we want to switch from pyramid to another framework, we would have another part of the system to redo entirely where a small adjustment might be needed if we separate them from the get go.  

I might be wrong, but I would keep my work as it has been done, and improve upon it.

ltkum:
**The following files have been modified during the merge**
I kept what was on master and used the new translation method where needed. Just to be on the safe side, can those who did the update test out if this still works as intended.

chsdi/templates/htmlpopup/gebietsauslaesse.mako
chsdi/templates/htmlpopup/teileinzugsbiete2.mako
chsdi/templates/htmlpopup/vorfluter.mako

chsdi/lib/translator.py
```diff
@@ -0,0 +1,87 @@
...
+
+    """
+    Database transation methods
+    """
+    @classmethod
+    def get_db_connection(cls):
+        host = cls._host if cls._host is not None else os.environ.get('DBHOST')
+        dbname = "bod_{}".format(cls._staging) if cls._staging is not None else \
+            "bod_{}".format(os.environ.get('DBSTAGING'))
+        return psycopg2.connect(host=host,
```
ltkum: Possible change : using sql alchemy instead of psychopg

chsdi/__init__.py
```diff
@@ -52,7 +53,7 @@ def main(global_config, **settings):
     config.include('pyramid_mako')
 
     # configure 'locale' dir as the translation dir for chsdi app
-    config.add_translation_dirs('chsdi:locale/')
+    # config.add_translation_dirs('chsdi:locale/')
```
ltkum: TODO : remove these

chsdi/lib/translator.py
```diff
@@ -0,0 +1,87 @@
...
+    @classmethod
+    def init(cls, db_host, db_staging, supported_languages):
+        cls._supported_languages = supported_languages.split()
+        cls._host = db_host
+        cls._staging = db_staging
+
+    @classmethod
+    def fill_translations(cls):
+        # using the Lock from threadings to ensure non concurrent writings
+        with cls._lock:
```
ltkum: TODO : error management here.

procrastinatio:
Superseeded by https://github.com/geoadmin/mf-chsdi3/pull/3509

PR #3402, opened 2020-02-25
remove double <i> tags in sphinxsearch response

ltclm:
The double ``<i>`` entries are due to missing translation of tlm categories.
The sphinxsearch service response ``rest/services/api/SearchServer`` are translated on the backend, only the labels.
With the latest tlm update we have new object categories ``Gebaeude`` without a translation.
I have added the missing translation to the bod and updated the po files of this branch:
```SQL
begin;
INSERT INTO translations (msg_id, de, fr, it, rm, en)
SELECT '<i>Gebaeude</i>', 'Gebaeude', fr, it, rm, en from translations WHERE bgdi_id = 3972;
rollback;
```
short term fix for **https://jira.swisstopo.ch/browse/BGDIINF_SB-982**
we should try to make the function more robust and avoid double tagging:
https://github.com/geoadmin/mf-chsdi3/blob/data/chsdi/views/search.py#L587

### DIFF
--- a/chsdi/locale/de/LC_MESSAGES/chsdi.po
+++ b/chsdi/locale/de/LC_MESSAGES/chsdi.po
@@ -4014,7 +4014,7 @@ msgid "ch.bafu.luftreinhaltung-stickstoff_kritischer_eintrag"
 msgstr "CLN Überschreitung"
 
 msgid "ch.bafu.mittlere-abfluesse"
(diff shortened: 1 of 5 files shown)


ltclm:
BGDIINF_SB-982

ltclm:
will be translated with releasenotes

PR #3409, opened 2020-03-05
[WIP] Add link to object to strassenverzeichnis tooltip

AFoletti:
This PR adds the "Link to object" element to the 
"Amtliches Verzeichnis der Strassen" tooltip
[Test](https://mf-chsdi3.dev.bgdi.ch/ltfoa/shorten/87d68ff3b2)


ltclm:
@AFoletti The "stable id" here is bgdi_id which is everything but not stable.
If we dont have a stable id provided by the datenherr, we cannot activate this feature.
In the streetnames data delivery there is no stable and unique attribute.

AFoletti:
@ltclm you are right.
But as per DH info we HAVE a stabil ID (esid).
It will suffice to modify the model to use that as primary key (see [model](https://github.com/geoadmin/mf-chsdi3/blob/7d722536ded11af60ad3f78ec99bfa1b7445650b/chsdi/models/vector/stopo.py#L3253) or is a fix in the DB needed too?

ltclm:
yes ambrogio, i have checked the data, 
the esid is not even unique its a simple attribute and not an id.
the problem is that there's sometimes a polygon and a line representation in the data with the same esid, thus two objects with the same id.

AFoletti:
OK. Returning the ball to the DH then. I'd say we put this in standby

PR #3410, opened 2020-03-05
Bugfix wmts cache settings

boecklic:
remove explicit `http_cache=0` on `wmtscapbilities`-route to allow varnish to cache WMTSCapabilites-file for 30min.


procrastinatio:
You also may cleanup the `wsgi.conf.ini` WMTS rewrite rules appears twice, once at about line 70 and also at the very end of the file.

PR #3414, opened 2020-03-06
Option to deactivate translation update

procrastinatio:
For now `make deploydev SNAPSHOT=true` force to update translation.
For code only deploy, it may not necessary to update these translation (translations are from the BOD, statyfied and then committed)

 `make deploydev SNAPSHOT=true UPDATE_TRANSLATION=false`

Or wait until https://github.com/geoadmin/mf-chsdi3/pull/3401 is finished, tested and merged

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ BODID ?=
 DEPLOY_TARGET ?=
 BRANCH_TO_DELETE ?=
 WMSSCALELEGEND ?=
+UPDATE_TRANSLATION ?= 'true'
 
 # Variables
 USER_SOURCE ?= rc_user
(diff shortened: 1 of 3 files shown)


ltkum:
I would wait until the PR is merged. We're going in a "code freeze" until the next data deploy, and we should be able to test thoroughly #3401 until then (I'll create a test link for that). That means that by the time we do the next "code deploy", we could start with the PR and then translations updates should be no more.

procrastinatio:
@ltkum OK, forgot about `data deploy`

procrastinatio:
Translation are now dynamic. So cloing without merging.

Issue #3449, opened 2020-04-06
Admin KML do not display any KML


procrastinatio:
Query returns nothing or `timestamp` are not stored.

https://github.com/geoadmin/mf-chsdi3/blob/data/chsdi/views/admin.py#L34-L37

procrastinatio:
OK fixed, thank you.

PR #3468, opened 2020-04-29
Issue 3449

ltkum:
in issue #3449 , we have a problem to display kmls in the admin page for them. 

After investigation, it seems the root of the issue is that we wrote the wrong bucket names in dynamodb as we save kmls.

### DIFF
--- a/chsdi/lib/files_handler.py
+++ b/chsdi/lib/files_handler.py
@@ -26,10 +26,10 @@ class DynamoDBFilesHandler:
         try:
             self.table.put_item(
                 Item={
-                    'adminId': admin_id,
-                    'fileId': file_id,
-                    'timestamp': timestamp,
-                    'bucket': self.bucket_name
+                    'adminId': {'S': admin_id},
+                    'fileId': {'S': file_id},
+                    'timestamp': {'S': timestamp},
+                    'bucket': {'S': self.bucket_name}
                 }
             )
         except Exception as e:
@@ -39,7 +39,10 @@ class DynamoDBFilesHandler:
         item = None
         try:
             item = self.table.get_item(Key={'adminId': str(admin_id)}).get('Item', None)
-        except Exception:
+        except Exception as e:
+            log.error("----------------------------------------")
+            log.error("...----------------------------------...")
+            log.error(e)
             pass
         return item
 
--- a/chsdi/lib/files_handler.py
+++ b/chsdi/lib/files_handler.py
@@ -26,10 +26,10 @@ class DynamoDBFilesHandler:
         try:
             self.table.put_item(
                 Item={
-                    'adminId': {'S': admin_id},
-                    'fileId': {'S': file_id},
-                    'timestamp': {'S': timestamp},
-                    'bucket': {'S': self.bucket_name}
+                    'adminId': admin_id,
+                    'fileId': file_id,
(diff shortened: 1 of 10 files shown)


chsdi/lib/files_handler.py
```diff
@@ -121,8 +123,7 @@ class FilesHandler(object):
     def __init__(self, request):
         self.request = request
         # Set up AWS DynamoDB and S3 handlers
-        self.dynamodb_fileshandler = DynamoDBFilesHandler(
-            self.dynamodb_table_name, self.bucket_key_name, self.region)
+        self.dynamodb_fileshandler = DynamoDBFilesHandler(self.dynamodb_table_name, self.bucket_name, self.region)
```
procrastinatio, line 126: Do we use this `bucket_key_name` anywhere?

chsdi/views/admin.py
```diff
@@ -36,17 +42,27 @@ def kml_load(api_url='//api3.geo.admin.ch', bucket_name=None, table_name=None, r
...
+    logging.debug("----------------------- ! ! ! ---------------------")
     for f in results['Items']:
         try:
             resp = requests.head("http:" + api_url + "/files/" + f['fileId'], headers={'User-Agent': 'mf-geoadmin/python'})
             if int(resp.status_code) == 200:
                 fileids.append((f['fileId'], f['adminId'], f['timestamp']))
-        except RequestException:
-            pass
+        except RequestException as e:
+            logging.error(e)
```
procrastinatio, line 65: Nice

chsdi/lib/files_handler.py
```diff
@@ -39,7 +39,10 @@ def get_item(self, admin_id):
         item = None
         try:
             item = self.table.get_item(Key={'adminId': str(admin_id)}).get('Item', None)
-        except Exception:
+        except Exception as e:
+            log.error("----------------------------------------")
```
procrastinatio, line 43: Cleanup?

chsdi/views/admin.py
```diff
@@ -18,6 +19,11 @@ def admin_kml(context, request):
     table_name = settings.get('geoadmin_file_storage_table_name')
     table_region = settings.get('geoadmin_file_storage_table_region')
     api_url = settings.get('api_url')
+    logging.debug("-----------------------------------------------------------------------")
+    logging.debug(bucket_name)
```
procrastinatio, line 23: Not sure if we leave this. Maybe add the name of the variable whose value is printed

procrastinatio:
Rebase you branch on `master`. The tests will pass (hopefully)

chsdi/lib/files_handler.py
ltkum: From what I've seen, not anymore.

chsdi/lib/files_handler.py
ltkum: It's still a work in progress (I should provide a test link), I'm keeping some logs here and there. But I'll cleanup most of the logs before going live

chsdi/views/admin.py
ltkum: In the long run, no real need to keep most of the debug logs in my opinion.

procrastinatio:
Looks OK. Ich teste noch.

procrastinatio:
Oops, wrong language.

PR #3498, opened 2020-06-17
add simplified geometry as highlight

Simon1193:


Simon1193:
[testlink](https://mf-chsdi3.dev.bgdi.ch/ltbei_geologische3d_modelle/shorten/89ed5f243a)

PR #3509, opened 2020-06-29
Dynamic translation from BOD

procrastinatio:
Dead simple approach to the dynamic translation issue. Retains the whole and complexe pyramid `i18n` mechanism, updating only from time to time the translation from the BOD.
The `.po/.mo` files are used as a backup.

BOD requests are cached for one hour.

Test with

http://mf-chsdi3.dev.bgdi.ch/dynamic_translation/testi18n?lang=en

The string ''ch.astra.mountainbikeland-sperrungen_umleitungen'' is not in the `.po` file of `master`, but in the BOD. 

or

http://map.geo.admin.ch/?api_url=//mf-chsdi3.dev.bgdi.ch/dynamic_translation

Some very unscientific performance tests:

1/ A thousand requests, on non-cached localhost (best case)
With po file: 13.92+/- 0.05
With this PR  14.31+/-0.13

2/ Testing a popup on 104 layers (worst case)
with po file:  35.47+/-1.03 
With this PR:  36.35+/-0.66

According to these two tests, the overhead ranges between 2.5 and 2.8% for the dynamic translation

See https://github.com/geoadmin/mf-chsdi3/issues/3282


ltclm:
@procrastinatio this will be deployed separately right?

procrastinatio:
Yes, this may

procrastinatio:
Well, `pyramid` still need this `.po` files, but they don't need to be up to date. So, either
* Removing them completely and generate them on the fly, when deploying.
* Empty files
* Update only when deploying (still in GitHub)
Up to you!

ltclm:
i propose to remove them completely from .git and generate them on the fly on the deploy target, with bod_<target> as datasource.

PR #3514, opened 2020-07-09
Updated legends

faselm:


faselm:
@rebert OK, I reran the make legends command with the WMSSCALELEGEND=5000 parameter

PR #3515, opened 2020-07-10
 add text in case attribute value is empty

Simon1193:
[testlink](https://mf-chsdi3.dev.bgdi.ch/ltbei_erneuerbarheizen/shorten/8a63691c3e)


Simon1193:
As askey by bfe, the empty value for attribute "consultingcost" are replaced by text "on request" in the four languages

PR #3519, opened 2020-07-17
remove oids from model

ltclm:
get rid of oids in data models


ltclm:
the oids in these two table are manually created and not internal oids. we can keep them in the table and dont have to change the model.

PR #3522, opened 2020-07-26
Service to replace the local access script in service-proxywms

procrastinatio:
Test: http://mf-chsdi3.dev.bgdi.ch/feature_wmsconfig/rest/services/wmsconfig

To replace: https://github.com/geoadmin/service-proxywms/blob/master/scripts/import-restrictions.py

### DIFF
--- a/chsdi/__init__.py
+++ b/chsdi/__init__.py
@@ -100,6 +100,7 @@ def main(global_config, **settings):
     config.add_route('backend_checker', '/backend_checker')
     config.add_route('downloadkml', '/downloadkml')
     config.add_route('translations', '/rest/services/translations')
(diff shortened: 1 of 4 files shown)


ltclm:
@procrastinatio @boecklic 
didnt we decide (years ago) to move the wmts config and wmts getcap from chsdi3 to service-proxywms?
This pr is doing the opposite in my eyes.
And i think service endpoint should be named ``rest/services/wmtsconfig`` because it is not about wms config.

ltclm:
this config file has been introduced to separate code deploy from config deploy, now we will have to touch and deploy chsdi if we want to change the configuration of service-proxywms

PR #3528, opened 2020-07-30
Replacing mako by envsubst

procrastinatio:
* Currently we have a very complex set of variables, scripts and makefile, with everything possible : from SCREAMING_SNAKE_CASE, to snake_case, kebap-case to set camelCase in code.
* Using `envsubst` for `Dockerfile`


procrastinatio:
Basically done in https://github.com/geoadmin/mf-chsdi3/pull/3725

PR #3529, opened 2020-07-30
Updated legends

faselm:


rebert:
After @mariokeusen this layer don't need a legend - and this here is crap

PR #3544, opened 2020-09-10
Expose some attributes from third party source (identify)

procrastinatio:
A purely constructed attribute *PDF* is added to the model _OerebkatasterZoom1(_  (https://github.com/geoadmin/mf-chsdi3/blob/data/chsdi/models/vector/stopo.py#L2402-L2410). The link is read from a request XML documents somewhere from a canton service (yes, via http request)

It basically exposes an attribute which does not exist in the DB table into the **identify** JSON. As it uses the _geometry_ to compute it's location, it won't work if the parameter **returnGeometry=false** is set

This PR should not exists in the first place, as it opens the door for any excess

**identify** (with this PR)
https://mf-chsdi3.int.bgdi.ch/feature_BGDIINF_SB-1301/rest/services/all/MapServer/identify?geometry=2601796.375,1200283.125&geometryFormat=geojson&geometryType=esriGeometryPoint&imageDisplay=1059,907,96&lang=en&layers=all:ch.swisstopo-vd.stand-oerebkataster&limit=10&mapExtent=2601618.350028069,1200166.769096322,2601883.100028069,1200393.519096322&returnGeometry=true&sr=2056&tolerance=10

without PR
https://mf-chsdi3.int.bgdi.ch/rest/services/all/MapServer/identify?geometry=2601796.375,1200283.125&geometryFormat=geojson&geometryType=esriGeometryPoint&imageDisplay=1059,907,96&lang=en&layers=all:ch.swisstopo-vd.stand-oerebkataster&limit=10&mapExtent=2601618.350028069,1200166.769096322,2601883.100028069,1200393.519096322&returnGeometry=true&sr=2056&tolerance=10

**htmlPopup**
https://mf-chsdi3.int.bgdi.ch/rest/services/ech/MapServer/ch.swisstopo-vd.stand-oerebkataster/845825/htmlPopup?coord=2601796.375,1200283.125&imageDisplay=1059,907,96&lang=en&mapExtent=2601618.350028069,1200166.769096322,2601883.100028069,1200393.519096322&sr=2056

See https://jira.swisstopo.ch/browse/BGDIINF_SB-1301 for rational

### DIFF
--- a/chsdi/models/vector/__init__.py
+++ b/chsdi/models/vector/__init__.py
@@ -12,6 +12,7 @@ from shapely.geometry import box
 from sqlalchemy.sql import func
 from sqlalchemy.orm.util import class_mapper
 from sqlalchemy.orm.properties import ColumnProperty
+from sqlalchemy.ext.hybrid import hybrid_property
 from geoalchemy2.elements import WKBElement
 from geoalchemy2.shape import to_shape
 
@@ -81,6 +82,16 @@ class Vector(object):
         geom = None
         bbox = None
         properties = {}
+        hybrids = dict(
+            (key, prop)
+            for key, prop in class_mapper(self.__class__).all_orm_descriptors.items()
+            if isinstance(prop, (ColumnProperty, hybrid_property))
+        )
+        # Add hybrid attributes (see https://jira.swisstopo.ch/browse/BGDIINF_SB-1301)
+        # These attributes are "constructed" and do not have columns in the DB
+        for i in hybrids.keys():
+            if not i.startswith('_'):
+                properties[i] = getattr(self, i)
         for p in class_mapper(self.__class__).iterate_properties:
             if isinstance(p, ColumnProperty):
                 if len(p.columns) != 1:  # pragma: no cover
@@ -105,6 +116,7 @@ class Vector(object):
                 elif (not col.foreign_keys
                       and not isinstance(col.type, GeometryChsdi)):
                     properties[p.key] = val
+
         properties = self.insert_label(properties)
         return id, geom, properties, bbox
 
--- a/chsdi/models/vector/stopo.py
+++ b/chsdi/models/vector/stopo.py
@@ -2428,22 +2428,22 @@ class OerebkatasterZoom1(Base, Oerebkataster, Vector):
         return "{}{}{},{}".format(self.oereb_webservice, path_xml, center.x, center.y)
 
     @hybrid_property
-    def egrids(self):
+    def _egrids(self):
         list_egrid = []
         try:
-            response = requests.get(self._oereb_xml_url)
+            response = requests.get(self._oereb_xml_url, timeout=5)
             if response.status_code == 200:
                 root = et.fromstring(response.text)
                 list_egrid = root.findall('{http://schemas.geo.admin.ch/V_D/OeREB/1.0/Extract}egrid')
         except BaseException:
-            pass
+            return []
         return [e.text for e in list_egrid]
 
     @hybrid_property
     def pdfs(self):
         if self._is_oereb_webservice:
             path_pdf = sanitize_url("{}/extract/reduced/pdf/".format(self.oereb_webservice))
-            return ['{0}{1}'.format(path_pdf, i) for i in self.egrids]
+            return ['{0}{1}'.format(path_pdf, i) for i in self._egrids]
         return []
 
 
--- a/chsdi/models/vector/stopo.py
+++ b/chsdi/models/vector/stopo.py
@@ -1,12 +1,18 @@
 # -*- coding: utf-8 -*-
 
+import requests
+import xml.etree.ElementTree as et
+
 from sqlalchemy import Column
 
 from sqlalchemy.types import Numeric, Boolean, Integer, Float, Unicode, BigInteger, SmallInteger
+from sqlalchemy.ext.hybrid import hybrid_property
+from geoalchemy2.shape import to_shape
 
 from chsdi.models import register, register_perimeter, bases
 from chsdi.models.types import DateTimeChsdi
 from chsdi.models.vector import Vector, Geometry2D
+from chsdi.lib.helpers import sanitize_url
 
 
 Base = bases['stopo']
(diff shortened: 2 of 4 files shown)


ltclm:
this is a really nice solution, thank you @procrastinatio.
But i think we should not start to customize the identify as we did in the htmlpopup service already (easy-boarding).
would it be possible to generate this pdf links with fme and store them in the database?

rebert:
@procrastinatio I agree with @ltclm  .... even the tooltip has been too much adaption toward the client... taking it to the service would make one (important) client happy, but for maintanance of chsdi it is a bad idea to add functionality...

rebert:
@hansmannj You have to decide if in this deploy or not

ltclm:
select by rectangle is timing out:
https://mf-chsdi3.dev.bgdi.ch/ltclm/shorten/8bb747ff26
image

hansmannj:
Dear @procrastinatio 

Thanks a lot for your effort here. And @ltclm  and @rebert  thanks for your comments.
I am on it and hope we'll have a decission on this today before lunchtime.

One question: is the timout issue related to the changes made by this PR?
I could not reproduce the timeout with rectangular selection and querry using the link you provided @ltclm.
Seems to work fine currently, even with larger rectangles and hence more objects

ltclm:
@hansmannj  please use the api endpoint from this pr. i have adapted the link:
https://mf-chsdi3.dev.bgdi.ch/feature_BGDIINF_SB-1301/shorten/8bbba157b6

ltclm:
we could offer a weekly update of this information from the cantonal services and add it to the automated data update in the database as a new table column. 

The cantonal XML services will be read once a week during the ordinary data update with FME.
with this data based solution we can:
* get rid of the htmlpopup/mako magic
* use a standard db column (identify=htmlpopup)
* one exception less

p1d1d1:
> 
> 
> we could offer a weekly update of this information from the cantonal services and add it to the automated data update in the database as a new table column.
> 
> The cantonal XML services will be read once a week during the ordinary data update with FME.
> with this data based solution we can:
> 
>     * get rid of the htmlpopup/mako magic
> 
>     * use a standard db column (identify=htmlpopup)
> 
>     * one exception less

@hansmannj  Don't you think that so a change should first be communicated and discussed with customer?

ltclm:
@p1d1d1 you're right, we should never have accepted this. that's why we're trying to move this information to the data level now.

hansmannj:
> we could offer a weekly update of this information from the cantonal services and add it to the automated data update in the database as a new table column.
> The cantonal XML services will be read once a week during the ordinary data update with FME.
> with this data based solution we can:
> 
> get rid of the htmlpopup/mako magic
> use a standard db column (identify=htmlpopup)
> one exception less



> we could offer a weekly update of this information from the cantonal services and add it to the automated data update in the database as a new table column.
> The cantonal XML services will be read once a week during the ordinary data update with FME.
> with this data based solution we can:
> * get rid of the htmlpopup/mako magic
> 
> * use a standard db column (identify=htmlpopup)
> 
> * one exception less
> 
> 
> @hansmannj  Don't you think that so a change should first be communicated and discussed with customer?

This is a nice solution @ltclm! What I like especially is, that we would not need to add customisations/exceptions to our existing services! Lets discuss that in the daily today with @procrastinatio, @rebert and @pauloamado.

@p1d1d1: This change is driven by a customers request (https://jira.swisstopo.ch/browse/BGDIINF_SB-1301), so the customer is already informed. @pauloamado is in contact with them and I'll check what he thinks about this solution.

hansmannj:
> @p1d1d1 you're right, we should never have accepted this. that's why we're trying to move this information to the data level now.

Yes, the data level is the best place for that.

p1d1d1:
> @p1d1d1: This change is driven by a customers request (https://jira.swisstopo.ch/browse/BGDIINF_SB-1301), so the customer is already informed. @pauloamado is in contact with them and I'll check what he thinks about this solution.

@hansmannj this is not 100% "same same". The JIRA ticket is about having the PDF link in the "identify" service. To generate this link, a previous call to cantonal services is required in order to get the EGRID, which is part of the PDF URL.

@ltclm proposes here to get rid of this call by reading the services once a week and store the info in the DB. I'm 100% fine with this, but the customer is not informed on this specific change, which is not related to the JIRA ticket. e.g. is a weekly update enough? Maybe not since some cantons have daily updates?

hansmannj:
> > @p1d1d1: This change is driven by a customers request (https://jira.swisstopo.ch/browse/BGDIINF_SB-1301), so the customer is already informed. @pauloamado is in contact with them and I'll check what he thinks about this solution.
> 
> @hansmannj this is not 100% "same same". The JIRA ticket is about having the PDF link in the "identify" service. To generate this link, a previous call to cantonal services is required in order to get the EGRID, which is part of the PDF URL.
> 
> @ltclm proposes here to get rid of this call by reading the services once a week and store the info in the DB. I'm 100% fine with this, but the customer is not informed on this specific change, which is not related to the JIRA ticket. e.g. is a weekly update enough? Maybe not since some cantons have daily updates?

@p1d1d1 you are right and that's a valid point. We have just discussed this in our daily and @pauloamado will check with the customer if the suggested solution is OK. We hope to have a decision on this issue by this afternoon. 
And again thanks to all for the effort and discussions done for this issue.

p1d1d1:
> [Update]: OK, thanks that you, @p1d1d1, will check with the customer if the solution is ok 😄

@hansmannj 
Just had a phone call with the customer. A weekly update is for them a "no go", instead they are OK with a daily update.
Just sent additionally an email to get a "traceable" confirmation.

ltclm:
@p1d1d1 
**the current situation is :**
* weekly data in feature geometry and attributes
* live data in htmlpopup
* customized service htmlpopup

**the situation with the new pr is:**
* weekly data in feature geometry and attributes
* live data in identify
* customized service identify

Thats why we have to find a data based solution:
* Both solutions are not consistent because we are combining live data with weekly data. 
* everything which is published in our services is part of the data delivery and has the same timestamp
* the customized services (like the two htmlpopup and identify examples here) wont be supported/operated in the future "self-service" bgdi anymore (whenever that will be) like it was in the past years. we should slowly start to go into that direction and standardize our services. so these two options are like a "no go" from the point of view "future bgdi services".


We will have to clarify with @eraymann if a daily pull and aggregation of the interlis data is possible first.

hansmannj:
Thanks @ltclm, this is a good explanation of the current situation and the options.
The weekly updates would avoid the inconsistencies that exist both in the current situation and in the PR's solution.

I know the standardization of our services will cause pain and inconveniences first, but it will pay out on the mid- to long-term.
As @ltclm said, I agree it makes sense to start going into the direction of the "future bgdi services" the sooner the better.
Every new exception and customization we introduce to our services will cause more pain and inconveniences later.

gjn:
The current situation as @ltclm describes is one of the reasons we face challenges in data updates and maintenance.

Streamlining is required and I agree it should happen in the data level. Then it's stream lined and there are no inconsistencies. So we should go the way @ltclm proposed (solve it at the data level). 

If weekly updates is not enough and more frequent updates are needed, then we need to address this in a seperate issue (initiative/epic), as more people will be involved.

ltclm:
@p1d1d1 
the weekly published data from mopublic has an egrid attribute which could be used for this layer.
you could ask @eraymann if its possible to harvest the interlis data each day and generate / provide the fgdb that we are using for the web publication.

p1d1d1:
@ltclm just had a meeting and agreed to proceed as you suggest! Additionally, in case of missing EGRID numbers, @eraymann will retrieve them and store in the mopublic DB. On the IGEB-B side, there will also be a daily update/post processing of the data to create and store the PDF link.

boecklic:
I will close this PR (see: https://jira.swisstopo.ch/browse/BGDIINF_SB-1301)

PR #3548, opened 2020-09-15
revert sedex gwr

ltclm:
Reverts geoadmin/mf-chsdi3#3531
we will have to revert this pr once the final decision has been taken.


rebert (commented):
what is the purpose of sitemaps.py?

ltclm:
thats an endpoint which is decomissioned for ages already. it has been decided to remove from the code.

PR #3559, opened 2020-10-12
Ltmbr add seilbahnen bundeskonzession

marionb:
Test link:
https://mf-chsdi3.dev.bgdi.ch/ltmbr_add_seilbahnen_bundeskonzession/shorten/8c48b773fe


chsdi/templates/htmlpopup/seilbahnenbundeskonzession.mako
```diff
@@ -0,0 +1,18 @@
...
+</%def>
+
+<%def name="betreiber()">
+  <%
+  layer = 'ch.bav.seilbahnen-bundeskonzession.'
+  %>
+  <tr><td class="cell-left">${_(layer + 'betreiber_tuabkuerzung')}</td> <td>${c['attributes']['betreiber_tuabkuerzung'] or '-'}</td></tr>
+</%def>
+
+<%def name="layer(translation)">
```
b4l, line 16: In general it's a bad practice to name an attribute and a function the same in python as it gets overloaded.

chsdi/templates/htmlpopup/seilbahnenbundeskonzession.mako
```diff
@@ -0,0 +1,18 @@
+<%def name="anlage()">
+  <%
+  layer = 'ch.bav.seilbahnen-bundeskonzession.'
```
b4l, line 3: Nice touch! Still, maybe just put it inline as it barely helps to increase readability or maintainability given there are only two instances or even only one further down.

b4l (commented):
Looks good! Left some minor remarks. I like the translation mako-magic. I was wondering whether it would be possible to expose this and the layer name/bodid through the base mako.

chsdi/templates/htmlpopup/seilbahnenbundeskonzession.mako
marionb: Thanks for the hint. Actually I did not want to use the variable any more, just the function.  I just forgot to change the code of this make. Good catch!

marionb:
@b4l 
> I was wondering whether it would be possible to expose this and the layer name/bodid through the base mako.
I thought about that as well. It would be very handy (potentially and save some debugging time). I'll have a look but not in this PR.

faselm (approved):

b4l:
@marionb did you redeploy and test your latest changes? Calling a function defined below does generally not work well in python, don't know about how it is handled in mako though.

PR #3577, opened 2020-11-12
Ltfas ch.swisstopo.geologie-gesteinsdichte

faselm:
Test link: https://mf-chsdi3.dev.bgdi.ch/ltfas_gesteinsdichte/shorten/8ced533c16


marionb:
There is a lot of attributes that you are defining in the tool-tip. Perhaps you should check if you can add some in an extended tool tip. I think the number of attributes in normal tooltip is max 8 (what is noted in the DIF). Perhaps also check with @mariokeusen.

faselm:
> There is a lot of attributes that you are defining in the tool-tip. Perhaps you should check if you can add some in an extended tool tip. I think the number of attributes in normal tooltip is max 8 (what is noted in the DIF). Perhaps also check with @mariokeusen.

Thanks for suggesting but it's already discussed and it's OK like that ;)

b4l (approved):

PR #3584, opened 2020-11-24
Ltmbr gletscherausdehnung BGDIDIC-1924

marionb:
test link: https://mf-chsdi3.dev.bgdi.ch/ltmbr_gletscherausdehnung_BGDIDIC-1924/shorten/8d26742d66


chsdi/templates/htmlpopup/gletscherausdehnung.mako
```diff
@@ -0,0 +1,10 @@
+<%inherit file="base.mako"/>
+
+<%def name="table_body(c,lang)">
+<%
+  layer = 'ch.swisstopo.geologie-gletscherausdehnung.'
+%>
+<tr><td class="cell-left">${_(layer + 'name')}</td> <td>${c['attributes']['name'] or '-'}</td></tr>
+<tr><td class="cell-left">${_(layer + 'sgi-id')}</td> <td>${c['attributes']['sgi_id'] or '-'}</td></tr>
+<tr><td class="cell-left">${_(layer + 'area_km2')}</td> <td>${c['attributes']['area_km2'] or '-'}</td></tr>
```
b4l, line 9: It would look nicer when rounded to a few digits. It's probably the customers decision though.

chsdi/models/vector/stopo.py
```diff
@@ -3412,3 +3412,18 @@ class GeologieFelslabore(Base, Vector):
     the_geom = Column(Geometry2D)
 
 register('ch.swisstopo.geologie-felslabore', GeologieFelslabore)
+
+
+class Gletscherausdehnung(Base, Vector):
+    __tablename__ = 'view_gletscherausdehnung_sgi_2016_glaciers'
```
b4l, line 3418: Maybe use a more generic name for future proving.

ltclm: agree with @b4l we should use a generic name here like ``view_gletscherausdehnung`` or ``view_gletscherausdehnung_tooltip``. it this layer will be updated one day, it will either replace the current data or create a time-enabled layer. We will have to find a generic / attribute / content based solution for the time-series then.

ltclm (changes requested):
technical implementation is all good, thank you.
Same concern here about precision of data area_km2. They should be rounded to a reasonable value. in identify and htmlpopup i would say. so you have to round it either in the fme workbench or in the view. but please ask the data owner about that.

b4l:
> Same concern here about precision of data area_km2. They should be rounded to a reasonable value. in identify and htmlpopup i would say. so you have to round it either in the fme workbench or in the view. but please ask the data owner about that.

I think we should preserve the data from the delivery as accurate as possible. Therefore manipulating precision in the workbench or identify seems wrong to me.
For the tooltip we may want to consider rounding floating point to 1 or 2 digits after period as the default.

ltclm:
> > Same concern here about precision of data area_km2. They should be rounded to a reasonable value. in identify and htmlpopup i would say. so you have to round it either in the fme workbench or in the view. but please ask the data owner about that.
> 
> I think we should preserve the data from the delivery as accurate as possible. Therefore manipulating precision in the workbench or identify seems wrong to me.
> For the tooltip we may want to consider rounding floating point to 1 or 2 digits after period as the default.

yes, agree, the data owner has to define that and deliver the data with the right precision.

chsdi/models/vector/stopo.py
marionb: @ltclm do you think is it enough if I change the name of the view for the moment. Or should I try to prepare something more for a potential update to make that as easy as possible in the future? For me the problem is a bit that we do not know exactly how the update will look.

marionb:
I'll check with @consta-v about the data precision in the delivery. For the time being I can round the data in the tool-tip to 2 digit. If they give the data in km2 I would expect the precision to be around that as well (but that is just a guess in to the blue).

chsdi/models/vector/stopo.py
ltclm: yes, thats enough for the moment, thank you!

marionb:
@b4l and @ltclm Apparently two digits are sufficient. I updated the tooltip.

b4l (approved):

faselm (approved):

chsdi/templates/htmlpopup/gletscherausdehnung.mako
```diff
@@ -0,0 +1,11 @@
+<%inherit file="base.mako"/>
+
+<%def name="table_body(c,lang)">
+<%
+  layer = 'ch.swisstopo.geologie-gletscherausdehnung.'
```
ltclm, line 5: you can use `c['layerBodId']` for the layername

ltclm (changes requested):
just a small change needed.
thank you!

ltclm (approved):

PR #3586, opened 2020-11-26
BGDIDIC-1933: Add ch.bfe.shared-mobility

b4l:
Test link: https://mf-chsdi3.dev.bgdi.ch/feature_BGDIDIC-1933/shorten/8d0ba22732


faselm (commented):
I assume it's normal all the tooltips and descriptions are in German. Layer, tooltip and legends look good. Something I notice is that, at low scale when clicking around many points, only one point is selected and not particularly the one closest to where I click. But I guess it's OK. Just wonder from what it comes from...

ltclm (approved):
looks good to me!
Only question: what are we supposed to show as "datenstand" in the legend?
https://mf-chsdi3.dev.bgdi.ch/feature_BGDIDIC-1933/rest/services/all/MapServer/ch.bfe.shared-mobility/legend?lang=de
The data is updated every 5 minutes.

b4l:
> Only question: what are we supposed to show as "datenstand" in the legend?

@ltclm i don't know, probably similar to https://s.geo.admin.ch/8c57f922d1
Can we handle this in the cron job?

cc @mariokeusen

ltclm:
we can offer an update of this metadata every 30m. you would have to add one of the json documents here:
https://github.com/geoadmin/bgdi-scripts/blob/master/geodata_cron_skripte/bgdipg01t/data_bod_timestamps.sh.
@mariokeusen what was the deal here with the customer?

PR #3604, opened 2021-01-14
BGDIDIC-1817: Update ch.bfe.energieforschung

b4l:
Test link: https://mf-chsdi3.dev.bgdi.ch/feature_BGDIDIC-1817/shorten/8e2c380785

Mapfile: https://github.com/geoadmin/wms-mapfile_include/pull/602

Script: https://github.com/geoadmin/bgdi-scripts/pull/798

Sphinx: https://github.com/geoadmin/service-sphinxsearch/pull/449


marionb:
@b4l

faselm:
> @b4l
> When I look at the test link there is something strange with the layer, It looks as though there is an overlapping tile or something:
> tile-cut50km
> It appears at a zoom level of 50km. When zooming in the layer looks normal. It appears on both firefox and chromium so I doubt that it is a browser issue.
> Is this intentional?

In my browser too, but I think it's because the tiles have not been generated on proxywms yet?

ltclm:
this is a wms layer. if you are using wms-c (tiled wms) as  datasource, you have to be sure that theres a gutter set in the table layers_js.
otherwise the symbols and labels can be truncated at the tile boundary.

AFoletti:
As @ltclm says. 😃 

And if I may add, could you please move the symbol a little up so that the geometry is near the tip of it? See others BFE layers as example.

**Wrong**:
energieforshung

**Right** (from [ch.bfe.erneuerbarheizen](https://s.geo.admin.ch/8e2e1d6d38)):
heizung

Thanks in advance!

ltclm (changes requested):
please fix the anchor point as described by @AFoletti and add a gutter.
besides that, looks good to me, thank you!

b4l:
@all thanks for the feedback, the fixes are in place. please have a quick look whether the issues are resolved. thanks!

ltclm (approved):
thank you!

marionb (approved):
Looks good now. Thanks!

AFoletti:
Thanks for the fixes!

PR #3610, opened 2021-01-27
BGDIDIC-1989: Add ch.vbs.panzerverschiebungsrouten

b4l:
Test link: https://mf-chsdi3.dev.bgdi.ch/feature_BGDIDIC-1989/shorten/8e6fdc8a07

Script: https://github.com/geoadmin/bgdi-scripts/pull/806

WMS: https://github.com/geoadmin/wms-mapfile_include/pull/605


faselm (commented):
Symbology, legends, tooltips look good!

marionb (approved):
LGTM, Thanks!

ltclm (approved):
lgtm but i would like to know why there is only a tooltip on the point features, its a bit confusing for an enduser in my opinion.

faselm:
> lgtm but i would like to know why there is only a tooltip on the point features, its a bit confusing for an enduser in my opinion.

I actually also desperately tried to get info from the line :) I agree, it's a bit confusing...

b4l:
@ltclm this is specified so in the DIF, also there is no such information for the lines. Maybe @mariokeusen has some more insights from the tank drivers and owners.

PR #3618, opened 2021-02-05
Adding support for WebMercator on feature metadata endpoint with grid layers

pakb:
Trying to get the same output for grid layers (S3 tiles) as for layers that live in the DB
meaning both with :
```json
{
    "feature": {
         "metadata": "here..."
    }
}
```


procrastinatio (approved):
Nice.

PR #3621, opened 2021-02-09
No need to wrap each feature in a feature entry when going through identify

pakb:
Small regression luckily caught soon after last merge...


procrastinatio (approved):
Nice catch.
If `Content-Type`is `'application/geo+json'`i the result is effectively a real `GeoJSON`

PR #3624, opened 2021-02-11
Update releasenote for Feb 24th with removed layers and layers to be removed in near future

mariokeusen:
Please doublecheck and merge.


consta-v:
Changes OK.

PR #3628, opened 2021-02-12
Create index.rst

davidoesch:
updated Terms of use link
removed ChargeableLayers list

@procrastinatio 
can you double check if now all ChargeableLayers appearing in the notChargeableLayers list?


procrastinatio (approved):
OK

PR #3643, opened 2021-02-22
Feature bgdiinf sb 1527 fix search

boecklic:


davidoesch:
Thx!

chsdi/views/search.py
```diff
@@ -205,19 +204,35 @@ def _swiss_search(self):
...
+                # we have results from both queries (exact + wildcard)
+                # prepend exact search results to wildcard search result
+                exact_results = temp[1].get('matches', [])
+                # exact matches have priority over prefix matches
+                # searchText=waldhofstrasse+1
+                # waldhofstrasse 1 -> weight 100
+                # waldhofstrasse 1.1 -> weight 1
+                for result in exact_results:
+                    if result['attrs']['detail'].startswith('%s ' % (' '.join(self.searchText))) or \
+                       result['attrs']['detail'] == ' '.join(self.searchText):
```
gjn, line 220: I think this could be improved by:
1) lowercase the searchText before doing the comparison (as @detail is all lowercase)
2) also look for exact matches _inside_ of detail

We still miss the umlauts with this appraoch (searchText 'Zürich' is 'zuereich' in the @detail field of sphinx).

Isn't there an Sphinx Weighting method that takes all this into account out of the box?

ltclm: thank you for the review and  feedback @gjn 
we will try to address these points in the new testlink / pr.

PR #3646, opened 2021-03-02
removing the JS API

procrastinatio:
Removing the JS API

[Demo](https://mf-chsdi3.int.bgdi.ch/remove_js_api3/)

@pakb Nice JS embedded in markdown!


chsdi/static/doc/source/index.rst
```diff
@@ -45,23 +45,54 @@ Use the GeoAdmin API Forum to ask questions: http://groups.google.com/group/geoa
...
-          center: [2680000, 1180000]
+          center: ol.proj.fromLonLat([7.44835, 46.94811]),
+          zoom: 7
         })
       });
+
     </script>
   </body>
 
-`Do you want to see the code? <api/quickstart.html>`_
```
pakb, line 63: Should't we provide here an exemple on codesandbox instead of nothing?
I've already a decent one here https://codesandbox.io/s/geoadmin-with-vanilla-openlayers-iwol2 but it is on my private codesanbox account. Do we have a swisstopo/geoadmin account there?

chsdi/static/doc/source/api/examples.rst
```diff
@@ -150,7 +139,7 @@ Other JS web mapping APIs integration
           break;
       };
     });
-    document.getElementById('cat1').innerHTML = cat1;
+    // document.getElementById('cat1').innerHTML = cat1;
```
pakb, line 142: Let's remove this altogether, shall we? 😄

chsdi/static/doc/source/index.rst
davidoesch: We have a Kogis codepen account. Can you move it there?

pakb: codepen doesn't do `import`, it's a bit old style. Should I create an account in codesandbox with the gmail account of IWI KOGIS as the user?

procrastinatio: @pakb We have a few dated (OL3 and 4) examples. I will move and update them to `codesandbox`

pakb (approved):
Great! Hopefully no new project will use this outdated JS API anymore

PR #3652, opened 2021-03-08
Revert "Feature bgdiinf sb 1527 fix search"

ltclm:
Reverts geoadmin/mf-chsdi3#3643


boecklic (approved):
As far as I can judge it's ok

PR #3671, opened 2021-03-26
fix 'NoneType' object has no attribute 'startswith' error

ltclm:
reported here https://jira.swisstopo.ch/browse/BGDIDIC-645

[fixed](http://mf-chsdi3.dev.bgdi.ch/feature_BGDIDIC-645_fix_mako/rest/services/ech/MapServer/ch.swisstopo.lubis-luftbilder-dritte-firmen/19910580133673/extendedHtmlPopup?lang=de)

[Testlink](https://mf-chsdi3.dev.bgdi.ch/feature_BGDIDIC-645_fix_mako/shorten/8f99d5b5c9)

### DIFF
--- a/chsdi/templates/htmlpopup/lubis.mako
+++ b/chsdi/templates/htmlpopup/lubis.mako
@@ -219,7 +219,7 @@ viewer_url = get_viewer_url(request, params)
         <td>${filename or '-'}</td></tr>
 % endif
 
-% if c['attributes'].get('doi_link', '') or '-'.startswith('http'):
+% if str(c['attributes'].get('doi_link', '')).startswith('http'):
     <tr><th class="cell-left">DOI Link</th>                          <td><a href="${c['attributes']['doi_link']}" target="_blank">${c['attributes']['doi_link']}</a></td></tr>
 % endif
     <tr><th class="cell-left">${_('tt_lubis_orthophoto')}</th>       <td>${c['attributes']['orthophoto'] or '-'}</td></tr>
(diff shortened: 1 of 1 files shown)


chsdi/templates/htmlpopup/lubis.mako
```diff
@@ -219,7 +219,7 @@ viewer_url = get_viewer_url(request, params)
         <td>${filename or '-'}</td></tr>
 % endif
 
-% if c['attributes'].get('doi_link', '').startswith('http'):
+% if c['attributes'].get('doi_link', '') or '-'.startswith('http'):
```
rebert, line 222: for what is the second argument in the .get function? if it would be a fallback value, we could put '-' there...?

ltclm: it will result in False if the attribute does not exist:
```python
 c['attributes'].get('doi_link', '-') or '-'.startswith('http')
False
```

ltclm: if the get('doi_link', '-') results in None we would end up with this error:
```python
None.startswith('http')
exec("""exec code in globs, locs""")
File "<string>", line 1, in <module>
File "<debugger>", line 1, in <module>
None.startswith('http')
```

b4l: `'-'.startswith('http')` will always be false and somewhat meaningless. better cast to str.

rebert (approved):
thanks

PR #3675, opened 2021-04-13
Update ch.pronatura.waldreservate

faselm:
Test link: https://mf-chsdi3.dev.bgdi.ch/ltfas_pronatura_waldreservate/shorten/8ff6c3dc61

### DIFF
--- a/chsdi/models/vector/dritte.py
+++ b/chsdi/models/vector/dritte.py
@@ -63,7 +63,6 @@ class PronaturaWaldreservate(Base, Vector):
     __table_args__ = ({'schema': 'pronatura', 'autoload': False})
     __template__ = 'templates/htmlpopup/pronatura.mako'
     __bodId__ = 'ch.pronatura.waldreservate'
-    __queryable_attributes__ = ['name', 'objnummer', 'gesflaeche', 'gesflaeche', 'gisteilobjekt']
     __label__ = 'name'
     id = Column('bgdi_id', Integer, primary_key=True)
     objnummer = Column('objnummer', Integer)
--- a/chsdi/models/vector/dritte.py
+++ b/chsdi/models/vector/dritte.py
@@ -72,6 +72,11 @@ class PronaturaWaldreservate(Base, Vector):
     gesflaeche = Column('obj_gesflaeche', Float)
     gisteilobjekt = Column('obj_gisteilobjekt', Float)
     mcpfe = Column('mcpfe', Unicode)
+    mcpfe_class_de = Column('mcpfe_class_de', Unicode)
+    mcpfe_class_fr = Column('mcpfe_class_fr', Unicode)
+    mcpfe_class_it = Column('mcpfe_class_it', Unicode)
+    mcpfe_class_en = Column('mcpfe_class_en', Unicode)
+    mcpfe_class_rm = Column('mcpfe_class_rm', Unicode)
     the_geom = Column(Geometry2D)
 
 register(PronaturaWaldreservate.__bodId__, PronaturaWaldreservate)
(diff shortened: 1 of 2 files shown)


chsdi/models/vector/dritte.py
```diff
@@ -72,6 +72,11 @@ class PronaturaWaldreservate(Base, Vector):
     gesflaeche = Column('obj_gesflaeche', Float)
     gisteilobjekt = Column('obj_gisteilobjekt', Float)
     mcpfe = Column('mcpfe', Unicode)
+    mcpfe_class_de = Column('mcpfe_class_de', Unicode)
+    mcpfe_class_fr = Column('mcpfe_class_fr', Unicode)
+    mcpfe_class_it = Column('mcpfe_class_it', Unicode)
+    mcpfe_class_en = Column('mcpfe_class_en', Unicode)
+    mcpfe_class_rm = Column('mcpfe_class_rm', Unicode)
```
b4l, line 79: Why specify rm but in the mako you map rm to de?

ltclm (changes requested):
thanks for this @faselm.
same here, either remove rm from model and table or fill it with data and remove the mapping.

chsdi/templates/htmlpopup/pronatura.mako
```diff
@@ -1,18 +1,18 @@
 <%inherit file="base.mako"/>
 
 <%def name="table_body(c, lang)">
-    <tr><td class="cell-left">${_('ch.pronatura.waldreservate.objnummer')}</td>                         <td>${c['attributes']['objnummer'] or '-'}</td></tr>
-    <tr><td class="cell-left">${_('ch.pronatura.waldreservate.name')}</td>                         <td>${c['attributes']['name'] or '-'}</td></tr>
```
rebert, line 5: name is the only Declaration that is being translated to rm (NOme)... which is not consistent

rebert (changes requested):
Nice. Only a translation the rm translation of ch.pronatura.waldreservate.name to adapt

faselm:
The rm translation has been fixed. However, it seems there are many inconsistencies in the translations for this layer and the bafu one, even on prod. In German, there is "Teilfläche Perimeter [ha]". Unless I'm missing something in German, perimeter in hectars doesn't make sense and the item in translated as "Area" and "Surface in English and French. Also as "Fläche" in rm... Not sure how all this should be fixed. Maybe you have an idea or I will also ask Mario...

ltclm:
@faselm you can fix the translations in the bod in the table ms_item:
```SQL
select * FROM ms_item where fk_layer_name like '%waldreservate' order by 3,2
```

the translations with for the attribute gisteilobjekt are empty there.

ltclm (changes requested):
looks good to me.
the attribute mcpfe is obviously not used anymore in chsdi. 
If thats correct, please remove it in this pr from model/mako. Since it is still used in the wms mapfile we will have to keep it in the table.

ltclm:
you would have to remove it from __queryable_attributes__ too, but please ask mario first.

faselm:
It's OK for Mario to remove mcpfe

chsdi/models/vector/dritte.py
```diff
@@ -63,15 +63,18 @@ class PronaturaWaldreservate(Base, Vector):
     __table_args__ = ({'schema': 'pronatura', 'autoload': False})
     __template__ = 'templates/htmlpopup/pronatura.mako'
     __bodId__ = 'ch.pronatura.waldreservate'
-    __queryable_attributes__ = ['name', 'objnummer', 'gesflaeche', 'gesflaeche', 'gisteilobjekt', 'mcpfe']
+    __queryable_attributes__ = ['name', 'objnummer', 'gesflaeche', 'gesflaeche', 'gisteilobjekt']
```
ltclm, line 66: you can remove the whole line from both layers ch.pronatura.waldreservate/ch.bafu.waldreservate, this feature has not been used at all in 2020 and 2021.

ltclm (changes requested):

ltclm (approved):

PR #3679, opened 2021-04-21
the delivered id is not stable

ltclm:


ltclm:
data owner insists on stable id

PR #3703, opened 2021-06-23
BGDIDIC-1954: Update ch.bafu.wasserbau-querprofilmarken

b4l:
Test link: https://mf-chsdi3.dev.bgdi.ch/feature_BGDIDIC-1954/shorten/9163c61d6c

### DIFF
--- a/chsdi/templates/htmlpopup/querprofilmarken.mako
+++ b/chsdi/templates/htmlpopup/querprofilmarken.mako
@@ -22,11 +22,19 @@
 </tr>
 <tr>
   <td class="cell-left">${_('ch.bafu.wasserbau-querprofilmarken.bww_km')}</td>
-  <td>${str(round(c['attributes']['bww_km'], 2)) or '-'}</td>
+  % if c['attributes']['bww_km']:
+  <td>${str(round(c['attributes']['bww_km'], 2))}</td>
+  % else:
+  <td>-</td>
+  % endif
 </tr>
 <tr>
   <td class="cell-left">${_('ch.bafu.wasserbau-querprofilmarken.lokale_km')}</td>
-  <td>${str(round(c['attributes']['lokale_km'], 2)) or '-'}</td>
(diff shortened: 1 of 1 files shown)


ltclm (changes requested):
htmlpopup is not always working:

https://mf-chsdi3.dev.bgdi.ch/feature_BGDIDIC-1954/rest/services/ech/MapServer/ch.bafu.wasserbau-querprofilmarken/71562/htmlPopup?coord=2644045.147216797,1130768.801147461&imageDisplay=1335,946,96&lang=en&mapExtent=2643107.394131605,1130145.3981681133,2645777.394131605,1132037.3981681133&sr=2056

b4l:
@ltclm the given example is working now. the root cause is that `round(c['attributes']['name'])` throws an error when the value is `null`. To prevent this we should change all instance of formatting patterns in makos like

b4l:
@ltclm the given example is now working. the root cause was an error thrown by `round()` experiencing a `null` input.

We probably want to apply the fix in the last commit throughput the code base and convert patterns in makos like:

`<td>${round(c['attributes']['bww_km'])) or '-'}</td>`

to

`  % if c['attributes']['bww_km']:
  <td>${round(c['attributes']['bww_km'], 2))}</td>
  % else:
  <td>-</td>
  % endif`

cc @rebert @marionb @faselm

ltclm:
thats a good point! we should try to find and fix this pattern in our code.

if it is ok to replace attribute values **None and 0 with '-'** we could it even simplify it to:
`<td>${round(c['attributes']['bww_km'] or 0, 2) or '-'}</td>`

b4l:
> if it is ok to replace attribute values **None and 0 with '-'** we could it even simplify it to:
> `<td>${round(c['attributes']['bww_km'] or 0, 2) or '-'}</td>`

I don't think we want to default to 0 and/or replace 0 with '-'

ltclm (approved):

PR #3707, opened 2021-07-01
Remove bakom layers

faselm:


ltclm (approved):
lgtm, thank you!
@consta-v do we have to deploy this before the next deploy?

rebert (approved):

consta-v:
Hi @ltclm 
For me, it is not necessary to deploy this change. The effort to do so is not justified for such a small change. You can let it like this. These layers will be updated in the next weeks anyway.

ltclm:
thank you for the feedback @consta-v 
we will publish this with the next deploy.

b4l:
@ltclm & @faselm i think we can close this as there will be an actual update so that the announcement just proceeds it.

PR #3719, opened 2021-08-10
BGDIDIC-2200: Add ch.bfe.waermepotential-gewaesser

b4l:
https://mf-chsdi3.dev.bgdi.ch/feature_BGDIDIC-2200/shorten/925af843a8


faselm:
Looks good, but I get: Could not retrieve information for ch.bfe.waermepotential-gewaesser for fr, it, en.. I think we had that once already and it was because of some missing translations in the bod if I remember correctly... The legends are HUGE. Is it desired like that?

faselm:
There is also a missing "P" in French for "otentiel de refroidissement en été" in the ms_item table

b4l:
@faselm thanks for the review! I fixed the fr translation mistake. The unretrievable information is due to lacking translations from geocat. Also the legends were delivered like this so i leave it to the customer to judge and adapt it.

faselm (approved):

ltclm (dismissed):
looks all good to me except the size of the legend images, width should be smaller than: 480px -> https://ltwiki.adr.admin.ch:8443/display/BGDI/03_VIEWER#id-03_VIEWER-E)Thelegend

ltclm (approved):
thank you!

PR #3723, opened 2021-09-01
Update legends ch.bfe.thermische-netze

faselm:
New pr as the client reversed the changes.
Test link: https://mf-chsdi3.dev.bgdi.ch/ltfas_bfe_thermische_netze/shorten/92cbbebf64


ltclm (approved):
can we completely remove the attribute contact from the table,  chsdi model and identify response too?

faselm:
> can we completely remove the attribute contact from the table, chsdi model and identify response too?

@ltclm I don't know but there is still data delivered for it in the raw data. I can ask Véronique..

ltclm:
> > can we completely remove the attribute contact from the table, chsdi model and identify response too?
> 
> @ltclm I don't know but there is still data delivered for it in the raw data. I can ask Véronique..

yes please, in this state we will still have the attribute in identtify response but not in htmlpopup response which is not ideal.

faselm:
@ltclm It will be still used in the extended tooltip...

AFoletti:
> @ltclm It will be still used in the extended tooltip...

Dear integrators
Just a heads-up from the BFE side of the barricade 😉 

If the data contains an attribute and it is referenced in the DIF table, then it is meant to be available via the BGDI services even if it does not come up in the normal or extended tooltip.
We sometimes build apps/visualizations on the BDGI services that use attributes not visible in any tooltip.
Thanks!

PR #3724, opened 2021-09-03
Add relnotes 20210908

faselm:


rebert (commented):
ltgm

mariokeusen:
Looks good to me apart from the ch.bafu.gewaesserschutz-badewasserqualitaet - update, which is missing on the Releasenote. @faselm could you please add that?

faselm:
@mariokeusen Salut Mario, the layer is already in the doc in fact (after ch.bfe.energieforschung)...

mariokeusen:
> @mariokeusen Salut Mario, the layer is already in the doc in fact (after ch.bfe.energieforschung)...

Ah, didn't realize it's not the same order than in the remember doc ;-)

PR #3735, opened 2021-09-15
small clean-up in the Makefile

hansmannj:
* added new Makefile targets to help target
* using different AWS region for ECR (dockerlogin target) than the default AWS region


procrastinatio (approved):
LGTM

PR #3736, opened 2021-09-16
BGDIDIC-1399: Cleanup model

b4l:


ltclm (approved):
thank you!
if possible please remove these columns from the workbench and the table after the deploy.

b4l:
> if possible please remove these columns from the workbench and the table after the deploy.

The workbenches are already up to date and there is a post deploy task note in the remember when deploy document.

PR #3739, opened 2021-09-27
Add legends

faselm:


b4l (approved):
Looks good. I am on mobile internet, it is very slow & only partially loading.

PR #3756, opened 2021-11-09
Ltmbr belastung personenverkehr

marionb:
test link https://mf-chsdi3.dev.bgdi.ch/ltmbr_belastung-personenverkehr/shorten/9431e542ee


faselm (approved):
LGTM

ltclm (approved):

ltclm:
if nr is a stable id, provided by the data owner we could adapt the chsdi model and:
* set nr as primary key in the models
* set stable_Id to true in the htmlpopup mako
but please let this be confirmed by @mariokeusen

marionb:
@ltclm good point about the sable ID's. I'm checking with @consta-v and will change it accordingly once I have a confirmation.

marionb:
@ltclm 
`nr` turned out to be neither unique nor stable so I will keep using `bgdi_id` for this and no stable id. It was a good point to check this thoroughly :-)

procrastinatio:
Did the rebase onto `data`. Wait until all tests are passing and merge.

PR #3765, opened 2021-11-22
Fix rounding ch.bafu.bundesinventare*

faselm:
Test: https://s.geo.admin.ch/94c4bacc40


chsdi/templates/htmlpopup/vogelreservate.mako
```diff
@@ -13,8 +13,8 @@
     <tr><td class="cell-left">${_('ch.bafu.bundesinventare-vogelreservate.teilgebiet')}</td>         <td>${c['attributes']['teilgebiet'] or '-'}</td></tr>
     <tr><td class="cell-left">${_('ch.bafu.bundesinventare-vogelreservate.schutzkategorie')}</td>         <td>${c['attributes'][schutzkat] or '-'}</td></tr>
     <tr><td class="cell-left">${_('ch.bafu.bundesinventare-vogelreservate.schutzebene')}</td>         <td>${c['attributes'][schutzeb] or '-'}</td></tr>
-    <tr><td class="cell-left">${_('ch.bafu.bundesinventare-vogelreservate.shape_area')}</td>          <td>${round(c['attributes']['shape_area']/10000) or '-'}</td></tr>
-    <tr><td class="cell-left">${_('ch.bafu.bundesinventare-vogelreservate.obj_gisflaeche')}</td>         <td>${round(c['attributes']['obj_gisflaeche']) or '-'}</td></tr>
```
marionb, line 17: I don't know the exact spec for this layer but would it not be better to round the data that is written in the DB so it will be the same every where and not just for the tool-tip?

b4l (approved):
This only changes the rounding precision in the tooltip and has nothing to do with the pr title?!

faselm:
Yes, edited, I prepared that for many layers but there were less modifs on chsdi than expected...

PR #3778, opened 2021-12-09
Add legends ch.swisstopo.strassenkarte-200

faselm:
Test: https://mf-chsdi3.dev.bgdi.ch/ltfas_swisstopo_strassenkarte_200/shorten/94cbf0f254


b4l (approved):
The small legend seems to be the same as the big one and thus unreadable given the small scale. Hope this is intended!?

faselm:
Yes, it is intended (the big legend is a pdf the user can zoom in normally)...

PR #3780, opened 2021-12-10
Add ltfas layers

faselm:


b4l (changes requested):
Please use the `relnotes_20211215`  branch as base and move your layers to the appropriate release date section. Thanks!

PR #3794, opened 2022-01-26
added xyz description

davidoesch:
@procrastinatio can you review and add an example in codepen?
@p1d1d1 feedback welcome

### DIFF
--- a/chsdi/static/doc/source/services/sdiservices.rst
+++ b/chsdi/static/doc/source/services/sdiservices.rst
@@ -1114,6 +1114,68 @@ Example
 * Switzerland is now adopting the new `LV95 frame <http://codepen.io/geoadmin/pen/GZKEam?editors=0010>`_.
 * All `available layers as WMTS <http://codepen.io/geoadmin/pen/MyYYXR?editors=0010>`_.
 
+.. _XYZ_description:
+
+XYZ
+----
+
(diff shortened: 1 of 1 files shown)


ltshb:
@davidoesch and @procrastinatio can this be merged and deployed with the next deploy on 2022-03-16 ?

davidoesch:
A go from my side

ltshb (changes requested):
This can not be merge like this. This is not production ready and still contain TODO.
@davidoesch what is XYZ ? Looks to be a quick copy paste of the WMTS chapter, I don't understand the goal of adding this chapter when it is already documented in WMTS. Note to not block current deploy of 2022-03-16 I'll remove this PR from milestone.

rebert:
going to make a PR on branch develop-2022-05-18

davidoesch:
so pretty please, with sugar on the top, can IGEB  please contribute to the issue as mentioned above?

ltshb:
@davidoesch please refer to the new PR #3837 not this one which is closed. IMHO it does not only need an example, but clearly a definition of what you want ! A copy paste of service WMTS chapter with just some small changes for describing something that the same is not a good documation at all. As I already say we have a process with jira and if you at first defined directly in a Jira ticket  and/or talked in our planning what you really want this would have been already done without sugar. So stop commenting this here and bring this in our regular planning meeting, otherwise we can stop scrum and planning.

PR #3795, opened 2022-02-01
BGDIDIC-295 & BGDIDIC-1992: Adapt tool-tip

marionb:
testlink:
https://mf-chsdi3.dev.bgdi.ch/data-JIRA-BGDIDIC-293-Adapt_tool-tip/shorten/95dbb9b5fd


faselm:
Only "Consulting Costs" is translated in the EN tooltip field names for me. Not sure it's intended

marionb:
@faselm that is a good point I will check that with @consta-v. It is actually this way in the DB. (With the German word in the English column)
In any case this it will not have an effect on the tool-tip code.

ltclm (approved):

ltclm:
@marionb please rebase, tests should pass first.

PR #3796, opened 2022-02-01
BGDIINF_SB-2163: Fixed CI branch name resolution

ltshb:
Removed old hack for branch name resolution. This hack did not work with
PR from fork as well as by some PR from dependabot.

Use the newest Codebuild variable instead and make sure that the branch
name used in docker tag don't contains `/` by replacing them to `_`
(`/` is not allowed in docker tag name). For this replacement we need
bash so make sure that Codebuild use bash instead of the default hash.


ltshb:
@ltclm could you have a look why the search test is not passing here ?

ltclm:
> @ltclm could you have a look why the search test is not passing here ?

the reason is the data behind this service. This data from BFS is updated every day. The address that is being used in the test does not exist anymore:
`chemin isabelle-de-montolieu 2 1010 lausanne 5586 lausanne ch vd`

we would have to prepare a hotfix for this.

rebert (approved):

ltclm:
https://github.com/geoadmin/mf-chsdi3/pull/3797

PR #3801, opened 2022-02-03
Fix link to terms of use

davidoesch:


rebert:
@davidoesch going to make a pr for the branch https://github.com/geoadmin/mf-chsdi3/tree/develop-2022-03-16 instead

PR #3805, opened 2022-02-07
Bug fix make templates

ltclm:
@procrastinatio im not sure if im breaking something in the docker world with this. Im sure there are better ways to fix this.
seems to be changed with this pr:
https://github.com/geoadmin/mf-chsdi3/pull/3725

make templates has to be executed as `source rc_user && make template`
CURRENT_DIRECTORY has not been injected in development.ini anymore.


procrastinatio:
Tests on python3 are passing, so at least the docker images is build and may be started.

PR #3808, opened 2022-02-16
Fixing service examples

rebert:
Some of the service examples have been fixed, because they didn't work anymore (esp. identify examples with old bfs layer data)


mariokeusen (approved):
Thx @rebert lgtm

PR #3810, opened 2022-02-16
Add fields BGDIDIC-935

faselm:


chsdi/models/vector/edi.py
```diff
@@ -80,6 +98,16 @@ class Gebaeuderegister(Base, Vector):
     wbez = Column('wbez', Unicode)
     wstat = Column('wstat', SmallInteger)
     wexpdat = Column('wexpdat', DateTime)
+    egrid = Column('egrid', SmallInteger)
```
ltclm, line 101: these atrributes are already part of the model

ltclm: they are part of the `bfs.gwr_gebaeude` table

PR #3811, opened 2022-02-24
New dynamodb parameter

procrastinatio:
To ease the migration of the dynamodb we add a new parameters in the old DB used by the old service. This is need because dynamodb's global secondary indices (GSI) do not support any operation other than equalité, which means scanning the intere DB to filter the result.

Testing: http://mf-chsdi3.dev.bgdi.ch/feature-BGDIINF_SB-2043/

Try mit https://mf-chsdi3.dev.bgdi.ch/feature-BGDIINF_SB-2043/shorten.json?url=https:%2F%2Fmf-geoadmin3.dev.bgdi.ch

or https://map.geo.admin.ch/?api_url=//mf-chsdi3.dev.bgdi.ch/feature-BGDIINF_SB-2043

### DIFF
--- a/chsdi/views/shortener.py
+++ b/chsdi/views/shortener.py
@@ -17,7 +17,7 @@ from chsdi.lib.helpers import check_url, make_api_url
 MAX_URL_LENGTH = 2048
 
 
-def _add_item(table, url):
+def _add_item(table, url, staging='prod'):
     url_short = _get_url_short(table, url)
     if url_short is None:  # pragma: no cover
         # Create a new short url if url not in DB
@@ -31,7 +31,8 @@ def _add_item(table, url):
                     'url_short': url_short,
                     'url': url,
                     'timestamp': time.strftime('%Y-%m-%d %X', now),
-                    'epoch': time.strftime('%s', now)
+                    'epoch': time.strftime('%s', now),
+                    'staging': staging,
                 }
             )
         except boto_exc.Boto3Error as e:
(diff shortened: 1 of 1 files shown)


chsdi/views/shortener.py
```diff
@@ -31,7 +31,8 @@ def _add_item(table, url):
                     'url_short': url_short,
                     'url': url,
                     'timestamp': time.strftime('%Y-%m-%d %X', now),
-                    'epoch': time.strftime('%s', now)
+                    'epoch': time.strftime('%s', now),
+                    'day': time.strftime('%Y-%m-%d', now),
```
ltshb, line 35: Would not be `'date'` as field name more appropriate here ?

ltshb:
Don't if this would help about performance, but in DynamoDB for KML we have a secondary index on the kml ID sorted by timestamp, then we can do a query using the timestamp as bigger than. That's what the migration script is using for subsequent migration to avoid re-scanning the whole table. Would it be not better to try this approach instead of adding redundant information ? I mean creating a secondary index based on the short_url and sorted by timestamp.

ltshb (changes requested):
To avoid having 3 times the same data in the Table, I propose to do as follow, add the `staging: dev|int|prod` data to the table. Although this info is not really useful because we have one table per staging, but it allow to create a sorted index on the `staging` key sorted by timestamp (you can use either the epoc or timestamp, doesn't matter, but the string timestamp might be either to use). Then you can run a query to return all entries with staging=prod with a timestamp bigger than x. That's what we do with service kml. @boecklic what do you think ?

procrastinatio:
But OK, I'll add a kind of dummy `staging` key and sort on the `timestamp`

ltshb (approved):
👍🏼

chsdi/views/shortener.py
procrastinatio: New key `staging`

PR #3815, opened 2022-03-07
Exclude geojson layer from tooltips layer (docu only)

procrastinatio:
`GeoJSON` have a client-side only tooltip and are hence not callable through the `identify` service. This PR corrects the doc, which is misleading (via `google-group GeoADMIN`)

### DIFF
--- a/chsdi/static/doc/source/api/faq/index.rst
+++ b/chsdi/static/doc/source/api/faq/index.rst
@@ -104,7 +104,8 @@ The following list contains all the free accessible layers:
 Which layers have a tooltip?
 ****************************
 
-Not all the layers have a tooltip. Below, you can find the complete list all the layers that have a tooltip:
+Not all the layers have a tooltip. The complete list of layer using the `htmlPopup <../../services/sdiservices.html#htmlpopup-resource>`_ service is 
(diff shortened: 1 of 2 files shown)


rebert:
@procrastinatio there are certain layers where the identify works even though it is a geojson layer. F.ex. ch.bfe.ladestellen-elektromobilitaet ...

rebert:
@procrastinatio what about a cleanup in bod? set the geojson layers that have no model in chsdi registered to `tooltip=false` ?

rebert (changes requested):
I will adapt the bod

procrastinatio:
No, check `mf-geoadmin3` code first. You can't change the value in the BOD, because you have geojson layer with tooltip (most BAFU layers) and geosjon layer without it (most MeteoSchweiz).

rebert:
@procrastinatio here the exception ch.bfe.ladestationen-elektromobilitaet https://github.com/SFOE/DIEMO-Documentation/blob/master/How%20to%20query%20DIEMO.md

procrastinatio:
OK, I was wrong `identify`has nothing to to with `tooltips` even if it's also used. We have to refer to `htmlPopup`

rebert (approved):

PR #3817, opened 2022-03-08
Removed useless and annoying linting git hook

ltshb:
Now you are allow to create commit even if the code doesn't pass linting.

Linting is ensured by the CI which runs linting and you are not allowed to merge a PR
where the CI is failing. On the other hand it is allow to create an push non conform code as
work in progress on a local or personal branch.


ltclm (approved):
i would personally prefer to know about annoying linting issues before committing code and sending it through ci pipelines. but commit hooks can be bypassed easily so its probably better like that.

ltshb:
@ltclm  the issue with the linting commit hook is that it prevent sometimes interactive rebasing. I mean linting can still be run manually before creating a PR. Or if this hook is really wanted we should move it to a push hook to prevent pushing code with linting issues, but again this would prevent to save a work in progress code that is not yet linting compliant.

procrastinatio (approved):
Great. Thx.
Maybe also remove the `git secret` which is a relict of a distant path.

ltshb:
@procrastinatio removing `git secrets` is a great idea, we can do this in a separate PR and maybe after the migration to Frankfurt, then we could remove all the infra deployment stuff from the repo and the `git secret` would make even less senses.

PR #3818, opened 2022-03-08
BGDIDIC-154: Rename attributes ch.swisstopo.amtliches-gebaeudeadressverzeichnis

b4l:
https://mf-chsdi3.dev.bgdi.ch/feature_BGDIDIC-154/shorten/9699369a39

### DIFF
--- a/chsdi/models/vector/stopo.py
+++ b/chsdi/models/vector/stopo.py
@@ -3097,20 +3097,20 @@ class AmtlichesAdressVerzeichnis(Base, Vector):
     __table_args__ = ({'schema': 'vd', 'autoload': False})
     __template__ = 'templates/htmlpopup/addressverzeichnis.mako'
     __bodId__ = 'ch.swisstopo.amtliches-gebaeudeadressverzeichnis'
-    __label__ = 'id'
-    __queryable_attributes__ = ['id', 'bdg_egid', 'adr_edid', 'str_esid', 'str_label', 'adr_number']
-    id = Column('adr_egaid', BigInteger, primary_key=True)
+    __label__ = 'adr_egaid'
+    __queryable_attributes__ = ['adr_egaid', 'bdg_egid', 'adr_edid', 'str_esid', 'stn_label', 'adr_number']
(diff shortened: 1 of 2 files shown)


ltclm (approved):
thank you!

b4l:
@ltclm in order to facilitate serving `adr_egaid` in the identify rest api we probably need to load this attribute into `bgdi_id` and set the id to bgdi_id in the model.

ltclm:
no please dont use bgdi_id for that, it is an internal counter with no relation to the data. please create a new primary key column and use this column in the chsdi  model and sphinx query.

chsdi/models/vector/stopo.py
```diff
@@ -3097,20 +3097,20 @@ class AmtlichesAdressVerzeichnis(Base, Vector):
     __table_args__ = ({'schema': 'vd', 'autoload': False})
     __template__ = 'templates/htmlpopup/addressverzeichnis.mako'
     __bodId__ = 'ch.swisstopo.amtliches-gebaeudeadressverzeichnis'
-    __label__ = 'id'
-    __queryable_attributes__ = ['id', 'bdg_egid', 'adr_edid', 'str_esid', 'str_label', 'adr_number']
-    id = Column('adr_egaid', BigInteger, primary_key=True)
+    __label__ = 'adr_egaid'
+    __queryable_attributes__ = ['adr_egaid', 'bdg_egid', 'adr_edid', 'str_esid', 'stn_label', 'adr_number']
+    id = Column('id', BigInteger, primary_key=True)
```
ltclm, line 3102: is it not possible to just referene adr_egaid as source here? Why do we have to fill two columns with the same value?

b4l: @ltclm according to the ticket we want to return `adr_egaid` as attribute name. if my test were right it is not possible to reference the same column to different attributes in the model, hence the duplication.

ltclm: ok, thank you!

PR #3822, opened 2022-03-09
test-ci

ltshb:


ltshb:
This was just a test for CI which worked see https://jira.swisstopo.ch/browse/BGDIINF_SB-2262 and also https://github.com/geoadmin/infra-terraform-bgdi-builder/pull/40

PR #3836, opened 2022-03-15
Fix for layer ch.swisstopo-vd.stand-oerebkataster did not work on all zoom levels

rebert:
… zoom levels


ltshb (approved):
Just modified the PR title ;-)

PR #3848, opened 2022-04-05
BGDIINF_SB-2284: Removed qrcode that has been migrated into a micro service

ltshb:
@boecklic do we want to keep the link to service qrcode in /dev endpoint of mf-chsdi3 ? Same questions for the others (kml, shortlink, wmts capabilities, ...)


procrastinatio (approved):
OK. Looks good.

PR #3849, opened 2022-04-05
BGDIINF_SB-2326: Changed loader.js endpoint to use sys-api3 instead of sys-public

ltshb:
@boecklic do we want to also have 3 staging buckets for the loader js I guess (to be consistent) ? Looking at the code changes in mf-chsdi3 would it not make more sense to change the loader.js once the migration is done ? 

As test we can use https://alert.swiss


chsdi/views/loader.py
```diff
@@ -17,9 +17,10 @@ def loadjs(request):
     ignore_polyfill = request.params.get('ignore_polyfill')
     # Determined automatically in subscriber
     lang = request.lang
-    public_bucket_host = request.registry.settings['public_bucket_host']
+    loader_js_bucket_host = request.registry.settings['loader_js_bucket_host']
     if vip is True:
-        public_bucket_host = public_bucket_host.replace('public', 'public-cdn')
+        # TODO what is the vip mode ?
+        loader_js_bucket_host = loader_js_bucket_host.replace('public', 'public-cdn')
```
ltshb, line 23: @boecklic @procrastinatio what is this `vip` mode in loader.js ?

procrastinatio: I guess, it was backen then when api3 was not behind CF. We (=Gilbert I guess) put a copy of the js api files in a bucket (public.geo.admin.ch ???) which was behind CF. It was an emergency hack because the air-raid-siren test page did destroy our servers. Now, as everything is behind CF, I think we may ditch this mode (I remember regexing in the compressed js files...)

ltshb: @boecklic @procrastinatio apparenlty the host `public-cdn.geo.admin.ch` is not anymore reachable, so we can remove this `vip` mode ?

ltshb: @procrastinatio I checked, currently the loader is cached by CF for 30 minutes

```bash
curl https://api3.geo.admin.ch/loader.js --head
HTTP/2 200 
content-type: application/javascript
content-length: 444946
date: Wed, 06 Apr 2022 10:19:05 GMT
server: Apache/2.4.10 (Debian)
access-control-allow-origin: *
access-control-allow-methods: POST, GET, OPTIONS
access-control-allow-headers: x-requested-with, Content-Type, origin, authorization, accept, client-security-token
x-ua-compatible: IE=Edge
cache-control: max-age=1800, public
x-varnish: 637440651 627667681
via: 1.1 varnish (Varnish/6.1), 1.1 9349ae4f82564896b96f5303b030d188.cloudfront.net (CloudFront)
x-cache-hits: 3
accept-ranges: bytes
vary: Accept-Encoding
x-cache: Hit from cloudfront
x-amz-cf-pop: ZRH50-C1
x-amz-cf-id: oWFqsWwfMU09VP0gCkvR-JRHHLlpxaH5Vp86MJSQ1yLhQyOfYXFssw==
age: 671
```
By the way the `vip` mode is currently not working as it would look for other resources in `public-cdn.geo.admin.ch` which doesn't exists anymore.

procrastinatio (approved):
Looks OK. But many things have change since we first implemented this. Now everything is behind CF, which was not the case back then. I would prefer to store these file in github and the static dir, instead of forgetting them in S3. And having a few seconds (10-60s) cache even on the `loader.js` is not a big deal IMHO (was taboo for a long time). These files will not longer change and the OL version is they use is very "mature" (5-6 years old)

PR #3871, opened 2022-05-09
BGDIINF_SB-2319: Searched for shop and removed what seems reasonable

rebert:


chsdi/static/doc/source/services/sdiservices.rst
```diff
@@ -646,7 +646,7 @@ The search service is separated in 3 various categories or types:
 * The **location search** which is composed of the following geocoded locations:
 
   * Cantons, Cities and communes
-  * All names as printed on the national map (`SwissNames <https://shop.swisstopo.admin.ch/de/products/landscape/names3D>`_)
+  * All names as printed on the national map
```
b4l, line 649: may leave `SwissNames`

chsdi/static/doc/source/services/sdiservices.rst
```diff
@@ -871,7 +871,6 @@ Height
 ------
 
 This service allows to obtain elevation information for a point.
-See `Height models <https://shop.swisstopo.admin.ch/de/products/height_models/alti3D>`_ for more details about data used by this service.
```
b4l, line 874: may just point to another source instead of removing it. same above and below

ltclm (approved):

PR #3873, opened 2022-05-13
Starting with releasenotes

rebert:
Releasenotes 2022-05-18


consta-v:
The announcements look good to me. The new layers have not been added at the moment.

faselm:
> Is there a reason why ch.bazl.luftraeume-kontrollzonen is listed as update twice ?

No, I removed it

b4l (approved):

PR #3879, opened 2022-05-19
Dockerization - WHAT DO WE DO WITH THIS ?

ltclm:
this branch can be used  as a shared playground for the docker runtime configuration.


hansmannj:
> this branch can be used as a shared playground for the docker runtime configuration.

Would that mean we can all commit to it in parallel?

dev_frankfurt.env
```diff
@@ -0,0 +1,58 @@
+PRINT_ENV=dev
+RANCHER_LABEL=dev
+IMAGE_TAG=staging
+NGINX_PORT=8009
+TOMCAT_PORT=8011
+API_URL=//mf-chsdi3.dev.bgdi.ch
```
ltshb, line 6: Most url are wrong and should be ‘sys-‘

hansmannj: Yes, good point. I have stopped at a point yesterday and we can correct everything in this file as we go

hansmannj: At least `VECTOR_BUCKET` should be already correct. The rest is more or less copy paste from `dev.env`

docker-entrypoint.sh
```diff
@@ -16,9 +16,10 @@ echo "INSTALLDIR=${INSTALLDIR}"
 
 pg_isready -d stopo_prod -h ${DBHOST} -p ${DBPORT} -U ${PGUSER}
 
-source rc_${DEPLOY_TARGET} && make production.ini development.ini
+cut -c8- ${DOCKER_ENV_FILE} > ${DOCKER_ENV_FILE}_no_export
```
hansmannj, line 19: This could help to get rid of `dev_frankfurt.env` and `dev.env`, as only `rc_dev` and `rc_dev_frankfurt` will be sufficient

dev_frankfurt.env
```diff
@@ -0,0 +1,58 @@
...
+APACHE_BASE_PATH=main
+APACHE_ENTRY_PATH=
+APACHE_LOG_LEVEL=debug
+APACHE_PORT=8009
+API_URL=//api3.geo.admin.ch
+AWS_DEFAULT_REGION=eu-central-1
+CACHE_CONTROL=no-cache
+CMSGEOADMINHOST=https://cms.geo.admin.ch
+DATAGEOADMINHOST=data.geo.admin.ch
+DBHOST=pg.bgdi.ch
```
ltclm, line 31: here we should switch to `pg-geodata-replica.bgdi.ch` which is only available in frankfurt

hansmannj:
From my point of view, this PR can be closed and the branch deleted, as we should go with a dedicated protected branch for the migration, starting clean from master. OK for you @ltclm and @procrastinatio?

ltclm:
@hansmannj ok for me

PR #3880, opened 2022-05-20
BGDIINF_SB-2328: Remove region specification from S3Connect()

hansmannj:
Currently the S3Connect() uses eu-west-1 as default region for the s3 buckets. The migrated buckets are in
eu-central-1. As the bucket names are unique across all regions, things should work without a region specified.
Hence the region references are deleted from S3Connect() to simplify things


chsdi/models/clientdata_dynamodb.py
```diff
@@ -21,8 +21,13 @@ def get(self):
                     'S3: Error during connection init %s' % e)
         return self.conn
 
+VECTOR_BUCKET_REGION = os.getenv('VECTOR_BUCKET_REGION', None)
```
rebert, line 24: in other services we used f.ex. settings.py to retrieve all env values. Then it is clear which env values are being captured and when. Maybe we should do so here as well.

hansmannj: Makes sense to have something like this in the long run (post-migration clean-up).
For now I'll adapt this PR to only rely on ENV vars and to not have any explicit region settings here at all.

ltshb (changes requested):
Actually `region` in S3 doesn't make sense at all and is probably ignored by boto3. S3 bucket are universally unique, meaning that we can't have two buckets with the same name in different account and/or region. We should simply remove the region argument totally from this file and it might be aslo a good idea to rename it to `s3_client.py`

procrastinatio (commented):
In a Pylons application all settings are normally defined in the `.ini` files. So this won't work in the vhost.

rebert (approved):

ltshb (approved):
👍🏼

PR #3881, opened 2022-05-22
use the new pg replica dns pg-geodata-replica.bgdi.ch in frankfurt en…

ltclm:
…vironment


ltshb:
IMHO the `*.env` file should stay with the ireland stuff. In frankfrut we use k8s environment to work with. Is it ok to have one single  frankfurt env file for dev now to simplify debugging on the developmement vhost. But then no environment file should be here.

procrastinatio:
I propose to close this one, and use PR https://github.com/geoadmin/mf-chsdi3/pull/3895

PR #3885, opened 2022-05-23
feature_BGDIINF_SB-2328: sorted and cleaned up dev.env entries

hansmannj:
* sorted `dev.env` alphabetically with `sort -t= dev.env > temporaryfile.env` in order to make it easier to identify redundant entries and removed those.
* adapted several URLs in `dev.env` to point to the current DEV in Frankfurt


dev.env
```diff
@@ -1,55 +1,45 @@
...
 DBPORT=5432
 DBSTAGING=dev
+DEBUG=True
 DEPLOY_TARGET=dev
 DYNAMIC_TRANSLATION=1
-GEOADMINHOST=map.geo.admin.ch
-GEODATA_STAGING=dev
-HOST=api3.geo.admin.ch
+GEOADMINHOST=mf-geoadmin3.dev.bgdi.ch
+GEODATA_STAGING=test
```
hansmannj, line 17: Not sure if that is correct. If I remember correctly, we changed that from `dev` to `test` last Friday @procrastinatio, right? Is `test` the correct value here?

dev.env
```diff
@@ -1,55 +1,45 @@
...
+GIT_BRANCH=
+HOST=sys-api3.dev.bgdi.ch
 HTTP_PROXY=http://ec2-52-28-118-239.eu-central-1.compute.amazonaws.com:80
+IMAGE_TAG=staging
 INSTALLDIR=/var/www/vhosts/mf-chsdi3/private/chsdi
 KEEP_VERSION=true
 KML_TEMP_DIR=/var/local/print/kml
 LINKEDDATAHOST=https://ld.geo.admin.ch
 MODWSGI_CONFIG=development.ini
```
hansmannj, line 26: Not sure if this is correct. Think I also saw `production.ini` for this var somewhere.. Do you know @procrastinatio?

procrastinatio (approved):
OK, for now. Thx

dev.env
procrastinatio: Both `production.ini` `development.ini`are tmplate generated. So we may use only one template or have three more static files

PR #3889, opened 2022-05-25
BGDIINF_SB-2328: Put all logs to stdout

ltshb:
On k8s all logs should be sent to stdout in order to read them from the
k8s log command or even better from the ELK stack.


procrastinatio (approved):
Fin for docker. Still looks OK for the vhost. Thx

PR #3894, opened 2022-05-30
BGDIINF_SB-2328 Adding missing target to Makefile

procrastinatio:
Add missing target to `Makefile` help and some doc clarification 

No functional changes


hansmannj (approved):
Thanks!

PR #3897, opened 2022-05-31
BGDIINF_SB-2328_: Remove apache proxy

procrastinatio:
The pylons chsdi3 application is served directly by `mod_wsgi` and
not by `gunicorn`as in other project. So no proxy needed.


hansmannj (approved):
:+1:

PR #3866, opened 2022-04-27
Helping the migration to S3 Frankfurt

procrastinatio:

PR #3898, opened 2022-05-31
BGDIINF_SB-2328: debug and log info for service `identify` on grid (S3)

procrastinatio:
The bucket was moved from renamed and the service has still errors.

Replacing https://github.com/geoadmin/mf-chsdi3/pull/3866


chsdi/models/s3_client.py
```diff
@@ -27,8 +30,11 @@ def get_file_from_bucket(bucket_name, file_name):
     try:
         response = conn.get_object(Bucket=bucket_name,
                                    Key=file_name)
+    except boto_exc.NoCredentialsError as c:
```
hansmannj, line 33: Would it probably also help to do something like a `log.error()` or `log.debug()` or the like with the anonymized details on the bucket and to not have any details on the bucket in the raised error message itself?

hansmannj, line 33: Would it probably also help to do something like a `log.error()` or `log.debug()` or the like with the anonymized details on the bucket and to not have any details on the bucket in the raised error message itself?

procrastinatio: Same, the error is catched after in https://github.com/geoadmin/mf-chsdi3/pull/3898/files#diff-ef400a87c6b46b2854e0f728421924a1111c8eedbc04056c6b3ae79bfc531e21R447

hansmannj: OK, I see. Doesn't make sense to log the same error twice, you're right.
The other point was, if it would make sense to only have the sensitive details of the error message (anonymized bucket name) in the logged error, and not in the more visible raised error message, that the user will directly see.

hansmannj: @procrastinatio: What do you think about this point? Would it be OK if we simply drop the anonymized bucket name from the raised error message and keep it only in the logs?

procrastinatio: This what I've done in https://github.com/geoadmin/mf-chsdi3/pull/3898/commits/e85c7ea5edadd59d8dfc6842d2bfc42d128addca

hansmannj: OK, sorry, I did not realize that. I had in mind, that "only" half of the name was anonymized.

hansmannj (approved):
:+1:

PR #3903, opened 2022-06-01
BGDIINF_SB-2328: removing the DEPLOY_TARGET variable from docker/frankfurt stuff

procrastinatio:
The variable was used to build and run tests the code prior to deploy it
to the `DEPLOY_TARGET` (no `CI` back then, and deploying was a lengthy and
risky process)


hansmannj (approved):
:+1:

PR #3904, opened 2022-06-01
BGDIINF_SB-2328: add Makefile.frankfurt target `rss`

procrastinatio:
This build the RSS feed from the releasenotes HTML page. This step
may or should be done outside of docker.


hansmannj (approved):
:+1:

PR #3905, opened 2022-06-01
BGDIINF_SB-2328: Discarding docker-compose

procrastinatio:


hansmannj (approved):
:+1:
Not sure, if docker-compose might be helpful for testing locally though.
But I guess, since we have a `dockerrun` target, we can as well drop docker-compose.

procrastinatio:
@hansmannj Yes, we may add it later, if needed. For now, it's a file less ;-)

PR #3908, opened 2022-06-02
BGDIINF_SB-2328: target to build the css file for the doc

procrastinatio:
Installing so much  `node` for a simple CSS is a bit overkill, IMHO

Before merging, check if some targets in `all` and `setup` are not missing.


hansmannj (approved):
LGTM

PR #3911, opened 2022-06-02
BGDIINF_SB-2328: allowing to use env vars from ENV_FILE in Makefile.frankfurt

hansmannj:
The function `setup_env` in `Makefile.frankfurt` was not working as expected. Some env vars from the ENV_FILE were not used when envsubst created base.ini and local.ini (this lead to an undefined port when running the serve target: `self.port = int(port)
ValueError: invalid literal for int() with base 10: ''
Press ENTER or change a file to reload.`. Hence that function was replaced by an auxiliary Makefile target environ_aux, which seems to work as expected. Probably a bit hacky. If there is a cleaner way to achieve that, that target can be replaced then. For now it seems to do what it should.

Needs to be improved and not to be merged yet.
The temporary file of the environ_aux target will be missing after a `clean` and when running a target, that does not create the temporary env_file..

### DIFF
--- a/Makefile.frankfurt
+++ b/Makefile.frankfurt
@@ -17,8 +17,11 @@ INSTALL_DIRECTORY := .venv
 
 # default configuration
 ENV_FILE ?= .env.default
-SERVER_PORT ?= 6543
-APACHE_PORT ?= 8080
+
+# read all env vars from ENV_FILE and export them all
+include $(ENV_FILE)
+export
+
 # Note the `fi`is a hack for `rm`(which didn't exist for a long time!)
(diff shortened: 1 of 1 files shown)


procrastinatio:
Well, currently it fails, because you are trying to include an non existing file
`make: *** No rule to make target 'temp_env_file.tmp'`

hansmannj:
> Well, currently it fails, because you are trying to include an non existing file `make: *** No rule to make target 'temp_env_file.tmp'`

Yes, if a target is called that does not create that file, it fails. I'll see, how I can fix that.

hansmannj:
> > Well, currently it fails, because you are trying to include an non existing file `make: *** No rule to make target 'temp_env_file.tmp'`
> 
> Yes, if a target is called that does not create that file, it fails. I'll see, how I can fix that.

Should be better now.

procrastinatio (approved):
OK, tested with custom `ENV_FILE`

PR #3914, opened 2022-06-06
BGDIINF_SB-2328: delted docker-entrypoint.sh | deleted obsolete lines from Dockerfile

hansmannj:
Since the pylons config files are now configures in k8s with staging specifig env var substitutions, this task does not longer need to be done via docker-entrypoint.sh, hence that file was deleted. Also the environ Makefile target call from within Dockerfile is no longer needed.


Dockerfile
```diff
@@ -58,7 +58,6 @@ LABEL author=$AUTHOR
 # FIXME: figure out the best way to build chsdi/static/css/extended.min.css (no need to incorporate the whole node.js to build a single CSS file!)
 # FIXME: the flag INSIDE_DOCKER_IMAGE=True is only useful when using Makefile.frankfurt but will be ignored otherwise (i.e. when using Makefiles
 # other than Makefile.frankfurt which won't be the case anyways..)
-RUN make -f ${MAKEFILE} cleanall setup environ fixrights INSIDE_DOCKER_IMAGE=True
+RUN make -f ${MAKEFILE} cleanall setup fixrights INSIDE_DOCKER_IMAGE=True
```
hansmannj, line 61: Not exactly sure, if the `environ` can be deleted here or not. Probably it should stay here, for `make dockerrun` work properly. Just needs to be made sure, that any envsubsts resulting from the `make environ` here will be overriden by k8s in `/base/deploy,yml`.
Probably there is a nicer way that allows to get rid of the `environ` target completely.

hansmannj: For the moment I'd suggest to keep `environ` here (I'll re-add that). I think, a consecutive call of `envsubst` on container start-up in k8s will override the ini files that are generated here (to be checked!). 
That makes it easier to do a `make dockerrun` and have the correct env vars in the ini files.
What do you think about that @procrastinatio?

hansmannj: For the moment I'd suggest to keep `environ` here (I'll re-add that). I think, a consecutive call of `envsubst` on container start-up in k8s will override the ini files that are generated here (to be checked!). 
That makes it easier to do a `make dockerrun` and have the correct env vars in the ini files.
What do you think about that @procrastinatio?

hansmannj: Probably we can have a quick look at the `environ` target together to clean-up a bit.

hansmannj:
Guess we leave the generation of .ini and .conf files in the docker-entrypoint.sh, so I'll close this PR

PR #3917, opened 2022-06-08
BGDIINF_SB-2416 removing the venv from docker image

procrastinatio:


ltshb (approved):
👍🏼

hansmannj (approved):
:+1: thanks!
As Brice suggested, the `INSIDE_DOCKER_IMAGE` flag and logic can be deleted in `Makefile.frankfurt` now

PR #3918, opened 2022-06-08
BGDIINF_SB-2328  Removing `pg-data-replica` temporarly

procrastinatio:
The DB is incomplete and some tests are failing.

Will be changed when tackling with performance.


rebert:
@procrastinatio since dominic copied the db from ireland to frankfurt again - I guess this is obsolete

procrastinatio:
I'm not aware of any changes

PR #3919, opened 2022-06-09
BGDIDIC-645: cast dict element to string before applying startswith

ltclm:
function, this is needed to make it work if dict element is of type None


rebert (approved):
thx

PR #3922, opened 2022-06-13
BGDIINF_SB-2328: Removed git-secrets

ltshb:
For Frankfurt this repository will not contain anymore any information about
the deployment. Therefore the git-secrets is kind of useless.


ltshb:
It might be a good idea to wait that the migration is done before removing git-secrets ?

hansmannj (approved):
:+1:

PR #3923, opened 2022-06-13
BGDIINF_SB-2415: http port cleanup

ltclm:
* PORT clean-up APACHE_PORT / SERVER_PORT -> HTTP_PORT
* inside container we always use default port 80
* ease the access to .env.* file content for make serve and make dockerrun targets
```
ENV_FILE=<my_env_file>> make dockerrun
ENV_FILE=<my_env_file>> make serve
```


Makefile.frankfurt
```diff
@@ -221,10 +228,10 @@ dockerbuild: guard-OPENTRANS_API_KEY guard-PGUSER guard-PGPASSWORD guard-APACHE_
 
 
 .PHONY: dockerrun
-dockerrun: guard-ENV_FILE guard-OPENTRANS_API_KEY guard-PGUSER guard-PGPASSWORD guard-APACHE_PORT
+dockerrun: LoadEnvFile guard-ENV_FILE guard-OPENTRANS_API_KEY guard-PGUSER guard-PGPASSWORD guard-HTTP_PORT
```
hansmannj, line 231: I think `LoadEnvFile` is not needed here, as `--env-file` is passed to `docker run` already.

ltshb (changes requested):
Sorry but a lot of work here is conflicting with #3924. Actually I'm not sure that the port cleanup is a good idea. We need to keep differentiating between the pylons application port with the apache port. Hopefully after the migration we might get totally rid of the apache, so I would let the port clean up as post migration cleanup tasks. We can talk about this in the refinement meetings.

Makefile.frankfurt
ltshb: Right no LoadEnvFile here.

Makefile.frankfurt
```diff
@@ -18,6 +18,13 @@ INSTALL_DIRECTORY := .venv
 # default configuration
 ENV_FILE ?= .env.default
 
+define load_env
+    $(eval ENV := $(1))
+    @echo " - load env $(ENV)"
+    $(eval include $(1))
+    $(eval export)
+endef
```
ltshb, line 26: I'm not an expert in makefile and bash but what are the advantage and disadvantage of this technique in comparison to a more simple way as in #3926 

```bash
environ:
   export $(shell cat $(ENV_FILE)) && \
   ...
```
With your `load_env` does the variable stays after the make ?

Makefile.frankfurt
ltclm: the content of .env file has to be loaded before to substitute HTTP_PORT variable in the docker run command

Makefile.frankfurt
ltclm: thats simpler, i will test it, thank you.

ltclm: @ltshb i have used your approach, less code and simpler, thank you!

ltshb:
As discussed in the planing meeting, we close this as in order to work locally using an ssh local port forwarding for the DB connection, we need to use the `--network=host` mode in docker run which may cause trouble if we use the standard 80 port in apache. Moreover having the possibility to change the port in docker container offer more flexibility for developing.

PR #3924, opened 2022-06-13
BGDIINF_SB-2426: Clean up the docker image

ltshb:


ltshb:
A local image built from this PR has been deployed on DEV, see https://github.com/geoadmin/infra-kubernetes/pull/235

.env.default
```diff
@@ -27,15 +24,13 @@ MODWSGI_USER=www-data
...
-PRINT_SERVER_HOST=service-print.dev.bgdi.ch
+PRINT_SERVER_HOST=sys-print.dev.bgdi.ch
 PRINT_TEMP_DIR=/var/local/print
 PUBLIC_BUCKET_HOST=sys-public.dev.bgdi.ch
 RANCHER_LABEL=dev
 ROBOTS_FILE=robots_prod.txt
 SERVER_PORT=6543
-SPHINXHOST=service-sphinxsearch.dev.bgdi.ch
-TOMCAT_PORT=8011
-TOMCAT_SERVER_URL=//vpc-alb-bgdi-dev.intra.bgdi.ch
```
ltclm, line 38: WSGI_PORT can be removed too

Dockerfile
```diff
@@ -1,48 +1,64 @@
...
+        proxy_http \
+        proxy_http2 \
+        rewrite \
+        setenvif \
+        status \
+        wsgi \
+        alias
 
-COPY . /var/www/vhosts/${VHOST}/private/chsdi
+COPY --chown=${USER}:${GROUP} . /var/www/vhosts/${VHOST}/private/chsdi
```
ltclm, line 56: just out of interest, is it planned to start the root process in this container as geodata or why do we change the ownership here and add this user? i personally would think it is a good idea to not run the container as root and change the user in the end of the dockerfile or in the entrypoint:
https://github.com/geoadmin/service-sphinxsearch/blob/develop-docker/scripts/docker-entry.sh

ltclm (approved):
thank you @ltshb

Dockerfile
ltshb: To be honest I'm not sure anymore of the reason here. But I'll create a task to review the whole user and permissions in Docker. But this task can be done as post migration cleanup.

PR #3925, opened 2022-06-13
BGDIINF_SB-2445: Python3 support bug fix

ltshb:
Python3 doesn't have urllib2 but urllib.

**TO BE MERGED AFTER #3924**

PR #3926, opened 2022-06-13
BGDIINF_SB-2446: Fixed local run with Makefile.frankfurt

ltshb:
Make sure that the correct environment are exported when running locally. Also
reduced the DB connection timeout to get an error rather sooner when the DB
connection is not possible.

**TO BE MERGED AFTER https://github.com/geoadmin/mf-chsdi3/pull/3925**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,7 @@
 *.swp
 *~
 *.orig
-.coverage
+.coverage*
 .idea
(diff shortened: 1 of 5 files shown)


ltshb:
TO BE MERGED AFTER #3925

hansmannj (approved):
:+1: thanks!

PR #3928, opened 2022-06-14
BGDIINF_SB-2447: No more auth for info.json

rebert:


ltshb (approved):
👍🏼 Maybe add a comment in the https://github.com/geoadmin/mf-chsdi3/wiki/TODO-Migration about removing the info.json.mako file after the migration

PR #3933, opened 2022-06-16
Improved building target dependencies

ltshb:
Now the translations and doc are only built when needed.

Also reorganized the makefile to group all the PHONY targets as well as all
the PHONY build targets. This make the navigation in makefile easier.

### DIFF
--- a/Makefile.frankfurt
+++ b/Makefile.frankfurt
@@ -13,7 +13,7 @@ SHELL = /bin/bash
 SERVICE_NAME := mf-chsdi3
 
 CURRENT_DIRECTORY := $(shell pwd)
-INSTALL_DIRECTORY := .venv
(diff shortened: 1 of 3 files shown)


procrastinatio (approved):
OK.
Maybe add a small doc to states that the all goal of this `.po/.mo`machanic is to create a `gettext` `domain` `chsdi`. The translations themselves are stored in the DB and refreshed at given intervalles (see https://github.com/geoadmin/mf-chsdi3/blob/a0f0473dd2700965ae585df4a3714a0b81a0faa8/chsdi/subscribers.py#L30-L67)

hansmannj (approved):
:+1: Thanks!

ltclm (approved):
thank you 👍

PR #3934, opened 2022-06-16
BGDIINF_SB_2418: added short comment in .env.default

hansmannj:
`KML_TEMP_DIR` env var in .env.default can probably be deleted in context of
https://jira.swisstopo.ch/browse/BGDIINF_SB-2456. Hence a short comment has been added
to .env.default


procrastinatio (approved):
OK.

PR #3936, opened 2022-06-17
BGDIINF_SB-2451: make code compatible with Python2.7 and Python3.7 at the same time

hansmannj:
For a certain period, the legacy code has to be compatible with both, Python2.7 and 3.7+
2to3-2.7 has been used to identify parts of the code, that needed to be changed. Six has been used to make those
parts compatible with both versions.

- [x] Check and make python modules compatible with both Python versions
- [x] Check and make Mako files compatible with both Python versions
- [x] Check if all `list()` wrapping suggested by `2to3` is really needed
- [x] Check what is Python version specific at code around line 312 in `/chsdi/views/features.py` (there's a `Python 2/3` comment)

### DIFF
--- a/chsdi/lib/filters.py
+++ b/chsdi/lib/filters.py
@@ -11,6 +11,7 @@ def full_text_search(query, ormColumns, searchText):
     def ilike_search(col):
         if col is not None:
             return cast(col, Text).ilike('%%%s%%' % searchText)
+    # TODO: clean-up when only Python 3.x and no longer 2.x is in use
     return query.filter(or_(
         *map(ilike_search, ormColumns)
     ))
--- a/chsdi/lib/filters.py
+++ b/chsdi/lib/filters.py
@@ -11,7 +11,6 @@ def full_text_search(query, ormColumns, searchText):
     def ilike_search(col):
         if col is not None:
             return cast(col, Text).ilike('%%%s%%' % searchText)
-    # TODO: clean-up when only Python 3.x and no longer 2.x is in use
     return query.filter(or_(
         *map(ilike_search, ormColumns)
     ))
--- a/chsdi/lib/helpers.py
+++ b/chsdi/lib/helpers.py
@@ -10,11 +10,6 @@ import unidecode
 from decimal import Decimal
 from past.utils import old_div
 
-try:
-    from StringIO import StringIO
-except ImportError:
-    from io import StringIO
-
 from six.moves import zip, reduce, zip_longest
 from itertools import chain
 
@@ -37,7 +32,11 @@ except ImportError:
 import xml.etree.ElementTree as etree
 from pyproj import Proj, transform as proj_transform
 from requests.exceptions import ConnectionError, Timeout, RequestException
-from requests.packages.urllib3.exceptions import InsecureRequestWarning
+# TODO: clean-up when only Python 3.x and no longer 2.x is in use
+try:
+    from requests.packages.urllib3.exceptions import InsecureRequestWarning
+except ImportError:
+    from urllib3.exceptions import InsecureRequestWarning
 from shapely.ops import transform as shape_transform
 from shapely.wkt import dumps as shape_dumps, loads as shape_loads
 from shapely.geometry.base import BaseGeometry
@@ -460,17 +459,12 @@ def float_raise_nan(val):
 
 def parse_box2d(stringBox2D):
     extent = stringBox2D.replace('BOX(', '').replace(')', '').replace(',', ' ')
-    # Python2/3
-    box = map(float, extent.split(' '))
-    if not isinstance(box, list):
-        box = list(box)
+    # TODO: clean-up when only Python 3.x and no longer 2.x is in use
+    box = list(map(float, extent.split(' ')))
     return box
 
 
-def is_box2d(box2D):
-    # Python2/3
-    if not isinstance(box2D, list):
-        box2D = list(box2D)
+def validate_box2d(box2D):
     # Bottom left to top right only
     if len(box2D) != 4 or box2D[0] > box2D[2] or box2D[1] > box2D[3]:
         raise ValueError('Invalid box2D.')
@@ -478,7 +472,7 @@ def is_box2d(box2D):
 
 
 def center_from_box2d(box2D):
-    box2D = is_box2d(box2D)
+    box2D = validate_box2d(box2D)
     return [
         box2D[0] + ((box2D[2] - box2D[0]) / 2),
         box2D[1] + ((box2D[3] - box2D[1]) / 2)
@@ -537,7 +531,7 @@ def format_scale(scale):
     scale_str = str(scale)
     n = ''
     while len(scale_str) > 3:
-        # Python2/3
+        # TODO: clean-up when only Python 3.x and no longer 2.x is in use
         scale_prov = old_div(int(float(scale_str)), 1000)
         n = n + "'000"
         scale_str = str(scale_prov)
@@ -563,7 +557,7 @@ def get_loaderjs_url(request, version='3.6.0'):
 
 def decompress_gzipped_string(streaming_body):
     if six.PY2:
-        string_file = StringIO(streaming_body.read())
+        string_file = six.StringIO(streaming_body.read())
         gzip_file = gzip.GzipFile(fileobj=string_file, mode='r', compresslevel=5)
         return gzip_file.read().decode('utf-8')
     else:
--- a/chsdi/lib/helpers.py
+++ b/chsdi/lib/helpers.py
@@ -459,7 +459,6 @@ def float_raise_nan(val):
 
 def parse_box2d(stringBox2D):
     extent = stringBox2D.replace('BOX(', '').replace(')', '').replace(',', ' ')
-    # TODO: clean-up when only Python 3.x and no longer 2.x is in use
     box = list(map(float, extent.split(' ')))
     return box
 
@@ -556,6 +555,7 @@ def get_loaderjs_url(request, version='3.6.0'):
 
 
 def decompress_gzipped_string(streaming_body):
+    # TODO: clean-up when only Python 3.x and no longer 2.x is in use
     if six.PY2:
         string_file = six.StringIO(streaming_body.read())
         gzip_file = gzip.GzipFile(fileobj=string_file, mode='r', compresslevel=5)
--- a/chsdi/lib/mortonspacekey.py
+++ b/chsdi/lib/mortonspacekey.py
@@ -179,9 +179,6 @@ class QuadTree:
 
     def _getCommonKey(self, keys):
         res = ''
-        # Python2/3
-        if not isinstance(keys, list):
-            keys = list(keys)
         for i in range(self.levels + 1):
             before = len(res)
             if not reduce(lambda has, k: (has and len(k) > i), keys, True):
--- a/chsdi/models/vector/__init__.py
+++ b/chsdi/models/vector/__init__.py
@@ -28,9 +28,6 @@ DEFAULT_OUPUT_SRID = 21781
 
 
 def get_resolution(imageDisplay, mapExtent):
-    # TODO: Python2/3
-    if not isinstance(imageDisplay, list):
-        imageDisplay = list(imageDisplay)
     bounds = mapExtent.bounds
     map_meter_width = abs(bounds[0] - bounds[2])
     map_meter_height = abs(bounds[1] - bounds[3])
@@ -40,9 +37,6 @@ def get_resolution(imageDisplay, mapExtent):
 
 
 def get_scale(imageDisplay, mapExtent):
-    # TODO: Python2/3
-    if not isinstance(imageDisplay, list):
-        imageDisplay = list(imageDisplay)
     resolution = get_resolution(imageDisplay, mapExtent)
     return resolution * 39.37 * imageDisplay[2]
 
--- a/chsdi/templates/luftbilder/viewer.mako
+++ b/chsdi/templates/luftbilder/viewer.mako
@@ -1,14 +1,13 @@
 <%namespace name="lubis_map" file="lubis_map.mako"/>
 
 <%
-from pyramid.url import route_url
-# TODO: clean-up when only Python 3.x and no longer 2.x are in use
-try:
-    from urllib2 import urlopen
-except ImportError:
-    # Python3 fallback
-    from urllib.request import urlopen
-from json import loads
+  from pyramid.url import route_url
+  # TODO: clean-up when only Python 3.x and no longer 2.x are in use
+  try:
+      from urllib2 import urlopen
+  except ImportError:
+      from urllib.request import urlopen
+  from json import loads
 
   c = context
   topic = 'luftbilder'
--- a/chsdi/templates/luftbilder/viewer.mako
+++ b/chsdi/templates/luftbilder/viewer.mako
@@ -1,13 +1,14 @@
 <%namespace name="lubis_map" file="lubis_map.mako"/>
 
 <%
-  from pyramid.url import route_url
-  try:
-      from urllib2 import urlopen
-  except ImportError:
-      # Python3 fallback
-      from urllib.request import urlopen
-  from json import loads
+from pyramid.url import route_url
+# TODO: clean-up when only Python 3.x and no longer 2.x are in use
+try:
+    from urllib2 import urlopen
+except ImportError:
+    # Python3 fallback
+    from urllib.request import urlopen
+from json import loads
 
   c = context
   topic = 'luftbilder'
--- a/chsdi/views/layers.py
+++ b/chsdi/views/layers.py
@@ -49,9 +49,7 @@ def layers_config(request):
     params = BaseLayersValidation(request)
     query = params.request.db.query(LayersConfig)
     layers = {}
-    # TODO: clean-up when only Python 3.x and no longer 2.x is in use
     for layer in get_layers_config_for_params(params, query, LayersConfig):
-        # layers = dict(list(layers.items()) + list(layer.items()))
         layers = dict(chain(layers.items(), layer.items()))
 
     return layers
@@ -170,7 +168,6 @@ def faqlist(request):
 
     query = params.request.db.query(LayersConfig)
     for layer in get_layers_config_for_params(params, query, LayersConfig):
-        # TODO: clean-up when only Python 3.x and no longer 2.x is in use
         k = list(layer.keys()).pop()
         lyr = list(layer.values()).pop()
         if 'parentLayerId' not in lyr and not k.endswith('_3d'):
--- a/chsdi/views/layers.py
+++ b/chsdi/views/layers.py
@@ -49,7 +49,7 @@ def layers_config(request):
     params = BaseLayersValidation(request)
     query = params.request.db.query(LayersConfig)
     layers = {}
-    # Python 2/3
+    # TODO: clean-up when only Python 3.x and no longer 2.x is in use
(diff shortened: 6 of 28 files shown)


chsdi/lib/validation/__init__.py
```diff
@@ -61,7 +61,7 @@ def srid(self):
 
     @srid.setter
     def srid(self, value):
-        if value in map(str, SUPPORTED_OUTPUT_SRS):
+        if value in list(map(str, SUPPORTED_OUTPUT_SRS)):
```
ltshb, line 64: `list` here is not needed and is bad. `map` in python3 returns an iterator that is more efficient than a list when used with a generator.

chsdi/lib/helpers.py
```diff
@@ -460,15 +495,13 @@ def float_raise_nan(val):
...
-    if not isinstance(box, list):
-        box = list(box)
+    # TODO: clean-up when only Python 3.x and no longer 2.x is in use
+    box = list(map(float, extent.split(' ')))
     return box
 
 
 def is_box2d(box2D):
-    # Python2/3
+    # TODO: clean-up when only Python 3.x and no longer 2.x is in use
```
ltshb, line 504: Looking at the code, the `if not isinstance(box2D, list)` don't make sense at all here. Moreover having a method called `is_` which return the expected value instead of `True/False` is not familiar and error prone, I mean, if you ask `is this a bbox 2D ?` you expect `yes` or `no` as answer and not the `bbox 2D`. 
So here if we already touch this code, I'll remove the `TODO comment` and remove the `if not isinstance()` which is unnecessary as every caller pass a list except for one call in https://github.com/geoadmin/mf-chsdi3/blob/fb968a4668835cc2eb90211e2683990b1ba20082/chsdi/views/search.py#L363 where the `self.bbox` is only set here https://github.com/geoadmin/mf-chsdi3/blob/fb968a4668835cc2eb90211e2683990b1ba20082/chsdi/views/search.py#L36 and where it should be a string, so in this case it will anyway raise an error. I guess this code is never reached. Anyway this code will be removed in Frankfurt as the search service has been migrated in service-search-wsgi.
Finally I would rename the function to `validate_box2d()`:
```python
def validate_box2d(box2D):
    # Bottom left to top right only
    if len(box2D) != 4 or box2D[0] > box2D[2] or box2D[1] > box2D[3]:
        raise ValueError('Invalid box2D.')
    return box2D
```

chsdi/lib/mortonspacekey.py
```diff
@@ -171,41 +164,47 @@ def bbox_to_morton(self, bbox):
...
     get key for each corner point of bounding box
     return key which is common to all results (from the start)
     '''
 
     def _getCommonKey(self, keys):
         res = ''
-        # Python2/3
+        # TODO: clean-up when only Python 3.x and no longer 2.x is in use
         if not isinstance(keys, list):
             keys = list(keys)
```
ltshb, line 176: Comment and `if not isintsance(...)` are useless, if you look at the code the function is always called with a list as parameter. So remove the comment and `if`

chsdi/lib/validation/__init__.py
```diff
@@ -61,6 +61,7 @@ def srid(self):
 
     @srid.setter
     def srid(self, value):
+        # 2to3 would wrap the map() inside a list(), which is not necessary here.
```
ltshb, line 64: Useless comment, it should be remove. In very near future we don't care anymore in python2.

chsdi/lib/validation/find.py
```diff
@@ -102,6 +102,7 @@ def returnGeometry(self, value):
 
     @srid.setter
     def srid(self, value):
+        # 2to3 would wrap the map() inside a list(), which is not necessary here.
```
ltshb, line 105: Dito remove comment

chsdi/lib/validation/find.py
```diff
@@ -121,7 +122,10 @@ def layerDefs(self, value):
             try:
                 defs = json.loads(urllib.parse.unquote(value))
                 if not (set(defs.keys()).issubset(set([self.layer]))):
-                    raise HTTPBadRequest("You can only filter on layer '%s' in 'layerDefs'" % self.layer)
+                    raise HTTPBadRequest(
+                        "You can only filter on layer '%s' in 'layerDefs'" %
+                        self.layer)
+                # 2to3 would wrap defs.values() in a list(), which is not necessary in this case
```
ltshb, line 128: Dito remove comment

chsdi/lib/validation/identify.py
```diff
@@ -125,7 +125,10 @@ def layerDefs(self, value):
                 defs = json.loads(value)
                 if self.layers != 'all':
                     if not (set(defs.keys()).issubset(set(self.layers))):
-                        raise HTTPBadRequest("You can only filter on layer '%s' in 'layerDefs'" % self.layers)
+                        raise HTTPBadRequest(
+                            "You can only filter on layer '%s' in 'layerDefs'"
+                            % self.layers)
+                # 2to3 would wrap defs.values() in a list(), which is not necessary in this case
```
ltshb, line 131: dito remove comment

chsdi/models/bod.py
```diff
@@ -52,6 +52,7 @@ def layerMetadata(self):
         primaryAttrs = ('layerBodId', 'idGeoCat', 'name', 'fullName')
         excludedAttrs = ('staging', 'searchUnicode', 'chargeable')
         meta = {'attributes': {}}
+        # 2to3 would wrap self.__dict__.keys() in a list(), which is not necessary in this case
```
ltshb, line 55: dito remove comment

chsdi/models/bod.py
```diff
@@ -106,6 +107,7 @@ def layerConfig(self, params):
         settings = params.request.registry.settings
         wmsHost = settings['wmshost']
         defaultResolution = 0.5
+        # 2to3 would wrap self.__dict__.keys() in a list(), which is not necessary in this case
```
ltshb, line 110: dito remove comment

chsdi/models/bod.py
```diff
@@ -404,7 +406,7 @@ class Catalog(Base):
     def to_dict(self, lang):
 
         self.label = self._get_label_from_lang(lang)
-
+        # 2to3 would wrap self.__dict__.keys() in a list(), which is not necessary in this case
```
ltshb, line 409: Dito remove comment

chsdi/models/vector/__init__.py
```diff
@@ -21,14 +21,13 @@
...
 MAX_FEATURE_GEOMETRY_SIZE = 1e6
 
 DEFAULT_OUPUT_SRID = 21781
 
 
 def get_resolution(imageDisplay, mapExtent):
-    # TODO: Python2/3
+    # TODO: clean-up when only Python 3.x and no longer 2.x is in use
     if not isinstance(imageDisplay, list):
         imageDisplay = list(imageDisplay)
```
ltshb, line 32: Here also the comment and `if not isinstance()` are useless, analyzing the code show that `imageDisplay` is always a list.

chsdi/models/vector/__init__.py
```diff
@@ -40,7 +39,7 @@ def get_resolution(imageDisplay, mapExtent):
 
 
 def get_scale(imageDisplay, mapExtent):
-    # TODO: Python2/3
+    # TODO: clean-up when only Python 3.x and no longer 2.x is in use
     if not isinstance(imageDisplay, list):
         imageDisplay = list(imageDisplay)
```
ltshb, line 44: Dito as above in `get_resolution`

chsdi/templates/luftbilder/viewer.mako
```diff
@@ -1,13 +1,10 @@
...
-  from pyramid.url import route_url
-  try:
-      from urllib2 import urlopen
-  except ImportError:
-      # Python3 fallback
-      from urllib.request import urlopen
   from json import loads
+  from pyramid.url import route_url
+  # TODO: clean-up when only Python 3.x and no longer 2.x are in use
+  from six.moves.urllib.request import urlopen
```
ltshb, line 7: The previous code was easier to clean up after the migration ! It would have just required to remove some lines of code. Now we will need to search for the python3 library that implement `urlopen`, so IMHO the previous code should be left as is, just add the `TODO` comment.

chsdi/views/catalog.py
```diff
@@ -30,13 +30,15 @@ def tree_data(G, root, attrs, meta):
...
-        label_key = lambda x: x['label'].replace(u'Ü', u'U').replace(u'Ä', u'A')
+        label_key = lambda x: x['label'].replace(u'Ü', u'U').replace(
+            u'Ä', u'A')
         nbrs = G[n]
         if len(nbrs) == 0:
             return []
         children_ = []
         nbrs = sorted(nbrs)
         for child in nbrs:
+            # 2to3 would wrap G.node[child].items() inside a list(), which is not necessary here.
```
ltshb, line 41: Dito remove useless comment.

chsdi/views/catalog.py
```diff
@@ -66,6 +68,7 @@ def add_children(n, G):
             d.pop('orderKey', None)
         return children_
 
+    # 2to3 would wrap G.node[child].items() inside a list(), which is not necessary here.
```
ltshb, line 71: Dito remove useless comment.

ltshb (changes requested):
Lot of formatting changes ! Make the review very hard, and you will have a lots of rebasing/merging issue with so many formatting ! This cause also trouble when the `develop-` branch will be merged !
In general we SHOULD always put formatting in separate PR (or at least in separate commit).
Due to the huge diff due to formatting, I'm not sure that I grasped all the python3 compatibility issues, but I guess there are just a few ? So I would redo it without the formatting plus considering the comment.

hansmannj:
> Lot of formatting changes ! Make the review very hard, and you will have a lots of rebasing/merging issue with so many formatting ! This cause also trouble when the `develop-` branch will be merged ! In general we SHOULD always put formatting in separate PR (or at least in separate commit). Due to the huge diff due to formatting, I'm not sure that I grasped all the python3 compatibility issues, but I guess there are just a few ? So I would redo it without the formatting plus considering the comment.

Yes, this tons of formatting thins are annoying. Probably I ran the `make autolint` target once. At least I did not manually introduce all those :wink: I'll revert them manually now, though :grin:

ltclm:
or it has been done by vscode, in that case we should disable auto-format option in the vscode settings.

chsdi/lib/helpers.py
hansmannj: Yes, there are some examples where there were already some `list()` conversions in the code, that might not be needed. But I rather left them, as they don't cause any harm, rather than probably overlooking some call, for which the conversion would make sense.
I could do the changes now, but this would take some more time and be slightly risky.
So I would suggest, to do this in context of post-migration clean-up, when switching to Python3 completely.

procrastinatio (commented):
I think the PR title is quite misleading. Should be ' reformatting code'. Porting the code to Python3 was much much more work.

chsdi/templates/luftbilder/viewer.mako
hansmannj: IMHO this is even simpler. We only need to remove `six.moves` from `six.moves.urllib.request`, so it will be `from urllib.request import urlopen` aftwards. The TODO comment is already in place.

chsdi/lib/validation/__init__.py
hansmannj: all those comments are now removed

chsdi/lib/mortonspacekey.py
```diff
@@ -78,8 +72,8 @@ def getIntersection(self, other):
     def create_quads(self):
         newWidth = self.width() / 2
         newHeight = self.height() / 2
-        quad0 = BBox(self.minx, self.miny + newHeight,
-                     self.minx + newWidth, self.maxy)
+        quad0 = BBox(self.minx, self.miny + newHeight, self.minx + newWidth,
```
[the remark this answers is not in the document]
hansmannj: Should be reverted now, as I reverted the auto-formatted changes

chsdi/lib/mortonspacekey.py
```diff
@@ -50,16 +46,14 @@ def pointAt(self, i):
...
+                and self.maxy >= y):
             return True
         return False
 
     def getIntersection(self, other):
-        if (other.maxx < self.minx or
-                other.minx > self.maxx or
-                other.maxy < self.miny or
-                other.miny > self.maxy):
+        if (other.maxx < self.minx or other.minx > self.maxx
```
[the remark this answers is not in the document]
hansmannj: should be reverted, as I reverted the changes introduced by auto-formatting

chsdi/lib/filters.py
```diff
@@ -11,8 +11,9 @@ def full_text_search(query, ormColumns, searchText):
     def ilike_search(col):
         if col is not None:
             return cast(col, Text).ilike('%%%s%%' % searchText)
+    # TODO: clean-up when only Python 3.x and no longer 2.x is in use
     return query.filter(or_(
-        *map(ilike_search, ormColumns)
+        *list(map(ilike_search, ormColumns))
```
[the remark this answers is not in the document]
hansmannj: reverted it

chsdi/models/vector/__init__.py
hansmannj: Not sure about that. For example in the `cadastralwebmap_opendata.mako` the default value for `imageDisplay` is a string ('400,600,96'). Hence I'd suggest to leave the `ìsinstance()` here.

chsdi/models/vector/__init__.py
hansmannj: as above

chsdi/models/vector/__init__.py
ltshb: No in `cadastralwebmap_opendata.mako` it is a list as well, you have a `setter` for `imageDisplay` that parse and translate strings into list, see https://github.com/geoadmin/mf-chsdi3/blob/ff57e114f3827716c280c990e1fcbdab77dac646/chsdi/lib/validation/identify.py#L175 and https://github.com/geoadmin/mf-chsdi3/blob/ff57e114f3827716c280c990e1fcbdab77dac646/chsdi/lib/validation/identify.py#L79
IMHO this should be removed now has this is not needed for python2 and `isinstance` is not cheap.

chsdi/models/vector/__init__.py
ltshb: Dito as above

chsdi/templates/luftbilder/viewer.mako
ltshb: This is true if you know out of the head that in python3 it is `from urllib.request import urlopen` 😉 otherwise not simpler as you have to search for the correct import in python3 first, while the other code is clear from the code what import is fort python2 and what for python3.

tests/integration/test_features_service.py
```diff
@@ -340,7 +340,7 @@ def test_feature_valid(self):
 
     def test_feature_too_many_featuresids(self):
         bodId = 'ch.bafu.bundesinventare-bln'
-        featureIds = ','.join(map(str, range(50)))
+        featureIds = ','.join(map(str, list(range(50))))
```
ltshb, line 343: `list` is not needed and inefficient here, `map` method require an iterator.

tests/integration/test_search.py
```diff
@@ -446,7 +446,7 @@ def test_search_locations_no_geometry(self):
         resp = self.testapp.get('/rest/services/ech/SearchServer', params=params, status=200)
         self.assertEqual(resp.content_type, 'application/json')
         self.assertGreater(len(list(resp.json['results'])), 0)
-        self.assertTrue('geom_st_box2d' not in resp.json['results'][0]['attrs'].keys())
+        self.assertTrue('geom_st_box2d' not in list(resp.json['results'][0]['attrs'].keys()))
```
ltshb, line 449: Here also `list` is not required and inefficient, `not in` works with iterator.

chsdi/lib/helpers.py
ltshb: Sorry I don't see any risk here, the code is pretty clear and simple, and if you don't do it now within this PR the first analysis done here will be very likely forgotten.

ltshb (changes requested):
The PR should be also rebased on master.

chsdi/models/bod.py
hansmannj: A thought on all those comments: Probably they do make sense. When someone in future wants to port the code to Python3 and uses 2to3 all those things will pop up. And in places, where I added this comment, I already tested it and checked, that the suggested changes of 2to3 are not needed. So probably re-adding those comments make sense.

chsdi/lib/helpers.py
hansmannj: Done

chsdi/models/vector/__init__.py
hansmannj: done

chsdi/lib/filters.py
```diff
@@ -11,6 +11,7 @@ def full_text_search(query, ormColumns, searchText):
     def ilike_search(col):
         if col is not None:
             return cast(col, Text).ilike('%%%s%%' % searchText)
+    # TODO: clean-up when only Python 3.x and no longer 2.x is in use
```
ltshb, line 14: Probably this comment doesn't make sense here ? If I'm not missing something the code is compatible for both python2 and python3 and there is nothing to cleanup.

chsdi/lib/helpers.py
```diff
@@ -563,7 +557,7 @@ def get_loaderjs_url(request, version='3.6.0'):
 
 def decompress_gzipped_string(streaming_body):
     if six.PY2:
-        string_file = StringIO(streaming_body.read())
+        string_file = six.StringIO(streaming_body.read())
```
ltshb, line 560: Here a TODO is missing.

chsdi/views/layers.py
```diff
@@ -49,7 +49,7 @@ def layers_config(request):
     params = BaseLayersValidation(request)
     query = params.request.db.query(LayersConfig)
     layers = {}
-    # Python 2/3
+    # TODO: clean-up when only Python 3.x and no longer 2.x is in use
     for layer in get_layers_config_for_params(params, query, LayersConfig):
         # layers = dict(list(layers.items()) + list(layer.items()))
```
ltshb, line 54: I think here as well there is nothing to clean up later instead of removing the commented line no 54 that could be done now.

chsdi/views/layers.py
```diff
@@ -170,7 +170,7 @@ def faqlist(request):
 
     query = params.request.db.query(LayersConfig)
     for layer in get_layers_config_for_params(params, query, LayersConfig):
-        # Python2/3
+        # TODO: clean-up when only Python 3.x and no longer 2.x is in use
         k = list(layer.keys()).pop()
         lyr = list(layer.values()).pop()
```
ltshb, line 175: There is nothing to clean up for python3 only support here.

chsdi/views/search.py
```diff
@@ -490,7 +490,8 @@ def _box2d_transform(self, res):
         try:
             box2d = res['geom_st_box2d']
             box_str = box2d[4:-1]
-            b = map(float, re.split(' |,', box_str))
+            # TODO: clean-up when only Python 3.x and no longer 2.x is in use
```
ltshb, line 493: Same here for py3 only the code cannot be changed. So comment can be removed.

ltshb (approved):
Looks good, I would just remove some useless comment.

PR #3938, opened 2022-06-20
BGDIINF_SB-2423: app_version

rebert:


Dockerfile
```diff
@@ -14,9 +14,9 @@ ENV GROUP geodata
 RUN apt-get update -qq \
     && DEBIAN_FRONTEND=noninteractive apt-get install -qq -y --upgrade ca-certificates \
     && DEBIAN_FRONTEND=noninteractive apt-get install -qq -y  \
-        libgeos-dev \
-        gettext-base \
-        apache2 libapache2-mod-wsgi-py3 \
+    libgeos-dev \
```
ltshb, line 17: Oops some unwanted whitespaces changes, probably done via vscode, see https://github.com/geoadmin/tool-dev-setup/pull/114

ltshb (changes requested):
Good just revert the white spaces in dockerfile. Maybe `base_version.ini` is more appropriate ?

procrastinatio (approved):
OK, but I'm not sure it has any real signification as the docker images have all kinds of tags.

hansmannj (approved):
:+1: Thanks!

Dockerfile
rebert: as the variable is called app_version all over, I would like to keep it app_version

ltshb (approved):
👍🏼

PR #3939, opened 2022-06-20
BGDIINF_SB-2459: Uses JSON logs output format in python application

ltshb:
Requires https://github.com/geoadmin/lib-py-logging-utilities/pull/26 to be merged and published.

### DIFF
--- a/25-mf-chsdi3.conf.in
+++ b/25-mf-chsdi3.conf.in
@@ -3,6 +3,20 @@
 # Managed by Puppet
 # ************************************
 
+# # Use JSON format for logs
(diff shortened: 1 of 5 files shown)


ltshb:
Unfortunately we cannot use JSON logs output as all logs are delivered by apache and that we cannot configure a consistent a valid JSON output in apache. I tried to hack the apache config to have a JSON like output, but the main issue is that the `%M` which returns either the message from the pylons application or from an apache message won't escape non json characters, like `"` in strings. 
So instead of using JSON a new PR has been open to improve the logging, see #3942 .s

PR #3941, opened 2022-06-22
Speed up dockerbuild process

ltshb:
Python package dependencies installation is quite slow and was done for every
build increment which was useless if the dependencies did not changed.

Now the dependencies are installed in a separate layer at the beginning of the
Dockerfile, which allow to cache this layer and wont rebuild it unless the
requirements-py3.txt file changes.

This speed up subsequent build quite a lot.


rebert (approved):

rebert:
nice

PR #3943, opened 2022-06-22
BGDIINF_SB-2423: Improved info.json build process and added git dirty to the version

ltshb:
When building a docker image from a dirty repository based on a tag, the version
was set to the tag ! Now if the repo is dirty a `-dirty` flag is appended to
the version.
The output of `git status --porcelain` is also added to the info.json.

I moved the info.json.in to its output folder in order to keep the root folder
as clean as possible.

Instead of using a intermediate ini file to setup the `app_version` we use
the feature of `envsubst` to only substitute one variable. This is done then
in Dockerfile with the help of `tee` command as `envsubst` cannot do in place
substitution.


rebert (approved):
better! thanks

PR #3951, opened 2022-06-27
Removed useless ENV variable and targets in Makefile

ltshb:
The APP_VERSION is set at compile time (see Dockerfile)

Also removed the docker target in old makefile to avoid confusion


rebert (approved):
thanks @ltshb. and thanks for the cleanup in the Makefile (Ireland)

PR #3957, opened 2022-06-27
BGDIINF_SB-2478: Temporarely fixed versioned cache prefix for static files

ltshb:
The static files are called by the makos using a version prefix. This prefix
depends on the `app_version` setting which is set to the GIT_TAG value during
build. For mf-chsdi3 the GIT_TAG is a milestone tag which is usually YYYY-MM-DD-rcX.
But we need to also consider unofficial version which have a suffix added to
the official version. This suffix is usually the git short hash of the last
commit, but can also contain a -dirty suffix.


rebert (approved):
great

PR #3958, opened 2022-06-27
Addition of releasenote announcements

mariokeusen:


consta-v (commented):
The announcements look good to me

b4l (approved):

PR #3960, opened 2022-06-28
BGDIINF_SB-2458: Removed HTTP_PROXY

ltshb:
This proxy settings were apparently needed for E2E Varnish tests, but with
the Frankfurt setup we don't use anymore Varnish and therefore this is not
needed anymore.


tests/e2e/__init__.py
```diff
@@ -23,9 +22,6 @@ def setUp(self):
         self.GOOD_REFERER = 'http://unittest.geo.admin.ch'
```
procrastinatio, line 22: Do we still use referrer somewhere? If not, may an additionnal cleanup PR

.env.default
```diff
@@ -17,7 +17,6 @@ DYNAMIC_TRANSLATION=1
 GEOADMINHOST=mf-geoadmin3.dev.bgdi.ch
 GEODATA_STAGING=test
 HOST=sys-api3.dev.bgdi.ch
-HTTP_PROXY=http://ec2-52-28-118-239.eu-central-1.compute.amazonaws.com:80
```
procrastinatio, line 20: Check with the `infra` if the host still does exist? If so, we may considere ditching it.

procrastinatio (approved):
OK. Maybe some additional steps not to forget.

.env.default
ltshb: Good point @ltrod @samfriedli do you know if this host `http://ec2-52-28-118-239.eu-central-1.compute.amazonaws.com` still exists ?

ltrod: Yes this host still exists. The instance is located the aws-admin account in Frankfurt (the only instance in Frankfurt in the aws-admin account)

tests/e2e/__init__.py
ltclm: yes, i think the GOOD_REFERER, BAD_REFERER can be removed too.

ltclm (approved):

.env.default
ltshb: Ok It might be a good idea to have a reminder to clean up this instance, but I guess at best after the migration. I'll created a jira ticket for this task, see https://jira.swisstopo.ch/browse/IGI_SB-1531

PR #3969, opened 2022-07-06
Try to solve strange issue with empty base.ini.in and push master.latest docker image tag.

ltshb:
One docker image built by the CI had an empty base.ini.in file. Checking the logs of the CI build did not revealed any issues that could have resulted to an empty base.ini.in file. However this file was built with `envsubst` and `tee` program using a pipe and `>/dev/null` so some error could have been missing. A CI build retry solved the issue.
    
So here I'm trying to do the envsubst with `sed -i` command with the hope that if this for any reason would fail, it would provide some useful information and stop the build process.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -69,15 +69,15 @@ ARG GIT_DIRTY=unknown
 ARG VERSION=unknown
 ARG AUTHOR=unknown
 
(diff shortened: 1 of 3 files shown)


ltclm (approved):
🤞

PR #3971, opened 2022-07-13
BGDIINF_SB-2505: fix transformation of  tolerance

ltclm:
tolerance will always be transformed into meters

pixel -> wgs84 -> DEFAULT_OUPUT_SRID (currently lv03)

[testlink](http://mf-chsdi3.dev.bgdi.ch/fix_BGDIINF_SB-2505_tolerance/rest/services/all/MapServer/identify?geometry=7.451798841795024,46.9283737686916&geometryType=esriGeometryPoint&geometryFormat=geojson&sr=4326&tolerance=10&imageDisplay=1695,643,96&mapExtent=46.92802164059145,7.450493425073449,46.92859979856202,7.452719182911476&layers=all:ch.bfs.gebaeude_wohnungs_register)

### DIFF
--- a/chsdi/models/vector/__init__.py
+++ b/chsdi/models/vector/__init__.py
@@ -18,13 +18,14 @@ from geoalchemy2.shape import to_shape
 if six.PY3:
     buffer = memoryview
 
-Geometry2D = GeometryChsdi(geometry_type='GEOMETRY', dimension=2, srid=2056)
-Geometry3D = GeometryChsdi(geometry_type='GEOMETRYZ', dimension=3, srid=2056)
+DEFAULT_DB_SRID = 2056
+Geometry2D = GeometryChsdi(geometry_type='GEOMETRY', dimension=2, srid=DEFAULT_DB_SRID)
+Geometry3D = GeometryChsdi(geometry_type='GEOMETRYZ', dimension=3, srid=DEFAULT_DB_SRID)
 
 
 MAX_FEATURE_GEOMETRY_SIZE = 1e6
 
-DEFAULT_OUPUT_SRID = 21781
+DEFAULT_OUTPUT_SRID = 21781
 
 
 def get_resolution(imageDisplay, mapExtent):
@@ -47,13 +48,13 @@ def has_buffer(imageDisplay, mapExtent, tolerance):
         and all(val != 0 for val in imageDisplay) and mapExtent.area != 0)
 
 
-def get_tolerance_meters(imageDisplay, mapExtent, tolerance, srid, srid_to=DEFAULT_OUPUT_SRID):
+def get_tolerance_meters(imageDisplay, mapExtent, tolerance, srid):
     if has_buffer(imageDisplay, mapExtent, tolerance):
         # if srs is geographic, convert bounds to meters
         if srid == 4326:
-            bounds = transform_round_geometry(mapExtent, srid, srid_to, rounding=True).bounds
+            bounds = transform_round_geometry(mapExtent, srid, DEFAULT_DB_SRID, rounding=True).bounds
         else:
-            bounds = trans_mapExtent.bounds
+            bounds = mapExtent.bounds
         map_meter_width = abs(bounds[0] - bounds[2])
         map_meter_height = abs(bounds[1] - bounds[3])
         img_px_width = imageDisplay[0]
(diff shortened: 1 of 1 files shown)


chsdi/models/vector/__init__.py
```diff
@@ -179,7 +180,7 @@ def label_column(cls):
 
     @classmethod
     def geom_filter(cls, geometry, imageDisplay, mapExtent, tolerance, srid):
-        tolerance_meters = get_tolerance_meters(imageDisplay, mapExtent, tolerance)
+        tolerance_meters = get_tolerance_meters(imageDisplay, mapExtent, tolerance, srid)
         geom_column = cls.geometry_column()
         wkb_geometry = WKBElement(buffer(geometry.wkb), srid)
         return func.ST_DWITHIN(geom_column,
```
ltshb, line 186: From the docu https://postgis.net/docs/ST_DWithin.html, the distance is in units of srid for geometry not in meter. Not sure here if `transform_geometry(geom_column, wkb_geometry, srid)` returns a geometry or a geography, neither I know what is the difference for postgresql between geometry and geography ?

ltshb (commented):
Nevertheless I tested this test deployment with my infra-e2e-tests `make test ENV=dev.env  TEST_NAMES=tests.test_mf_chsdi3` and all test passed even the one that trigger the ticket https://jira.swisstopo.ch/browse/BGDIINF_SB-2505.

chsdi/models/vector/__init__.py
ltclm: yes the interaction with postgis is always in lv95
```
WHERE ST_DWithin(bfs.gwr_chsdi.the_geom, ST_Transform(ST_GeomFromWKB(%(ST_GeomFromWKB_1)s, %(ST_GeomFromWKB_2)s), %(ST_Transform_1)s), %(ST_DWithin_1)s)
 LIMIT %(param_1)s
2022-07-13 11:30:43,421 INFO  [sqlalchemy.engine.base.Engine][waitress] {'ST_DWithin_1': 4.278382581649971, 'ST_GeomFromWKB_1': <read-only buffer for 0x7f42da81a870, size -1, offset 0 at 0x7f42da81ab30>, 'param_1': 201, 'ST_Transform_1': 2056, 'ST_GeomFromWKB_2': 4326}
```
the input geometry (WGS84) will be transformed to lv95 before it will be used for the st_dwithin function call in postgis:

https://github.com/geoadmin/mf-chsdi3/blob/master/chsdi/models/vector/__init__.py#L185

since the postgis interaction is always in lv95 we can always convert the tolerance into meters. I think we can even skip this transformation for LV95,LV03 and WebMerc which is already in meters https://epsg.io/3857.

ltshb: But `DEFAULT_OUPUT_SRID` is LV03 and not LV95 ? If I understand correctly we have to differentiate between the default SRID of the DB which is LV95, probably set here https://github.com/geoadmin/mf-chsdi3/blob/191da9d9f2129e1c05cf744af3a756c4b894bfa3/chsdi/models/vector/__init__.py#L21 ? and the default output SRID which is LV03. So if this is correct I suggest to create a new variable; `DEFAULT_DB_SRID` set to LV95 and use this in the model https://github.com/geoadmin/mf-chsdi3/blob/191da9d9f2129e1c05cf744af3a756c4b894bfa3/chsdi/models/vector/__init__.py#L21 and also in the `get_tolerance_meters()` method. However in this method I would not use it as parameter simply hardcoded in the transform.

```python
bounds = trans_mapExtent.bounds
# if srs is geographic, convert bounds to meters
if srid == 4326:
    bounds = transform_round_geometry(mapExtent, srid, DEFAULT_DB_SRID, rounding=True).bounds            
```

I agree that if the tolerance is in meter, we only need/should transform WGS84.

chsdi/models/vector/__init__.py
```diff
@@ -47,9 +48,13 @@ def has_buffer(imageDisplay, mapExtent, tolerance):
         and all(val != 0 for val in imageDisplay) and mapExtent.area != 0)
 
 
-def get_tolerance_meters(imageDisplay, mapExtent, tolerance):
+def get_tolerance_meters(imageDisplay, mapExtent, tolerance, srid, srid_to=DEFAULT_OUTPUT_SRID):
```
ltshb, line 51: `srid_to=DEFAULT_OUTPUT_SRID` parameter can be removed here.

PR #3973, opened 2022-07-15
BGDIINF_SB-2515: Remove Apache compression

ltshb:
When using the Apache compression delfate module, apache removes the Content-Length
header. Apparently this is due to the chunked encoding feature of apache. I
tried to increase the DeflateBufferSize like mentioned in the docu, but still
the Content-Length for small request was missing.

The main issue is that if the client don't send the "Accept-Encoding" header
CF set it to "Accept-Encoding: gzip" and then decompress the answer. In this
case then the Content-Length header is missing.

For now I removed the compression in service and let CF to do the compression.

In future it might a better idea to do compression directly in the service,
or if we change apache by another wsgi server we might have a better compression.

See BGDIINF_SB-2486 and BGDIINF_SB-2510

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,7 +42,6 @@ RUN echo "ServerName localhost" | tee /etc/apache2/conf-available/fqdn.conf \
         auth_basic \
         authz_groupfile \
         autoindex \
-        deflate \
         dir \
         env \
         expires \
--- a/apache/wsgi-py3.conf.in
+++ b/apache/wsgi-py3.conf.in
@@ -12,10 +12,6 @@
 RewriteEngine on
 ExpiresActive on
 
-AddOutputFilterByType DEFLATE text/css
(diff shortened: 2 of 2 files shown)


boecklic (commented):
Is it possible that we cause troubles here for users that for some reason rely on and expect that the response uses deflate compression?

ltshb:
> Is it possible that we cause troubles here for users that for some reason rely on and expect that the response uses deflate compression?

Mostly not because CloudFront will compress it anyway. Only for small response, CF will not compress it. But if a user is having problem because of the answer not being compressed anymore that IMHO more of his issue, because we advertise the compression header if compressed, so the user has to check the Content-Encoding first.

ltshb:
Note the current setup in Ireland, does not removed the Content-Length header when gzip, it seems there that compression is not done by Apache !

boecklic (approved):

PR #3977, opened 2022-07-19
Removed unused query parameter "destination" in stationboard endpoint

ltshb:
This code is not used anymore and has no effect


ltclm (approved):
👍

PR #3978, opened 2022-07-19
BGDIINF_SB-2488 bugfix : sdi services crashes on queries with specific SRID

ltkum:
Issue BGDIINF_SB-2488 : sdi service crashes on queries with SRID 3857 and 4326
Jira ticket : https://jira.swisstopo.ch/browse/BGDIINF_SB-2488

Cause of the problem :

The function aggregating the informations about all zoom levels was
checking through all zoom levels (0 to 28), but the problematic SRID
had less than 28 zoom level, encountering a problem within an assertion
which crashed the whole function.

Mitigation :

We use the RESOLUTION table length from the geoadmin tilegrid object
created with the srid, ensuring we no longer can try to access zoom
levels that are not available for a specific srid.

### DIFF
--- a/chsdi/models/bod.py
+++ b/chsdi/models/bod.py
@@ -328,7 +328,7 @@ def computeHeader(mapName, srid):
     maxZoom = len(gagrid.RESOLUTIONS)
     dpi = 90.7
     lods = []
-    for zoom in range(minZoom, maxZoom + 1):
+    for zoom in range(minZoom, maxZoom):
         lods.append(
             {'level': zoom,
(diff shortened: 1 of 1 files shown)


ltshb (commented):
Looks good to me, did you test it with the e2e test locally ?
See https://github.com/geoadmin/infra-e2e-tests/blob/739f0b7cab2d18071667bd03fff26d310b318dba/tests/test_mf_chsdi3/test_map_server.py#L125

ltshb (approved):

PR #3979, opened 2022-07-20
BGDIINF_SB-2517: Set default cache-control header within pyramid application

ltshb:
Do not set anymore the default cache-control within Apache but directly
within the application. The final goal is to remove as much as possible logic
from Apache.


ltshb:
I'm currently debugging to see why the new unittest for cache control works locally but not on the ci.

boecklic (approved):
:+1:

PR #3983, opened 2022-08-08
Improved python3 doc for local work

ltshb:


PYTHON3.md
```diff
@@ -37,32 +41,52 @@ change the variables you want and use them with
...
 ```
 
-## Docker run
+Then set the `DBHOST` environment variable to `localhost` (you can do this in your own environment file e.g. `.env.mine` and run the make file as follow: `summon make -f Makefile.frankfurt ENV_FILE=.env.mine serve`)
+
+### S3 Vector Bucket Access
+
+To have access to the S3 bucket, you can either set your `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment variable. **:warning: When using those variables make sure that the command are not saved into history file !**
+
+Alternatively if you are using `zsh` you can use the `aws` plugin (see [oh-my-zsh aws plugin](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/aws)) with the following command:
```
boecklic, line 67: Hmm, I'm not sure it's good idea to disclose all this information in a public repo. We can either move it to infra-kubernetes to a readme in this services folder or think about a general place where we document how to develop locally (I think we have to do this anyway somewhere for data integration).

boecklic (changes requested):

boecklic (approved):
After discussion: Ok like this for the moment, we'll reconsider and discuss general local development after migration is done

PR #3987, opened 2022-08-17
BGDIINF_SB-2516: Moved CORS headers settings from apache into pyramid app

ltshb:
This will allow in future to get rid of apache.


rebert (approved):
this is a unqualified approve - but in the code i could not see anything really bad

PR #3991, opened 2022-08-18
BGDIINF_SB-2353 Fixed CI manual build on branch master

ltshb:
When triggering a manual build in the CI giving the master branch as source
version (e.g. to redo a merge build that as not been automatically triggered)
the GIT_BRANCH variable was set to the tag value instead of `master`.

NOTE: `git describe` will return a tag if any is found, not a branch name, while `git show-ref --heads` will show the branch name on which the commit points.


rebert (approved):
thanks

PR #3992, opened 2022-08-18
BGDIINF_SB-2540: OpenData av hopping that this fix will work in a https setup

rebert:
My first fix was only working on the http:localhost but not on https://sys-api3. This will hopefully do it.


boecklic (approved):

ltshb:
@rebert @boecklic as quick fix OK. But I'm a bit concern at this code and have some questions:

1. From where comes the `download_url` ? Is it from the User ? Is it configured somewhere ?
2. Using as before `request.scheme` would be only correct if the download_url would/should point to the same server (aka sys-api3) but if this download_url is another server aka host name, then the scheme should be configured the same way as the external server and not hardcoded to HTTPS. I also see a security issue in this snippet that has nothing to do with this PR, `verify = false` this could lead to man of the middle attack !

PR #3994, opened 2022-08-18
Using an environment variable for the data.geo.admin.ch protocol

rebert:


rebert:
going to make a PR against current milestone branch

PR #3995, opened 2022-08-19
BGDIINF_SB-2454: drop Ireland support, only go for Frankfurt

hansmannj:
After migration to Frankfurt, obsolete files like rc_*, deploy related
scripts and config files, py2 buildspec file and git hooks became
obsolete and are deleted now. Also the README.md was updated to reflect
the Frankfurt setup.


.gitignore
```diff
@@ -26,12 +24,10 @@ chsdi/static/doc/build/
 chsdi/static/css/extended.min.css
 chsdi/static/css/blueimp-gallery.min.css
 chsdi/static/info.json
-deploy/hooks/pre-create-code
 node_modules
 MANIFEST
 buildout_branch.cfg
 25-mf-chsdi3.conf
-deploy/deploy-branch.cfg
 deploy/conf/00-branch.conf
```
ltshb, line 31: Should this not also be removed ?

.gitignore
```diff
@@ -26,12 +24,10 @@ chsdi/static/doc/build/
 chsdi/static/css/extended.min.css
 chsdi/static/css/blueimp-gallery.min.css
 chsdi/static/info.json
-deploy/hooks/pre-create-code
 node_modules
 MANIFEST
 buildout_branch.cfg
```
ltshb, line 29: Same for this ?

.gitignore
ltshb: Below should we also not remove `rancher-*.yml` and what about `tests/varnish*` ?

ltshb (commented):
About the Readme what did you do exactly ? I think it might be wise to do the following for the moment:
1. Rename README.md to README_IRELAND.md
2. Rename PYTHON3.md to README.md
We can then delete README_IRELAND.md together with the Makefile.Ireland.

.gitignore
hansmannj: Good catches, thanks! I'll scan through the .gitignore again and try to find everything that I did not find in the first shot

hansmannj:
> About the Readme what did you do exactly ? I think it might be wise to do the following for the moment:
> 
>     1. Rename README.md to README_IRELAND.md
> 
>     2. Rename PYTHON3.md to README.md
>        We can then delete README_IRELAND.md together with the Makefile.Ireland.

I kind of merged them so that the resulting README.md represents the Frankfurt setup.
But your suggestion makes sense, would make it easier to check former states, if needed. I'll do it that way

ltshb (approved):

PR #4002, opened 2022-08-24
BGDIINF_SB-2554: using host header and protocol in url construction in loader.js

hansmannj:
For all request objects the `wsgi.url_scheme` is now adapted based on the existence and value of `CloudFront-Forwarded-Proto` or `X-Forwarded-Proto` headers, to allow correct URL constructions.


chsdi/views/loader.py
```diff
@@ -34,13 +40,14 @@ def loadjs(request):
...
-        return 'https://%s/%s/%s%s.%s' % (
-            loader_js_bucket_host, s3_resources_path, filename, mode_str, extension)
+        return '%s/%s/%s%s.%s' % (
+            host, s3_resources_path, filename, mode_str, extension)
 
     ga_css = get_resource_url('ga', 'css')
     ga_js = get_resource_url('ga', 'js', mode_str)
     epsg_21781_js = get_resource_url('EPSG21781', 'js')
     epsg_2056_js = get_resource_url('EPSG2056', 'js')
+    api_url = '%s%s' % (host, request.path)
```
hansmannj, line 50: Not sure if this adaption is needed here

chsdi/views/loader.py
```diff
@@ -50,7 +57,7 @@ def get_resource_url(filename, extension, mode_str=''):
             'ga_js': ga_js,
             'epsg_21781_js': epsg_21781_js,
             'epsg_2056_js': epsg_2056_js,
-            'api_url': request.path_url.replace('/loader.js', ''),
+            'api_url': api_url.replace('/loader.js', ''),
```
hansmannj, line 60: Dito, not sure about this adaption here

chsdi/views/loader.py
```diff
@@ -16,7 +16,13 @@ def loadjs(request):
     ignore_polyfill = request.params.get('ignore_polyfill')
     # Determined automatically in subscriber
     lang = request.lang
-    loader_js_bucket_host = request.registry.settings['loader_js_bucket_host']
+
+    if "loader_js_bucket_localhost" in request.registry.settings:
+        host = request.registry.settings["loader_js_bucket_localhost"]
+    elif "CloudFront-Forwarded-Proto" in request.headers:
+        host = '%s://%s' % (request.headers["CloudFront-Forwarded-Proto"], request.host)
```
ltshb, line 23: You should also add an `elif "X-Forwarded-Proto" in request.headers:` which is the standard header for such use case.

Also rename the `host` variable to `base_url`. `host` usually refer to an hostname and not a full url with protocol.

chsdi/views/loader.py
ltshb: This is not good, the API url refer to the mf-chsdi3 API url not to the bucket URL. But the original code has also an issue, it doesn't take into account the Forwarded Proto header ! So it will come with http instead of https for the official api call (not hproxy).

chsdi/views/loader.py
ltshb: But it would be much better to solve this in a generic way like in our Flask application, where the header check is done for all request object. Otherwise we need to implement this check everytime we need to generates url !

ltshb: In our Flask we have the header check set in gunicorn, maybe  we can do this in Apache as well, because we wants to get rid of apache I would only check for Cloudfront-Forwarded-Proto in Apache, later we moving to gunicorn we can use the same logic as in our Flask application: https://github.com/geoadmin/service-shortlink/blob/a8786e675a9c01584a2473fe05c92bf48251a78c/wsgi.py#L55

ltshb:
Based on mod_wsgi docu https://modwsgi.readthedocs.io/en/master/configuration-directives/WSGITrustedProxyHeaders.html we could use such apache config

```apache
# Reads the value of Cloudfront-Forwarded-Proto into TempValue
SetEnvIf CloudFront-Forwarded-Proto ".+" CF_FORWARDED_PROTO=$0

RequestHeader setifempty X-Forwarded-Proto %(CF_FORWARDED_PROTO)e env=CF_FORWARDED_PROTO

WSGITrustedProxies 1.2.3.4
WSGITrustedProxyHeaders X-Forwarded-Proto
```

One issue here is `WSGITrustedProxies` which I don't know if it supports wildcard or subnet (If found nothing in docu about this) ! Also I don't know how we could get the prox address in apache config. `WSGITrustedProxies` is unfortunately required by `WSGITrustedProxyHeaders` which would set the correct `wsgi.url_scheme` variable.

Because we are planning anyway to replace Apache by gunicorn See BGDIINF_SB-2486, we might use a temporary workaround using the pyramid Request Factory (https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/hooks.html#changing-the-request-factory) in which we would correct the `wsgi.url_scheme` based on headers. This hack could be later on removed.

chsdi/views/loader.py
hansmannj: > You should also add an `elif "X-Forwarded-Proto" in request.headers:` which is the standard header for such use case.
> 
> Also rename the `host` variable to `base_url`. `host` usually refer to an hostname and not a full url with protocol.

How about `host_url`, which would be Pyramid-like: https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/webob.html#urls

chsdi/views/loader.py
hansmannj: But then I could simply adapt line 50 as follows:
`api_url = host` (or `host_url`), as defined in the if.. elif.. above
Then it should use the correct proto and only the `host_url`

hansmannj: But as I understand it, currently the `'api_url': request.path_url.replace('/loader.js', '')` also is not only restricted to the API url, but has the full request path in it

chsdi/views/loader.py
ltshb: simply using `host_url` won't solve it. It is correct to use such variable in code, but for know we have the issue that CF send Cloudfront-Forwarded-Proto instead of X-Forwarded-Proto and that Apache don't set the correct scheme in wsgi.url_scheme !

hansmannj: Yes, as discussed, the generic way might be best. I just meant as a variable name. Using `host_url` should be quite the way, it is used by Pyramid, too.

ltshb (approved):
LGTME

chsdi/subscribers.py
```diff
@@ -61,8 +61,6 @@ def add_localizer(event):
     # TODO: clean-up when only Python 3.x and no longer 2.x is in use
     if not six.PY3:
         request.lang = request.lang.encode('ascii', 'ignore')
-    # The load balancer forwards requests as http, therefore we need to check X-Forwarded-Proto
-    request.scheme = request.headers.get('X-Forwarded-Proto', request.scheme)
```
hansmannj, line 65: This was playing tricks on me and overriding the newly implemented request factory.
Deleting this should do the job @ltshb @boecklic

ltshb: That was a good example of bad code, changing the `request.scheme` inside a `add_localizer()` method. Once again it shows that good code with proper naming avoid headache and long debugging session.
@hansmannj very good job and finding, one step ahead to cleaner, better and more stable code :-)

ltshb (approved):
👍🏼

chsdi/__init__.py
```diff
@@ -21,12 +25,30 @@ def cleanup(request):
...
+    # e.g. gunicorn.
+
+    def __init__(self, environ, **kwargs):
+        logger.warning("********************** BEFORE: %s", environ)
+        if "HTTP_CLOUDFRONT_FORWARDED_PROTO" in environ:
+            environ["wsgi.url_scheme"] = environ["HTTP_CLOUDFRONT_FORWARDED_PROTO"]
+        elif "HTTP_X_FORWARDED_PROTO" in environ:
+            environ["wsgi.url_scheme"] = environ["HTTP_X_FORWARDED_PROTO"]
+        super().__init__(environ, **kwargs)
+        logger.warning("*********************** AFTER: %s", environ)
```
ltshb, line 41: Oops do not Forget to clean up the code here and to remove the logger to speed up the code.

hansmannj: Oops, thanks. They came with the last commit, I guess. Will remove them of course :wink:

ltshb (approved):

PR #4003, opened 2022-08-24
Fix for dataintegration:

rebert:
- right now it is not possible to see the legends and others when
serving api_url=//localhost, as the protocol is being switches from http to https.


ltclm (approved):
👍

rebert:
@procrastinatio any concerns here?

procrastinatio (approved):
OK

PR #4004, opened 2022-08-25
Fixed mako with versionned urls

ltshb:
Some makos uses versionned urls, for those we have an apache rewrite url
that should match an official release version optionally followed by a git
hash.

This might be an issue if the git repository contains tag that don't match
our milestone version scheme. Therefore now we check if the GIT_TAG is valid
and if not we use a dummy version for the APP_VERSION.

This is only an issue on develop-* branches as here we don't do any git tagging on PR merge,  os it might take an invalid tag that has been done manually for any reason. That was the case with the manual tag `LAST_IRELAND_WORKING_SETUP_DEVELOP` that was created during the migration.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,12 +65,14 @@ RUN pip3 install -e .
 ARG GIT_HASH=unknown
 ARG GIT_BRANCH=unknown
 ARG GIT_DIRTY=unknown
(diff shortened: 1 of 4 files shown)


rebert (approved):
wow! thanks @ltshb

PR #4006, opened 2022-08-29
BGDIINF_SB-2506 feature : using pipenv to build the project

ltkum:
This should solve the following Issue : 

Work done : 
Creating a Pipenv file based on the old requirement file
Removing the requirement file
Updating the Makefile
Updating the Dockerfile


Dockerfile
```diff
@@ -20,6 +20,7 @@ RUN apt-get update -qq \
         libgeos-dev \
         gettext-base \
         apache2 libapache2-mod-wsgi-py3 \
+        gcc \
```
ltshb, line 23: Why `gcc` ?

Dockerfile
```diff
@@ -28,12 +29,15 @@ RUN apt-get update -qq \
     && mkdir -p /var/www/vhosts/${VHOST}/private \
     && mkdir -p /var/www/vhosts/${VHOST}/cgi-bin \
     && mkdir -p /var/www/vhosts/${VHOST}/htdocs \
-    && mkdir -p /var/www/vhosts/${VHOST}/logs \
-    && pip3 install -q --upgrade pip==21.2.4 setuptools --index-url ${PYPI_URL}
+    && mkdir -p /var/www/vhosts/${VHOST}/logs
+
+RUN pip3 install pipenv && pipenv --version
```
ltshb, line 34: Installation of pipenv should stay in the same docker layer as the other apt install.

Dockerfile
```diff
@@ -28,12 +29,15 @@ RUN apt-get update -qq \
...
+
+WORKDIR /var/www/vhosts/${VHOST}/private/chsdi
 
 # FIXME: use pipenv
```
ltshb, line 38: Remove comment

.gitignore
```diff
@@ -21,6 +22,7 @@ apache/tomcat-print.conf
 chsdi.egg-info
 chsdi/locale
 chsdi/static/doc/build/
+chsdi/static/doc/Pipfile
```
ltshb, line 25: ?

Dockerfile
```diff
@@ -28,12 +29,15 @@ RUN apt-get update -qq \
     && mkdir -p /var/www/vhosts/${VHOST}/private \
     && mkdir -p /var/www/vhosts/${VHOST}/cgi-bin \
     && mkdir -p /var/www/vhosts/${VHOST}/htdocs \
-    && mkdir -p /var/www/vhosts/${VHOST}/logs \
-    && pip3 install -q --upgrade pip==21.2.4 setuptools --index-url ${PYPI_URL}
+    && mkdir -p /var/www/vhosts/${VHOST}/logs
+
+RUN pip3 install pipenv && pipenv --version
+
+WORKDIR /var/www/vhosts/${VHOST}/private/chsdi
```
ltshb, line 36: I would not change the workdir before copying the pipenv file.

Dockerfile
```diff
@@ -58,9 +62,7 @@ RUN echo "ServerName localhost" | tee /etc/apache2/conf-available/fqdn.conf \
         alias
 
 COPY --chown=${USER}:${GROUP} . /var/www/vhosts/${VHOST}/private/chsdi
-WORKDIR /var/www/vhosts/${VHOST}/private/chsdi
```
ltshb, line 61: See comment above.

Makefile
```diff
@@ -13,11 +13,20 @@ SHELL = /bin/bash
...
+REQUIREMENTS_TIMESTAMP = $(TIMESTAMPS)/.requirements.timestamp
+DEV_REQUIREMENTS_TIMESTAMP = $(TIMESTAMPS)/.dev-requirements.timestamps
+
+# PIPENV files
+PIP_FILE = Pipfile
+PIP_FILE_LOCK = Pipfile.lock
 
 # default configuration
-ENV_FILE ?= .env.default
-include $(ENV_FILE)
```
ltshb, line 20: Those are needed !

Dockerfile
```diff
@@ -58,9 +62,7 @@ RUN echo "ServerName localhost" | tee /etc/apache2/conf-available/fqdn.conf \
...
 
-RUN pip3 install -e .
```
ltshb, line 63: Currently how the project is setup, this is probably still needed, and I would not remove this. To be checked if the pipenv editable mode (https://pipenv-fork.readthedocs.io/en/latest/basics.html#editable-dependencies-e-g-e) would do the same job, otherwise we still have the possibility to use pip3 for this task (pipenv is just a wrapper around pip)

ltshb (changes requested):
chsdi3 don't use yapt nor pylint (and will probably never as it would be way to much work) so you cannot use copy paste from other makefile here, so the best is just to do the minimal changes in the makefile.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.7-buster
+FROM python:3.7-slim-buster
```
ltshb, line 1: I would stay on the `python:3.7-buster` images for chsdi3.

Makefile
```diff
@@ -175,26 +182,21 @@ help:
...
 # TODO: add targets `translate` when merged
 .PHONY: all
 all: setup config-templates lint build
 
 
 .PHONY: setup
-setup: $(VENV) $(NODE_MODULES)
-
+setup: $(NODE_MODULES)
+	if ! [ -d "$(VENV)" ] ;then pipenv install ;fi
```
ltshb, line 194: dev dependencies are missing `pipenv install --dev`

ltshb: For the serve to work it might be also needed to do as before the `pipenv install -e .`

requirements-py3.txt
```diff
@@ -63,3 +63,4 @@ WebTest==2.0.33
 zope.deprecation==4.4.0
 zope.interface==4.6.0
 unidecode==1.1.1
+
```
ltshb, line 66: requirements-py3.txt should be removed.

ltshb (commented):
Several issues:
- CI does not pass
- Rebase into master and change base of the PR, do it on master and then merge master into develop-*
- Not working locally  (same error as on CI)
- squash all commits

PS: in future use the correct nomenclature for the branch name in order to have the correct label to the PR.

Makefile
ltshb: Seems that adding `pipenv install -e .` here fixed the issue when serving locally

Pipfile
```diff
@@ -0,0 +1,76 @@
...
+translationstring= "~=1.3"
+unittest-xml-reporting= "~=2.5.2"
+venusian= "~=1.2.0"
+waitress= "~=1.4.3"
+WebOb= "~=1.8.5"
+WebTest= "~=2.0.33"
+"zope.deprecation"= "~=4.4.0"
+"zope.interface"= "~=4.6.0"
+unidecode= "~=1.1.1"
+
```
ltshb, line 74: You need to add `[dev-packages]` section even if empty in order to pipenv to work correctly with `install -e .`

ltshb: Using here `~=` resolve to some dependencies warning, might be a good idea to stick to `==` like in pip for this PR.

requirements-py3.txt
ltkum: I tried, but apparently, the setup.py file is using this in installation for pipenv install -e (got an error saying the file couldn't be found)
We'll have to make changes in setup.py to be able to remove this file, and it would be best to do it in a second time.

I can include a comment in the package to specify this.

ltshb: Then we have to modify `setup.py`. This has to be done within this PR, having two requirement files is not a good, so we have to be sure that we can get rid of the old one, otherwise it don't make sense to go to pipenv.
One idea could be to modify `setup.py` to generate the requirements using `pipenv lock -r` command instead of reading the old requirement file.

hansmannj (commented):
When scanning through the CodeBuild logs I noticed, that there are quite some "useless" lines showing pipenv's progress when installing stuff.
It might make sense to add a `PIPENV_NOSPIN: 1` in the `env` section of `buildspec-py3.yml`. Could make reading the logs a bit easier :wink:

PR #4010, opened 2022-09-02
releasenotes for deploy 2022-09-07

rebert:


davidoesch:
all ok from my side. please take into account any changes which are introduced to fix side effects of the migration which go live with trhis deploy

rebert:
@davidoesch good point. But these side-effects are already on prod since the migration to Frankfurt. The data deploy should change nothing (beside of data changes).

ltclm (approved):

PR #4011, opened 2022-09-05
BGDIINF_SB-2506: using pipenv to build the project

ltkum:
This should allow the use of pipenv within chsdi3

I decided to make a new branch to allow for some much needed cleanup after switching the base to master, and cherry-picked my changes.

everything discussed within https://github.com/geoadmin/mf-chsdi3/pull/4006 should have been corrected.


setup.py
```diff
@@ -7,9 +7,21 @@
...
+if not os.path.exists(os.path.join(here, REQUIREMENTS_FILE)):
+    requirements = open(os.path.join(here, 'Pipfile')).read().split('\n')
+    requirements = requirements[requirements.index('[packages]') + 1:
+                                requirements.index('[requires]')]
+    requirements.pop(requirements.index('[dev-packages]'))
+    requirements = [req.replace('"', '') for req in requirements]
+    requirements = [req.replace(' = ', '') for req in requirements]
+    with open(os.path.join(here, REQUIREMENTS_FILE), 'w+') as file:
+        for req in requirements:
+            file.write(f"{req}\n")
```
ltshb, line 24: This could be much simpler, simply call `pipenv lock -r` no need to parse pipfile !

```python
import subprocess

requires = subprocess.check_output('pipenv lock -d -r', shell=True, stderr=None).decode().split('\n')
```

Then no need to generate a local file and add it to git ignore and dockerignore. Moreover the code above would be problematic when the file already exists it won't be updated !

ltshb (changes requested):

setup.py
ltkum: I didn't know about the -r flag, thanks for the input, it will indeed greatly simplify this.

ltshb:
@ltkum if the setup.py requires make trouble we can totally remove it as we handles dependencies outside of setup.py anyway and as we don't publish mf-chsdi3 as a pip package.

It seems also that `PIPENV_NOSPIN` as no effect anymore (same on other project) maybe something changed in pipenv and needs to be addapted in the buildspec/makefile to make pipenv less verbose.

buildspec-py3.yml
```diff
@@ -152,6 +153,7 @@ phases:
         --build-arg VERSION="${GIT_TAG}"
         -t "${DOCKER_IMG_TAG}" -t "${DOCKER_IMG_TAG_LATEST}" .
       - echo Build completed on $(date)
+      - pipenv install -e .
```
ltshb, line 156: This should not be needed here.

buildspec-py3.yml
```diff
@@ -23,6 +23,7 @@ env:
     DBSTAGING: dev
     GEODATA_STAGING: test
     ENV_FILE: .env.ci
+    PIPENV_NOSPIN: True
```
ltshb, line 26: Try to add also `CI: 1` here in plus of the PIPENV_NOSPINT

buildspec-py3.yml
ltkum: it is absolutely needed here, as the tests use the chsdi package. I could put it right before the tests, though

Makefile
```diff
@@ -255,12 +254,13 @@ legends: $(VENV) guard-BODID guard-WMSHOST
 
 
 .PHONY: serve
-serve: $(VENV) config-templates set-app_version
+serve: setup config-templates set-app_version
+	pipenv install -e .
```
ltshb, line 258: This should be part of the setup target as before, this will also solve the CI test issue.

buildspec-py3.yml
ltshb: See makefile comment

Makefile
```diff
@@ -176,24 +183,16 @@ help:
...
-	@echo "${GREEN}requirements-py3.txt has changed...${RESET}";
-
-
-# TODO: replace through pipenv as in the other projects.
-$(VENV): requirements-py3.txt
-		test -d "$(VENV)" ||  ( $(SYSTEM_PYTHON_CMD) -m venv $(VENV) && \
-		${PIP} install $(PIP_QUIET) --upgrade pip==21.2.4 setuptools --index-url ${PYPI_URL}  && \
-		${PIP} install $(PIP_QUIET) -r requirements-py3.txt --index-url ${PYPI_URL}  -e . )
+setup: $(NODE_MODULES)
+	if ! [ -d "$(VENV)" ] ;then pipenv install && pipenv install --dev;fi
```
ltshb, line 195: `pipenv install --dev` is sufficient here no need to do a `pipenv install` as well.

ltshb (commented):
The code looks good. However after testing this adds a big drawback, each docker build takes more than 2 minutes even if only one python file changed. This is due to the way we install using the `setup.py` and apparently the only way to improve the subsequent build time is to get rid of `setup.py`. I'm not sure if the setup.py is currently needed/used by the apache setup and this could interfere with BGDIINF_SB-2486 ? 
But before this changes subsequent build were much faster and because we are constantly developing on this repo due to data integration it is quite important to keep it fast.

Dockerfile
```diff
@@ -59,8 +61,7 @@ COPY --chown=${USER}:${GROUP} . /var/www/vhosts/${VHOST}/private/chsdi
 
 WORKDIR /var/www/vhosts/${VHOST}/private/chsdi
 
-RUN pipenv install --system --deploy --ignore-pipfile
-
+RUN pip3 install -e .
```
ltshb, line 64: @ltkum Very good idea 👍🏼

ltshb (approved):
Great thanks. I just pushed a dependency issue (already present before this PR).

ltshb:
@ltkum as I need this for my next PRs to avoid conflict I allow myself to merge it.

PR #4013, opened 2022-09-05
With this fix the make legends can call localhost as well:

rebert:
- localhost is served on http (not https)

- the help in make legends is more informativ

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,7 @@ DOC_BUILD := chsdi/static/doc/build
 
 # Create legends
 WMSSCALELEGEND ?=
+
 NODE_MODULES := node_modules
 
 ifeq ($(CI_QUIET), 1)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -238,7 +239,11 @@ rss: $(VENV) doc chsdi/static/doc/build/releasenotes/index.html
 
 .PHONY: legends
 legends: $(VENV) guard-BODID guard-WMSHOST
-	scripts/downloadlegends.sh $(WMSHOST) $(BODID) $(WMSSCALELEGEND)
+	WMSPROTOCOL="https"; \
```
ltclm, line 242: or we take the whole url into WMSHOST, including protocol.
`WMSHOST=https://localhost:9X78`

ltclm (approved):
thank you!

Makefile
rebert: I did not want to get into .env.default vars ...

PR #4018, opened 2022-09-08
BGDIINF_SB-2456: Removed the download kml endpoint

ltshb:
This endpoint was only used as a workaround from the mf-geoadmin3 map viewer
for MacOS Firefox browser (apparently not all version only some olders version),
see BGDIINF_SB-2591.


hansmannj (approved):
:+1: Thanks

procrastinatio (approved):

PR #4019, opened 2022-09-08
Removed useless service-alti apache config

ltshb:
This apache config is a left over for when service-alti was still part of chsdi3.


hansmannj (approved):
:+1:

procrastinatio (approved):

PR #4021, opened 2022-09-08
Feat bgdidic 2474 bfe elektrische anlagen ueber 36

faselm:
This layer should be good but I can't see how to provide a test link. I can test successfully locally with:

https://mf-geoadmin3.dev.bgdi.ch/index.html?configs_url=%2F%2Flocalhost:9577%2Fconfig&api_url=%2F%2Flocalhost:9577&lang=en&topic=ech&bgLayer=voidLayer&layers=ch.bfe.elektrische-anlagen_ueber_36&layers_opacity=0.8&E=2668196.06&N=1195104.19&zoom=2

How can I make a "public" link? The wiki is still with deploybranch but the cmd doesn't exist any more...


marionb:
@faselm if I understood correctly you can only provide a test link on  once the branch is merged. Before that you have to test with locahost.

faselm:
> @faselm if I understood correctly you can only provide a test link on once the branch is merged. Before that you have to test with locahost.

@marionb OK I see. I thought we could give the branch as an argument somehow but that's OK that way as well, thanks

rebert (approved):

PR #4024, opened 2022-09-19
BGDIINF_SB-2486: Removed static assets url rewrite from apache - #2

ltshb:
Now all assets URL rewrites are handled by pyramid using the Akhet library.


hansmannj (approved):
Looks good as far as I can tell :+1: Thx

procrastinatio (approved):
OK

PR #4022, opened 2022-09-15
BGDIINF_SB-2486: Replace Apache by gunicorn

ltshb:
Improvements with this changes:
- Size of the docker image reduced from ~760MB to ~490MB
- Logging simplified with JSON
----
E2E Tests
- E2E tests run on INT with 2022-09-07-rc2 (Apache)
  - 48.5s, 32.5s, 36.8s, 35.5s, 34.9s, 41s => avg ~38.2s
- E2E tests run on INT with Gunicorn
  - 82.7s, 59.4s, 68.4s, 71.2s, 47.7s, 66.2s, 61.4s => avg ~65.3s

Performance Tests
image
- BASE: GUNICORN
- TARGET: APACHE
Unfortunately we are much slower with gunicorn (3 times slower)!

PR #4025, opened 2022-09-19
BGDIINF_SB-2486: Moved the new config style url rewrite inside python - #3

ltshb:
For this support I had to move the `add_localizer()` hook from `NewRequest` to
`ContextFound` event. In the first one no routing is done and we cannot take
the `lang` parameter from the path for the new config style urls, while in the
ContextFound the routing has already be done and the `lang` url parameter has
already been parsed.

### DIFF
--- a/apache/wsgi-py3.conf.in
+++ b/apache/wsgi-py3.conf.in
@@ -37,13 +37,6 @@ RedirectMatch ^${APACHE_ENTRY_PATH}$ ${APACHE_ENTRY_PATH}/
 RewriteRule ^${APACHE_ENTRY_PATH}/\d{4}-\d{2}-\d{2}-rc\d+(?:-[^/]*)?/(.*)$ ${APACHE_ENTRY_PATH}/$1 [E=${APACHE_BASE_PATH}setyearcache:true,PT]
 Header merge Cache-Control "max-age=31536013, public" env=${APACHE_BASE_PATH}setyearcache
 
-# New style config url (see mf-geoadmin3 #4687)
-RewriteRule ^${APACHE_ENTRY_PATH}/configs/(de|fr|it|rm|en)/layersConfig\.json ${APACHE_ENTRY_PATH}/rest/services/all/MapServer/layersConfig?lang=$1 [PT]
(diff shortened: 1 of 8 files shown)


ltshb:
@pakb @boecklic I'm don't know why those new config style url have been in the past added, but looking at the access logs they are not all used. @pakb could you check in the old viewer and new viewer to see which ones are effectively used by the viewers ?
1. `/configs/(de|fr|it|rm|en)/layersConfig.json`
2. `/configs/(de|fr|it|rm|en)/translations.json`
3. `/configs/(de|fr|it|rm|en)/catalog.<MAP>.json`
4. `/configs/(de|fr|it|rm|en)/services.json` (from my understanding this one is not used and could be removed ?)
5. `/configs/services.json`

procrastinatio (approved):
OK.

pakb:
1. `/configs/(de|fr|it|rm|en)/layersConfig.json`
Doesn't seem to be used by either project

2. `/configs/(de|fr|it|rm|en)/translations.json`
Doesn't seem to be used by either project

3. `/configs/(de|fr|it|rm|en)/catalog.<MAP>.json`
Used only in tests, so maybe it's safe to remove... The main call (when creating the static files) is made through the standard enpoint (https://github.com/geoadmin/mf-geoadmin3/blob/da76d114559986fe2640684326907734178da72c/mk/serviceconfig.mk#L19-L28)
Potential usage:
https://github.com/geoadmin/mf-geoadmin3/blob/490a0f0e7d92768b3581cccf88b16b1b0164a505/test/specs/js/CatalogtreeController.spec.js#L41

4. `/configs/(de|fr|it|rm|en)/services.json`
https://github.com/geoadmin/mf-geoadmin3/blob/da76d114559986fe2640684326907734178da72c/mk/serviceconfig.mk#L19-L28
https://github.com/geoadmin/mf-geoadmin3/blob/e05e96ae7197e6d181478a7ba2a67ba64e66e39a/src/js/GaModule.js#L183-L191

5. `/configs/services.json`
It looks like it is also made static somehow in the Makefile, and is stored on the S3 bucket

ltshb:
As discussed with @boecklic because we cannot get totally rid of Apache due to performance see #4022 we keep those URL rewrite in Apache config because it simpler to have them here as in the code.

PR #4026, opened 2022-09-19
BGDIINF_SB-2486: Removed makos versioned url - #4

ltshb:


rebert (commented):
I tried to get it running and I'm getting the error message that the module akhet cannot be found. But this is not related directly to this PR I guess...

ltshb:
@rebert on master we are know using pipenv instead of pip, so try to do 
```bash
git clean -xdf --exclude=.env.mine
make cleanall
make setup
```
This should fix your issue

rebert (approved):
Thanks

ltshb:
@boecklic @rebert what do you think about cache-control should have the static asset a long max-age as the other requests ?

rebert:
Global:
- favicon = 326B
- touch-icon-bund-152x152 = 1.19KB

Only on extended tooltip hydropower, klaeranlagen etc.:
- extended.min.css=8.18KB
- jquery.min.css = 31.03KB
- blueimp-gallery.min.css = 6.65KB
- blueimp-gallery.min.js = 30.49KB

Only on tooltip modelled wind speed:
- d3.min.js = 148.17KB
- 3d-tip.js =8.08KB

As the size is relatively small and the tooltips and extendend tooltips are normaly used by humans (not like API identify etc.) I would not add an extra caching policy for these assets.

boecklic:
What's the default cache duration again? Are those global assets used in the viewer or the api doc?

ltshb:
> What's the default cache duration again? Are those global assets used in the viewer or the api doc?

@boecklic see description, currently we set the default to 30 minutes. Those are the asset with their urls defined in makos (html popups) so only used by the viewer as far as I know/understand. I will add a environment variable to set another cache control value for static assets in general which might be a good idea ?

procrastinatio (approved):
OK. But I guess it needs some changes client-side

ltshb:
@procrastinatio do you mean that the old viewer is directly using those urls ? From my understanding I thought that the client (viewer) is using those urls from the makos template, so this would mean that nothing has to be done on client side (unless clearing cache), but I might be wrong.

PR #4027, opened 2022-09-21
BGDIINF_SB-2486: Improving docker image size

ltshb:
- Changing docker base image to the slim version plus not installing dev package reduced the size of the image from 760MB down to 480MB.
- Using docker multi-stage build reduced even more the size down to 356MB
- Use the docker image slim version to reduce image size
- Moved development dependencies into the dev section in Pipfile, this also
reduce the docker image size as dev dependencies are not installed on docker
- Clean up environment variables, moved some variable that are not really configuration
variables like INSTALL_DIR, VHOST_DIR, APACHE_BASE_PATH, APACHE_ENTRY_PATH, MODWSGI_USER
into the dockerfile so that the user don't have to worry about it
- Use the same http port when working locally with Waitress or with Docker
- Remove waitress from the production.ini file, waitress is the wsgi server
used to serve the application locally outside of docker. Docker use mod_wsgi
as wsgi server and not waitress.
- Do not copy anymore the whole repo into the docker image but only the relevant
directories. This avoid add files into the root of the repo that we don't want
to come into the docker image and we forget to add them to dockerignore.
- To simplify the docker file, did some file re-organization
- Don't generates the apache templates when working locally, they are only generated
inside the docker container where they are used.
- By default in makefile use `.env.mine` which is created from `.env.default`
if it doesn't exists. This simplify the local development
- Clean up old stuff from makefile and improve/simplify dependencies. The main
issue in dependencies was the `setup` target that was not re-done when the
pipfiles changed.

For deployment see also https://github.com/geoadmin/infra-kubernetes/pull/276


Makefile
```diff
@@ -384,22 +376,31 @@ $(LANGUAGES_MO_FILES): $(LANGUAGES_PO_FILES)
...
 guard-%:
 	@ if test "${${*}}" = ""; then \
-	  echo "Environment variable $* not set. Add it to your command."; \
-	  exit 1; \
+		echo "Environment variable $* not set. Add it to your command."; \
+		exit 1; \
 	fi
 
+
+guard-VENV:
```
ltkum, line 392: This is a really elegant was of enforcing the setup requirement while not having a need to force a check on every target. I really like it.

ltkum (approved):
From what I've seen, this is solid work. Good usage of multi stage builds in docker in particular.

From what I've seen, there is no problem and we can merge this, but I'd wait for someone else expertise on apache files, in case there is something there that I missed.

boecklic (approved):

PR #4030, opened 2022-09-27
BGDIINF_SB-1918: Define in doc, that not always sr=4326 is possible

rebert:


hansmannj (approved):
:+1: Thanks!

ltclm (approved):
thank you!

PR #4034, opened 2022-09-29
BGDIINF_SB-2496: cleanup - remove all python2 relicts and the module six

rebert:


procrastinatio (approved):
Why not, but I'ld have cleanup a bit more.

chsdi/lib/helpers.py
```diff
@@ -263,7 +252,7 @@ def merge_statements(statements, operators):
         full = [x for x in chain.from_iterable(zip_longest(statements, operators))
             if x is not None]
 
-        return unicode(" ".join(full))
+        return str(" ".join(full))
```
ltshb, line 255: `str` is not needed here

ltshb (approved):

PR #4042, opened 2022-10-14
BGDIINF_SB-2631: Added support for JSON logs

ltshb:
Note it is recommended to use APACHE_LOG_LEVEL to warning to avoid non json message from the mod_wsgi. The mod_wsgi will still log message in a non json format, but with the level to warning no message  should appear unless unexpected behavior.

NOTE: require https://github.com/geoadmin/lib-py-logging-utilities/pull/33


rebert (approved):
Thanks

boecklic (approved):

hansmannj (approved):
Thanks! :1st_place_medal:

PR #4045, opened 2022-10-19
BGDIINF_SB-2497: using nose2 instead of nosetests

hansmannj:


hansmannj:
Not exactly sure about the pseudo e2e tests. They don't seem to be doing too well, but also in the current master state, too.

Makefile
```diff
@@ -310,27 +309,45 @@ dockerpull:
...
 test: guard-VENV clean_logs local-templates build
-	@echo "${GREEN}Unit testings...${RESET}";
+	mkdir -p junit-reports/integration
+	@echo "${GREEN}Unit testing (integration tests)...${RESET}";
+	${PYTHON} ./scripts/pg_ready.py
+	${NOSE} \
+	-c tests/nose2_tests_integration.cfg \
+	-s tests/integration \
+	--verbose
+	-N 7
```
ltshb, line 319: NB process to 7 might not be a good choice for every one ! This needs to be set by an ENVIRONMENT variable and the default to `0` which means that nose2 set it to the actual number of cpu

Makefile
```diff
@@ -310,27 +309,45 @@ dockerpull:
...
+	@echo "${GREEN}Unit testing (functional tests)...${RESET}";
+	mkdir -p junit-reports/functional
 	${PYTHON} ./scripts/pg_ready.py
-	$(NOSE) --verbosity=2 --cover-erase tests/ -e .*e2e.*
+	${NOSE} \
+	-c tests/nose2_tests_functional.cfg \
+	-s tests/functional \
+	--verbose \
+	-N 7
```
ltshb, line 327: Dito

Makefile
```diff
@@ -310,27 +309,45 @@ dockerpull:
...
 
 .PHONY: teste2e
 teste2e: guard-VENV clean_logs local-templates build
+	mkdir -p junit-reports/e2e
 	@echo "${GREEN}Pseudo E2E tests...${RESET}";
-	$(NOSE) tests/e2e/
+	${NOSE} --verbose \
+		-c tests/nose2_tests_e2e.cfg
+		-s tests/e2e \
+		-N 7
```
ltshb, line 350: Dito

ltshb (commented):
I'm not sure this is really a plus to split functional test from integration in the nose2 configs. But if we want to keep it I would try to use the `-s` and `-t` options instead of using file prefix.
Did you check if functional tests failed that integration don't start ? Otherwise if functional tests failed and integration passed, then you will miss the functional test failure.

Makefile
hansmannj: Good point, thanks! The `N` is more a relict of my local testing, since I wanted to make that faster.
Might be good to leave it parallelized tough, but as you say, using an env var for it makes sense :+1:

.env.default
```diff
@@ -16,6 +16,7 @@ GEODATA_STAGING=test
 HOST=sys-api3.dev.bgdi.ch
 LINKEDDATAHOST=https://geo.ld.admin.ch
 MODWSGI_CONFIG=development.ini
+NUM_PROC=0
```
ltshb, line 19: I would use the same name as in `infra-e2e-tests`; `PROCESS=0` for consistency

ltshb (approved):
You might want to either remove `coverage` from the dev dependencies or to add it to the nose2 config 
```
[coverage]
always-on = True
```
If it works with nose2, I would keep it.

tests/nose2.cfg
```diff
@@ -17,4 +17,11 @@ clear-handlers = True
 log-level = DEBUG
 
 [multiprocess]
-always-on = True
\ No newline at end of file
+always-on = True
+test-run-timeout = 600
```
hansmannj, line 21: I did not find a lot of hints on this in the docu. Hope it is the same as `process-timeout` in `nosetests`, but am no exactly sure about this.

tests/nose2.cfg
```diff
@@ -0,0 +1,27 @@
...
+log-level = DEBUG
+
+[multiprocess]
+always-on = True
+test-run-timeout = 600
+
+[coverage]
+always-on = True
+coverage-report = term \
+                  xml
```
ltshb, line 26: The coverage report is missing from the ci logs ? Probably due to these lines, why did you added them ?

tests/nose2.cfg
```diff
@@ -0,0 +1,27 @@
...
+cover-package = chsdi
```
ltshb, line 27: This is redundant with the .coveragerc file

hansmannj: True, it has always been redundant though. In the .coveragerc and also in setup.cfg. When I deleted it from setup.cfg I moved it to nose2.cfg. But I guess once is enough and I'll try if it works when dropping it from nose2.cfg

tests/nose2.cfg
hansmannj: To have both, a report in terminal and also a more detailed report as .xml file.
I'll check why the report is missing in ci, that's strange

hansmannj: Still not in the CodeBuild output, but it appears in the dedicated report group now at least.
Digging further

hansmannj: Now we have the coverage report in CodeBuilds terminal output (but therefore no longer in an xml file). For me its fine like that. OK for you?

ltshb (changes requested):
I'm a bit wondering how the coverage report in codebuild appears ?
image
The test report is due to the configuration in buildspec:
```yaml
reports:
  reportGroupJunit:
    files:
      - "junit-reports/**/*"
    file-format: JunitXml
```
We have a ticket to do some cleanup in the codebuild test reports, see https://jira.swisstopo.ch/browse/BGDIINF_SB-2595, in this ticket a proposal was to use `reports` instead of `reportGroupJunit`, but if we also have coverage report (which is not a bad idea to have) we might want to use `test` instead of `reportGroupJunit` and `coverage` instead of `reportGroupCoverage`. The question is how the coverage group name is configured ? for junit based on the docu it is the name in the buildspec, from here above `reportGroupJunit` appended to the project name.

hansmannj:
> I'm a bit wondering how the coverage report in codebuild appears ? image The test report is due to the configuration in buildspec:
> 
> ```yaml
> reports:
>   reportGroupJunit:
>     files:
>       - "junit-reports/**/*"
>     file-format: JunitXml
> ```
> 
> We have a ticket to do some cleanup in the codebuild test reports, see https://jira.swisstopo.ch/browse/BGDIINF_SB-2595, in this ticket a proposal was to use `reports` instead of `reportGroupJunit`, but if we also have coverage report (which is not a bad idea to have) we might want to use `test` instead of `reportGroupJunit` and `coverage` instead of `reportGroupCoverage`. The question is how the coverage group name is configured ? for junit based on the docu it is the name in the buildspec, from here above `reportGroupJunit` appended to the project name.

I already removed it. I was configured in buildspec, too, as the unit test reporting.
But I did not find a way to have both, an xml report used for the CodeBuild reports section and a terminal output.
That's why there was configured both:
```
coverage-report = term \
                  xml
```

But for some reasons, there will be no terminal output in CodeBuild, when configured this way.
So we can either have the coverage report in the terminal output OR as report group.

Makefile
```diff
@@ -310,27 +313,55 @@ dockerpull:
 
 .PHONY: test
 test: guard-VENV clean_logs local-templates build
```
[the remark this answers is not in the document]
hansmannj: Probably makes sense. They were splittet, at least on directory level, but executed within the same target before.
Having two targets may be handy. I'll do that

hansmannj:
Found a way to have coverage reports in terminal when local, and in CodeBuild report gropups for the CI.
This needs separate .coveragerc files for the ci-integraion and ci-functional tests (.coveragerc_ci_int and .coverage_ci_func).
Seems to work as expected

Makefile
```diff
@@ -125,7 +128,8 @@ help:
 	@echo -e "\033[1mFORMATING, LINTING AND TESTING TOOLS TARGETS\033[0m "
 	@echo "- lint/autolint      Python code quality assurance"
 	@echo "- shell              Pylons shell (for debugging)"
-	@echo "- test               Functional and integration nose tests"
+	@echo "- test-functional    Functional nose2 tests"
+	@echo "- test-integration   Integration nose2 tests"
```
ltshb, line 132: Keep also the `test` target in the help.

Makefile
```diff
@@ -308,29 +312,65 @@ dockerpull:
...
+	@echo "${GREEN}Unit testing (functional tests)...${RESET}";
+	${PYTHON} ./scripts/pg_ready.py
+
+	${NOSE} \
+		-s tests/functional \
+		-t $(PROJECT_PATH) \
+		--verbose \
+		-c tests/nose2.cfg \
+		--junit-xml-path junit-reports/functional/nosetest.xml \
+		-N ${PROCESS}
```
ltshb, line 340: I would add a shortcut to run both target
```Makefile
.PHONY: test
test: test-functional test-integration
```

buildspec-py3.yml
```diff
@@ -193,7 +193,15 @@ phases:
...
       - "junit-reports/**/*"
     file-format: JunitXml
+  cov_func:
+    files:
+      - "coverage_functional.xml"
+    file-format: COBERTURAXML
+  cov_int:
+    files:
+      - "coverage_integration.xml"
+    file-format: COBERTURAXML
```
ltshb, line 207: Here I would only use one group for coverage in Codebuild. Otherwise we will end up have hundreds of groups.

```yaml
reports:
  cover:
    files:
      - "coverage*.xml"
    file-format: COBERTURAXML
```

ltshb: If having multiple files for coverage in Codebuild is an issue, then I'll drop out the coverage report in codebuild.

Makefile
hansmannj: Good idea!

hansmannj:
Combining both xml files (ci-integration and ci-functional) into one reportGroup for CodeBuild results in every file for which coverage is reported appears twice. Once from the integration and once from the functional test coverage report xml.
IMHO that's ok and still better than having no information at all. Agree?

ltshb (approved):

PR #4057, opened 2022-10-25
Fixing logging for Kibana

ltshb:
In Kibana the request.statusCode field is used as integer and not as string.

Also changed the query_string nomenclature to be consistent.

K8s logging needs to be changed see https://github.com/geoadmin/infra-kubernetes/pull/286


ltclm (approved):
thank you @ltshb !

rebert (approved):
lgtm

PR #4058, opened 2022-10-26
Fixed app_version being set to 0000-00-00-rc0- on new release

ltshb:


rebert (approved):
@ltshb you are fast. thanks

PR #4059, opened 2022-11-03
Use Pipfile.lock on the CI

ltshb:
This avoid using other package version on the CI as local to avoid having
test working locally but failing on CI.


hansmannj (approved):
Nice idea :+1:

PR #4064, opened 2022-11-15
BGDIDIC-2472: update legends

faselm:


ltclm (changes requested):
might be due to my age but i cant read anything on these images.
if the data owner did not deliver custom legends, the translations have to be shortened.

faselm:
The customer will deliver the legends at some point. I think the legends can stay like that according to the message from Mario regading the legend format: "Yes, indeed, they decided to deliver some png legends for map.geo (waiting for them). Nevertheless it makes sense to keep and translate the current legend for the wms users".

PR #4065, opened 2022-11-16
BGDIDIC-2494: add legends

faselm:


ltclm (changes requested):
de and fr are mandatory and the translations have to be delivered.

faselm:
But where? all translations were inserted into ms_class (all in German for now even for fr, it, en, rm)

ltclm:
yes that's the point. the data owner has to deliver french tranlsation.

faselm:
Does it has to put it itself into the DB?

ltclm:
> Does it has to put it itself into the DB?

Hi marc, please ask @mariokeusen for the missing translations and update the table ms_class. Thank you!

faselm:
The test is passing now. Does that mean the data provider has put the translation? where can I find it?

ltclm:
the legend translation is provided by dataowner / mario DIF to you. you have to update ms_class and update the legends.
There are no tests on the translation.

rebert (approved):
thx

ltclm (approved):
thanks!

PR #4066, opened 2022-11-17
BGDIDIC-2472: update legends

faselm:


ltclm (approved):
thank you!

rebert (changes requested):
The width of the legend images should not be larger than 480px

rebert (approved):

PR #4067, opened 2022-11-18
BGDIDIC-2472: add field translations

faselm:


chsdi/templates/htmlpopup/einschraenkungen_drohnen.mako
```diff
@@ -31,6 +32,16 @@
...
+    auth_url = 'auth_url_%s' % lang
+    auth_name = 'auth_name_%s' % lang
+    auth_service = 'auth_service_%s' % lang
+    auth_url_txt = ' / '.join(c['attributes'][auth_url]) if c['attributes'][auth_url] else '-'
+    auth_name_txt = ' / '.join(c['attributes'][auth_name]) if c['attributes'][auth_name] else '-'
+    auth_service_txt = ' / '.join(c['attributes'][auth_service]) if c['attributes'][auth_service] else '-'
+    auth_contact = ' / '.join(c['attributes']['auth_contact']) if c['attributes']['auth_contact'] else '-'
+    auth_email = ' / '.join(c['attributes']['auth_email']) if c['attributes']['auth_email'] else '-'
+    auth_phone = ' / '.join(c['attributes']['auth_phone']) if c['attributes']['auth_phone'] else '-'
+    auth_intervalbefore = ' / '.join(c['attributes']['auth_intervalbefore']) if c['attributes']['auth_intervalbefore'] else '-'
```
rebert, line 44: are you sure, that ' / '.join will return the right value? I tested ' / '.join('har'), which leads to h / a / r

chsdi/templates/htmlpopup/einschraenkungen_drohnen.mako
```diff
@@ -31,6 +32,16 @@
     zone_name = 'zone_name_%s' % lang
     zone_restriction = 'zone_restriction_%s' % lang
     zone_message = 'zone_message_%s' % lang
+    auth_url = 'auth_url_%s' % lang
```
rebert, line 35: does the db guarantee that NULL values are not allowed here?

rebert (changes requested):
In general I would adapt the translations to the right layer:
ch.bazl.gebietsbeschraenkungen-drohnen.xxxx should be ch.bazl.einschraekungen-drohnen.xxx

chsdi/templates/htmlpopup/einschraenkungen_drohnen.mako
faselm: yes, because I think join deals with the array of string(s) coming from the db differently that the array of characters in a string. At least the values are OK in the test link (1 value when the array has 1 string and many values separated by / when the array has multiple strings)

chsdi/templates/htmlpopup/einschraenkungen_drohnen.mako
faselm: this is tested later I think with:

' / '.join(c['attributes'][auth_url]) if c['attributes'][auth_url] else '-'

faselm:
> In general I would adapt the translations to the right layer: ch.bazl.gebietsbeschraenkungen-drohnen.xxxx should be ch.bazl.einschraekungen-drohnen.xxx

I renamed the layer in the db and in the mako

chsdi/templates/htmlpopup/einschraenkungen_drohnen.mako
rebert: @faselm yes, indeed - in the extended part it is being tested. In the regular tooltip not.

rebert (approved):

chsdi/templates/htmlpopup/einschraenkungen_drohnen.mako
faselm: I've just added it

PR #4074, opened 2022-11-23
BGDIDIC-935: add new attribute gkat to basic tooltip

ltclm:
replace data.geo.admin.ch download links with madd link


rebert (approved):
LGTM

PR #4079, opened 2022-11-29
BGDIDIC-2472: update legends

faselm:


ltclm (approved):
the english image is not transparent. besides that looks good to me.

rebert (changes requested):
thanks. but it seems, that _en is not transparent

rebert (approved):

PR #4089, opened 2022-12-08
Revert "Updating milestone branch from Master"

ltclm:
Reverts geoadmin/mf-chsdi3#4088


ltshb (changes requested):
I would not do a revert, but try to understand why it is broken otherwise we will likely have the same issue during the deploy day.

PR #4090, opened 2022-12-08
Fixing pool pre ping default value.

ltshb:
When no pool pre ping environment variable is defined then the config
`sqlalchemy.pool_pre_ping` is an empty string.


ltclm (approved):
thank you

PR #4095, opened 2022-12-16
BGDIINF_SB-2589: global config for aerialimages and histmaps

rebert:
- The data path towards histmaps and aerialimages is set in the env

- Usefull for migration and testing towards sys-aerialimages and sys-hist technical URLs


ltclm (approved):
thank you!

PR #4098, opened 2023-01-16
BGDIDIC-1526: decommission ch.astra.strassenverkehrszaehlung_messstel…

faselm:


ltclm (commented):
is it only the legends that can be removed?

faselm:
> is it only the legends that can be removed?

I found the model/templates that didn't show up during my first search...

ltclm (approved):

PR #4101, opened 2023-01-17
BGDIINF_SB-2704: set custom cache headers for 4xx and 5xx responses

ltclm:


chsdi/response_callbacks.py
```diff
@@ -11,6 +11,17 @@ def add_default_cache_control(request, response):
...
+    # overwrite with these 4xx/5xx cache settings
+    # no cache on these 5xx errors, they are supposed to be temporary
+    if response.status_code in (502, 503, 504, 507):
+        response.headers['Cache-Control'] = 'no-cache'
+    # short cache duration for other 5xx errors
+    elif response.status_code >= 500:
+        response.headers['Cache-Control'] = 'public, max-age=10'
+    # use cloudfront default caching (ttl 10s) for 4xx errors
+    elif response.status_code >= 400:
+        response.headers.pop('Cache-Control', None)
```
ltshb, line 23: This is wrong the default TTL of cloudfront is 86400 seconds ! I would not rely on Cloudfront configuration in services.
For 400 I would use the same default cache setting from the environment variable see line 11.

chsdi/response_callbacks.py
```diff
@@ -11,6 +11,17 @@ def add_default_cache_control(request, response):
         dft_cache_control = request.registry.settings['default_cache_control']
         response.cache_control = CacheControl.parse(dft_cache_control, type='response')
 
+    # overwrite with these 4xx/5xx cache settings
+    # no cache on these 5xx errors, they are supposed to be temporary
+    if response.status_code in (502, 503, 504, 507):
+        response.headers['Cache-Control'] = 'no-cache'
+    # short cache duration for other 5xx errors
+    elif response.status_code >= 500:
+        response.headers['Cache-Control'] = 'public, max-age=10'
```
ltshb, line 20: In `pyramid` you cannot directly set the Cache-Control header it will not work as expected, you need to do it via `cache_control` property see line 12.

rebert (approved):
thx

ltshb (approved):
👍🏼

PR #4133, opened 2023-02-20
BGDIINF_SB-2794: Fix the rss newsfeed of the releasenotes

rebert:


procrastinatio (approved):
OK. But one day, we should do it differently.

rebert:
@procrastinatio Absolutely - I hope too, thaw we will do it differently one day.

PR #4147, opened 2023-03-02
BGDIDIC-1709: hydroweb turnaround with vectorStyles due to schedule

rebert:


marionb (approved):
It looks OK to me!
should I test this in some form?

rebert:
@marionb thanks for approving. I will test it directly on dev

PR #4151, opened 2023-03-09
BGDIDIC-1789: Rounding coordinates and alti to .3f

rebert:
- The datenherr wishes to have the coordinates rounded on .3f

  This will only be visible in the tooltip. The data and the model

  are currently holding the easting and northing together as a string


chsdi/templates/htmlpopup/fixpunkte.mako
```diff
@@ -9,13 +9,27 @@
       return value if c['layerBodId'] in result else result
 %>
 </%def>
+<%def name="fengshui_coord(coord)">
```
marionb, line 12: Interesting name :+1:

marionb (approved):
code LGTM thanks!

PR #4152, opened 2023-03-09
BGDIDIC-1789: Fixing tooltip when no alti is provided

rebert:


chsdi/templates/htmlpopup/fixpunkte.mako
```diff
@@ -22,14 +22,22 @@
       return coord 
 %>
 </%def>
+<%def name="format_alti(alti)">
+<%
+    if alti is None:
+      return '-'
+    else:
+      return '{0:.3f}'.format(alti)
```
marionb, line 30: You don't need this `else` since you already have a return statement in the `if`
And If you want to compress it to one line you can write
```suggestion
return '-' if alti is None else '{0:.3f}'.format(alti)
```

marionb (approved):

chsdi/templates/htmlpopup/fixpunkte.mako
marionb: But it will also work what you have ;-)

PR #4163, opened 2023-03-22
BGDIINF_SB-2838: Avoid crash in mako due to 104 Connection Broken Exception

ltshb:
We had a one time failure in this mako during E2E test which was triggered by
the infamous 104 Connection Broken.

So I added proper exception handling here, mf-chsdi3 should not crash if another
service is not available !

### DIFF
--- a/chsdi/templates/htmlpopup/windatlas50.mako
+++ b/chsdi/templates/htmlpopup/windatlas50.mako
@@ -1,12 +1,22 @@
 <%inherit file="base.mako"/>
 <%!
+import logging
 import requests
+from requests.exceptions import RequestException
+logger = logging.getLogger(__name__)
+
 def getAltitude(baseUrl, center):
-    fullUrl = 'http://' + baseUrl +'/rest/services/height'
-    response = requests.get(
-        fullUrl + '?easting=%s&northing=%s&elevation_model=COMB' % (center[0], center[1]),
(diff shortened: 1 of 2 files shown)


rebert:
@ltshb Merci 🚀

chsdi/templates/htmlpopup/windatlas50.mako
```diff
@@ -1,12 +1,22 @@
...
+    query = '?easting=%s&northing=%s&elevation_model=COMB' % (center[0], center[1])
+    try:
+        response = requests.get(
+            url + query,
+            headers={'Referer': url, 'User-Agent': 'mf-geoadmin/python'}
+        )
+        result = response.json()
+    except (RequestException, ConnectionError) as err:
+        logger.exceptions('Failed to get the altitude: %s', err)
+        return 0
```
rebert, line 19: ok, for them who know that 0 can't be a realistic height... in the tooltips we ofthen use String - to indicate that we could not obtain a value

ltshb: The method `getAltitude()` is expecting to return a number (see lines that call the method). If the number is `0` it will be displayed with `-`. Not a great code as in theory altitude of 0 is valid (not in switzerland) but I don't want to change too much this code.

rebert (approved):
You are right - zero (0) is `falsy`. Thanks for the fix.

PR #4166, opened 2023-03-28
BGDIINF_SB-2158: update license

ltkum:
We are harmonizing licenses across all of geoadmin public repositories, with a standard BSD 3-Clause License. Thus using this standard template.


ltshb (changes requested):
I would not change existing license ! Depending on the license this is not legally allowed.

PR #4171, opened 2023-04-05
bug_BGDIINF_SB-2831 Using the external WMTS GetCapabilities 

procrastinatio:
... on the same environment (dev, int and prod) as the one used to serve the doc


rebert (approved):
thanks!

PR #4175, opened 2023-04-13
Feat bgdidic 2583 swisstne base

faselm:
This pr should be good however the tooltips work when testing with:

http://localhost:9178/rest/services/ech/MapServer/ch.swisstopo.swisstne-base/1/htmlPopup

but not when clicking on the map:

https://mf-geoadmin3.dev.bgdi.ch/index.html?api_url=%2F%2Flocalhost:9178&lang=de&topic=ech&bgLayer=ch.swisstopo.pixelkarte-grau&layers=ch.swisstopo.swisstne-base&E=2583445.96&N=1196859.23&zoom=8

with error: CORS Missing Allow Origin

Does it works for you? any idea from where this kind of error could come from?


faselm:
OK I got it:
Object of type UUID is not JSON serializable
I will look at it

faselm:
The pr should be able to be merged however as only a table column need to be converted

PR #4180, opened 2023-04-27
BGDIDIC-1490: add link translations

faselm:


ltclm (approved):
great, thats going into the right direction.
we can show the same attributes / information in getfeatureinfo/htmlpopup and WMS GetFeatureInfo 👍

rebert (approved):
👍

PR #4181, opened 2023-04-27
BGDIDIC-2331: link_original will be provided as full link by data-owner

ltclm:
* service parity wms getfeatureinfo = identify = htmlpopup

new content in data has to be confirmed by dataowner before merge and deploy of this


consta-v:
Thanks @ltclm ! It looks good to me. We have to wait for a delivery from the geology before merging this change. I don't know yet when they will be able to deliver the corrected data (especially if it will be before May 10th or not). I'll keep you informed.

PR #4182, opened 2023-04-27
BGDIDIC-1432: update legends

faselm:
Have the big legends to be in .pdf or .png also work?


ltclm (changes requested):
i think the legends have to be transparent, but im not sure if that's still the case.
@rebert do you know that?

faselm:
Hi, does it need to be transparent then?

faselm:
Hi, can I merge this pr or do I have to ask the data provider to send transparent legends?

ltclm (approved):
yes @faselm we already have plenty of non-transparent legends.
this seems to be no longer a requirement.

PR #4186, opened 2023-05-03
BGDIDIC-2590: make kantonid as primary key

faselm:


ltclm (approved):
👍

faselm:
This one should be good now

PR #4196, opened 2023-05-24
BGDIDIC-997: fix download link

faselm:


rebert (approved):
of course this has to do with the data. But having a complete URL in the data, will allow that this attribute could be served in the GetFeatureInfo of the WMS as well

faselm:
@rebert OK, thanks I will change it in the fme instead

PR #4197, opened 2023-05-25
Doc cleanup regarding 2D/3D vectortiles

procrastinatio:
@stebie What hostnames do we want to publish for `vectortiles`?

@stebie I did also fix the `3D` terrain and buildinds example (https://codepen.io/geoadmin/pen/zBEYGE?editors=0010) and use a current version of `cesium` (at the time of writing). The tiles are still supported, but many attributes are labelled as deprecated (so won't be supported any soon).


rebert (approved):
👌

PR #4205, opened 2023-06-14
BGDIDIC-1293: update field names

faselm:


ltclm (approved):
please make sure that 
* the columns will be renamed in all the other services too (wms, wmts, sphinx) and in the integration scripts
*  the old columns will be removed after the deploy. please add a comment / task to the remember when deploy document with the clean-up sql statement

thank you

PR #4213, opened 2023-06-21
BGDIDIC-1709: new legends 4 hydroweb-messstationen_temperatur

rebert:


ltclm (approved):
they are not transparent. but that's not mandatory anymore i think.

PR #4219, opened 2023-06-26
BGDIINF_SB-2950: by default pyramid did not catch all exception and log them

ltshb:
By default pyramid did not logged the exception, they were simply printed on the
console. Except when running locally with waitress server which logged the exception
using the logging configuration.

This made the debugging of crash on PROD harder as we needed to look into the
non json index in kibana and that each backtrace line was treated as a separate
log.

Now with the pyramid_exclog package exception are logged based on the logging
configuration and the exc_info flag is set to true and the backtrace is set
into the field exc_text. This will help to debug crash with kibana.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -32,6 +32,7 @@ simplejson = "~=3.17"
 unidecode = "~=1.3"
 Akhet = "~=2.0"
 logging-utilities = "~=3.1.1"
+pyramid-exclog = "~=1.1"
 
 [dev-packages]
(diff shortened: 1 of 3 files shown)


rebert (approved):
merci

ltshb:
NOTE: I tested it locally by adding a temporary view that crashed and as mentioned above when using only waitress this fix is not needed but when using mod_wsgi not proper log was done, now the pyramid_exclog does

PR #4226, opened 2023-07-10
BGDIDIC-2617: delete hyperlink field

faselm:


ltclm (approved):
thank you,
could you please remove the column from the table/view as post deploy task.

PR #4231, opened 2023-08-16
BGDIINF_SB-3049: considering infinity to be an invalid coordinate for…

rebert:
… geometries


ltclm (approved):
thank you.
but it would be good to know how this regression has been introduced...
if tests are passing, its good for me!

PR #4240, opened 2023-08-24
BGDIDIC-2599 update legend

marionb:


rebert (approved):
I can't see any difference between the two images :-)

marionb:
@rebert thanks - I forgot to update the the translations so now change is more visibel ;-)

PR #4254, opened 2023-08-31
BGDIINF_SB-3066: Legal notice

rebert:


chsdi/static/doc/source/conf.py
```diff
@@ -176,7 +177,7 @@
 #html_show_sourcelink = True
 
 # If true, "Created using Sphinx" is shown in the HTML footer. Default is True.
-#html_show_sphinx = True
+html_show_sphinx = True
```
stebie, line 180: This is intended?

rebert: yes, it changes nothing. but it is more explicite

ltshb (approved):
LGTME, could you add a quick printscreen of the output ?

ltclm (approved):

PR #4262, opened 2023-09-12
BGDIINF_SB-3033: Remove versions from doc vectortiles

rebert:


chsdi/static/doc/source/services/sdiservices.rst
```diff
@@ -1340,9 +1340,6 @@ A previous versioned style request is in the following form:
     <Scheme>://<ServerName>/styles/<layername>/<version>/style.json
```
ltflb-bgdi, line 1340: Should this not be removed as well?

ltflb-bgdi (approved):
LGTM

PR #4268, opened 2023-09-25
BGDIDIC-1256: add field gwr_egid

faselm:


ltclm (changes requested):
lgtm but can you please ask if htmlpopup is really not needed?

PR #4270, opened 2023-10-03
Feat bgdidic 2651 geologie reflexionsseismik

faselm:


ltclm (approved):
lgtm, thank you.
What about the column `access_int`, is is not used in the htmlpopup i think it can be removed from the model too.

PR #4277, opened 2023-10-26
BGDIDIC-1296 update legend png

marionb:


ltclm (changes requested):
we still have the limitation that png legends in the legend response should not be wider than 480px:

https://api3.geo.admin.ch/rest/services/all/MapServer/ch.bazl.segelflugkarte/legend?lang=de

If it is necessary to create more detailed legends, you have to create a thumbnail png with max 480px and a _big.pdf for each language with the detailed view.

p.e. ch.bazl.segelflugkarte
you also have to add this layer to this static list:
https://github.com/geoadmin/mf-chsdi3/blob/develop-2023-11-01/chsdi/templates/legend.mako#L8

marionb:
@ltclm thanks for the hint. A simple resize of the image should do :-)

ltclm (approved):
great, thank you

PR #4301, opened 2023-12-11
BGDIDIC-1057: use ident field instead of bgdi_id

faselm:


rebert (commented):
As I understood the idea is to add the link to object in the tooltip as well by adding
` <% c['stable_id'] = True %>` into the mako

rebert (approved):

PR #4318, opened 2024-02-05
BGDIDIC-2544: update legends

faselm:


ltclm (approved):

ltclm:
> Looks like some tests are failing here. I've just checked out from develop and updated the 5 legend images so I don't see what could go wrong with this pr (I still don't have access AWS to see the error)... make test returns an error:
> 
> /rest/services/all/MapServer/ch.swisstopo.lubis-luftbilder_farbe/19981551013722/extendedHtmlPopup
> 
> with 404 error though. Not sure if it's that that causes the problem in this pr

Salut @faselm 
This is a left-over of the lubis migration to stac. I have prepared a fix:
https://github.com/geoadmin/mf-chsdi3/pull/4319

you will have to rebase this pr when the fix has been merged.

PR #4323, opened 2024-02-07
BGDIDIC-2743: fix point filter

faselm:


chsdi/templates/htmlpopup/armasuisse_natur_landschaft_armee.mako
```diff
@@ -33,7 +33,7 @@
         <td class="cell-left">${_('ch.armasuisse.natur-landschaft_armee.schutz')}</td>
         <td>${c['attributes'][schutz_text] or '-'}</td>
     </tr>
-    %if c['attributes']['typ_de']:
```
marionb, line 36: looks like `typ_de` is not used any more as attribute so you can probably remove it from the model on line 197.

ltclm (approved):

PR #4326, opened 2024-02-13
Feat bgdidic 2763 standorte mobilfunkanlagen

faselm:


ltclm (approved):
the attribute koord is only used in identify? there is no reference to it in the htmlpopup.

PR #4344, opened 2024-03-01
BGDIDIC-2768: add view for popup window

faselm:


ltclm (commented):
why do we need this aggregation here?

ltclm (approved):

PR #4353, opened 2024-03-12
PB-130: HTTP 405 when method not allowed.

rebert:
- OPTIONS: empty responses are not being zipped anymore `content-length: 0`

- HTTP 405 on not in GET HEAD OPTIONS

### DIFF
--- a/chsdi/__init__.py
+++ b/chsdi/__init__.py
@@ -34,16 +34,6 @@ class WsgiSchemeAdaptedRequest(Request):
         super().__init__(environ, **kwargs)
 
 
-# This is a wrapper function around all views. If OPTIONS is given, an empty string will be returned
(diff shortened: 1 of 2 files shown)


ltclm (approved):
thank you :+1:

chsdi/__init__.py
```diff
@@ -58,7 +48,20 @@ def main(global_config, **settings):
...
+    def options_view(view, info):
+        def wrapper_view(context, request):
+            if request.method == 'OPTIONS':
+                return Response()
+            # HTTP 405: method not allowed
+            elif request.method not in request_method:
+                context.code = context.status_code = 405
+            return view(context, request)
+        return wrapper_view
+
```
ltshb, line 64: Because it is not only for options I would rename it, maybe something like `def methods_view`

ltshb (approved):

PR #4390, opened 2024-06-17
BGDIDIC-2837: add legends

faselm:


rebert (approved):
is it ok when the text is aligned at he buttom of the icon?

PR #4393, opened 2024-06-20
BGDIDIC-2272: update timeseries arealstatistik-bodenbedeckung

ltclm:
* filter identify results with timeInstant
* change feature id to stable_id for arealstatistik


rebert (approved):
:+1:

PR #4398, opened 2024-06-24
BGDIDIC-2832: remove decimals from integer values in mako

ltclm:


chsdi/templates/htmlpopup/hartsteinvorkommen.mako
```diff
@@ -94,7 +94,7 @@
     </tr>
     <tr>
       <td class="cell-left">${_(f'{layerid}.elevation')}</td>
-      <td>${c['attributes']['elevation'] or '-'}</td>
+      <td>${int(round(c['attributes']['elevation'])) or '-'}</td>
```
marionb, line 97: I did not test this explicitly here but I would guess that this gives an error if `c['attributes']['elevation']` is `None`.

Here my alternative suggestion:
```
int(round(c['attributes']['elevation'])) if c['attributes']['elevation'] is not None else '-'
```

ltclm: thank you :+1: 
i have changed it

marionb (approved):
:+1:

PR #4401, opened 2024-06-26
BGDIDIC-2837: convert float to int

faselm:


ltclm (changes requested):
please also do as proposed by @marionb 
```
int(round(c['attributes']['elevation'])) if c['attributes']['elevation'] is not None else '-'
```
seems to be more robust if the attribute contains NULL values

chsdi/templates/htmlpopup/vbs_ladestationen.mako
```diff
@@ -46,11 +46,11 @@
     </tr>
     <tr>
         <td class="cell-left">${_('ch.vbs.ladestationen.XKoordinate')}</td>
-        <td>${c['attributes']['x_koordinate'] or '-'}</td>
+        <td>${int(round(c['attributes']['x_koordinate'])) if c['attributes']['x_koordinate'] is not None else '-'}</td>
```
marionb, line 49: Don't these coordinates already come as int values? When I look at the the layer in dev they are always `xyz.0`
In this case I think it makes more sens to integrate them in t the DB as int so you don't have to round them here and use the actual values the data owner provides?

rebert (approved):

chsdi/templates/htmlpopup/vbs_ladestationen.mako
marionb: Yes it is in the table that need to be changed as well as the model here:
- https://github.com/geoadmin/mf-chsdi3/blob/develop-2024-07-03/chsdi/models/vector/vbs.py#L409
- https://github.com/geoadmin/mf-chsdi3/blob/develop-2024-07-03/chsdi/models/vector/vbs.py#L417-L418

If you add these changes  you will not need this extra code and the make will be easier to read and maintain.

marionb (changes requested):
see my comment above

chsdi/templates/htmlpopup/vbs_ladestationen.mako
ltclm: yes, the values are integer in the identify result but the htmlpopup converts them for unknown reasons into x.0 numbers...

ltclm (approved):

marionb (approved):

PR #4403, opened 2024-06-26
BGDIDIC-2841: update rm legends

faselm:


ltclm (approved):

rebert (approved):
this is more my personal opinion. when we keep the legend small it is obvious that it is possible to click on it to get the pdf. This bigger version is still not readable... Personally I would keep the smaller one...

PR #4411, opened 2024-07-04
PB-764 made polyfill url configurable via env

boecklic:


pakb (approved):
As discussed with Stefan, IMHO we could ditch the polyfill thing, as it is there to care for browsers way past their age (2013 to 2015).

But what you did fix the issue for the time being, so good that way too

PR #4418, opened 2024-07-17
Hotfix for polyfill

boecklic:
@ltshb is this the right thing to do to bring the hotfix functionality back to the milestone branch? The idea is that we can use `develop-XXX.latest` again in kubernetes instead of the fixed tag


ltshb:
> @ltshb is this the right thing to do to bring the hotfix functionality back to the milestone branch? The idea is that we can use `develop-XXX.latest` again in kubernetes instead of the fixed tag

Yes and you HAVE to ignore the `Update branch` button 😉

ltshb (approved):

PR #4426, opened 2024-08-05
BGDIDIC-706: update legends

faselm:


rebert:
imo the text is to small

ltclm:
@rebert @faselm that was exactly what my concern about the removal of the character 255 limitation was about, i think we should try to understand why the dataowner needs more than 255 character for a class or attribute name.
https://jira.swisstopo.ch/browse/BGDIDIC-1330

rebert:
or the contextual info can be added as text in the infobox...

PR #4435, opened 2024-09-01
PB-885: use new datasource for swissimage hist

ltclm:


rebert (approved):
:+1:

ltshb:
@ltclm can I use this PR as test for the CI migration ? See https://github.com/geoadmin/infra-terraform-bgdi-builder/pull/283, so as soon as this pr is approved I'll apply it and then we can use this PR to trigger a PR build and then a release build to see if it works.

PR #4450, opened 2024-10-11
BGDIDIC-701: delete model

faselm:


ltclm (changes requested):
please also remove the legends

rebert (approved):
merci

PR #4467, opened 2024-11-01
PB-974: Migration of the API for open transport data from TIRAS to OJP 2.0

hansmannj:
The TRIAS API which we use for the open transport stop data will be decomissioned soon, hence we migrate to OJP 2.0

### DIFF
--- a/chsdi/lib/opentransapi/opentransapi.py
+++ b/chsdi/lib/opentransapi/opentransapi.py
@@ -5,13 +5,15 @@ import xml.etree.ElementTree as et
 from pytz import timezone
 from datetime import datetime
 from dateutil import tz
+import re
 
 
 class OpenTrans:
 
     def __init__(self, open_trans_api_key):
         self.open_trans_api_key = open_trans_api_key  # Get API key from config .ini
-        self.url = 'https://api.opentransportdata.swiss/trias2020'  # URL of API
+        self.url = 'https://api.opentransportdata.swiss/ojp20'  # URL of API
+        self.station_id = None
 
     def get_departures(self, station_id, number_results=5, request_dt_time=False):
         if not request_dt_time:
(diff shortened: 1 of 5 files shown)


rebert:
LGTM. The CI seems not to have the credentials to access OJP 2.0. We should put the new credential to SSM, and update the CI towards this. Maybe we will have to take care about the transition. K8s pods will have to use the new key as well.

chsdi/lib/opentransapi/opentransapi.py
```diff
@@ -5,13 +5,15 @@
...
 from dateutil import tz
+import re
 
 
 class OpenTrans:
 
     def __init__(self, open_trans_api_key):
         self.open_trans_api_key = open_trans_api_key  # Get API key from config .ini
-        self.url = 'https://api.opentransportdata.swiss/trias2020'  # URL of API
+        self.url = 'https://api.opentransportdata.swiss/ojp20'  # URL of API
```
ltshb, line 15: Should come from environment variable and not hardcoded here.

chsdi/lib/opentransapi/opentransapi.py
```diff
@@ -41,75 +55,87 @@ def _check_element(self, el_name, el):
...
+                'estimatedDate': self._convert_estimated_date(el.find('.//ojp:ServiceDeparture/ojp:EstimatedTime', ns)),
                 'destinationName': el_destination_name,
-                'destinationId': el_destination_id})
+                'destinationId': el_destination_id
+            })
 
         return results
 
+    def create_ojp_payload(self, station_id, request_dt_time, number_results=5):
+        payload = f"""<?xml version="1.0" encoding="UTF-8"?>
```
ltshb, line 93: White space should be stripped before sending to the API to save bandwith

ltshb (changes requested):
Can we easily detect the version of the api.pentransportdata ? if yes we should handle both the old and new version. This way it will make the transition much easier as using the same container with different configuration can be used with both api version.
If not I propose to do it via configuration variable. 
So add two environment variable:
- OPENTRANSPORT_API_URL
- OPENTRANSPORT_API_VERSION

Also we need to make sure that CI uses at least the new version and has the permission for it. But the best practice would be that unit test don't uses the service but mock the service to avoid external data access which can trigger CI failure due to external service failure/unavailibility.

hansmannj:
> Can we easily detect the version of the api.pentransportdata ? if yes we should handle both the old and new version. This way it will make the transition much easier as using the same container with different configuration can be used with both api version. If not I propose to do it via configuration variable. So add two environment variable:
> 
> * OPENTRANSPORT_API_URL
> * OPENTRANSPORT_API_VERSION
> 
> Also we need to make sure that CI uses at least the new version and has the permission for it. But the best practice would be that unit test don't uses the service but mock the service to avoid external data access which can trigger CI failure due to external service failure/unavailibility.

Implemented stripping of whitespaces, taking the URL form the env. Not sure what you mean by version detection.
I think we should only support one version at a time, and we have the control which version we use.
Agreed, we need to be careful when switching.

chsdi/lib/opentransapi/opentransapi.py
```diff
@@ -41,75 +55,89 @@ def _check_element(self, el_name, el):
...
+        <OJP xmlns='http://www.vdv.de/ojp' xmlns:siri='http://www.siri.org.uk/siri' version='2.0' xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance' xsi:schemaLocation='http://www.vdv.de/ojp ../../../../OJP4/OJP.xsd'>
+            <OJPRequest>
+                <siri:ServiceRequest>
+                    <siri:RequestTimestamp>{request_dt_time}</siri:RequestTimestamp>
+                    <siri:RequestorRef>swisstopo_Abfahrtsmonitor</siri:RequestorRef>
```
hansmannj, line 98: This RequestorRef was suggested by the OJP team

PR #4487, opened 2024-11-12
PB-974: Migration from trias to ojp for transport stop requests

rebert:


hansmannj:
Guess this has been moved to another PR by @rebert during deploy, when changing the base branch.

rebert:
yes, it is not possible to merge develop towards master with open prs on develop

hansmannj:
> yes, it is not possible to merge develop towards master with open prs on develop

Yes, sorry for bugging the deploy process and thanks for moving the commmits :pray:

ltshb:
> > PR title should be updated and CI is not passing
> 
> Yes, need to adapt CI to use the new secrets, will hopefully find some time for that, soon. Can you in the meantime have a look at the unit tests please? Is the mocking what you asked for in the comments in the old PR? One second please, just updating the tests. In one case, currently way too much is mocked. Will take some time, I'll give you a ping @ltshb once I am ready. Sorry, only noticed this now. Guess I'll only mock "send_post()"

Well by mocking we should not need any credential because no access to the real service should be done, unit test should not do any access to the service so they should pass even if the service is down

hansmannj:
@rebert @ltshb : Guess now this PR is ready for review.
I tried to only mock "send_post" and condensed the tests to those that seem to make sense.
There's one left, that can be deleted, too, as it only tests an exception we provoke using `side_effects()`.
Now I need to update CI to use the new SSM secrets.

rebert (commented):
lgtm. but I'm not allowed to approve :-)

tests/integration/test_stationboard.py
```diff
@@ -1,47 +1,130 @@
...
+                    <siri:ResponseTimestamp>{now}</siri:ResponseTimestamp>
+                    <OJPStopEventDelivery>
+                        <siri:ResponseTimestamp>{now}</siri:ResponseTimestamp>
+                        {departure_events}
+                    </OJPStopEventDelivery>
+                </siri:ServiceDelivery>
+            </OJPResponse>
+        </OJP>"""
+
+    @patch.object(opentransapi.OpenTrans, "send_post")
```
ltshb, line 82: Why mocking your library OpenTrans.send_post ? you don't fully test it this way. You should mock the requests library, you can use https://requests-mock.readthedocs.io/en/latest/overview.html to do so, this will can the post and return your mocked response and you can fully test your code.

ltshb: Also the opentrans API key should be removed from CI see https://github.com/geoadmin/infra-terraform-bgdi/blob/dd382ba8c135d491f8b55803925baba501973910/tf/aws/swisstopo-bgdi-builder/codebuild_projects/mf-chsdi3/buildspec.yml#L23C5-L23C22

ltshb: Is the key in SSM already the new one ? https://github.com/geoadmin/infra-terraform-bgdi/blob/dd382ba8c135d491f8b55803925baba501973910/tf/aws/swisstopo-bgdi-builder/shared/secrets/opentrans_api.tf#L1

ltshb (changes requested):
Unittest mocking needs adaptation as well as API key need to be removed from CI

tests/integration/test_stationboard.py
hansmannj: > Is the key in SSM already the new one ? https://github.com/geoadmin/infra-terraform-bgdi/blob/dd382ba8c135d491f8b55803925baba501973910/tf/aws/swisstopo-bgdi-builder/shared/secrets/opentrans_api.tf#L1

If it comes from this PR, then yes: https://github.com/geoadmin/infra-terraform-bgdi/pull/509

hansmannj: > Also the opentrans API key should be removed from CI see https://github.com/geoadmin/infra-terraform-bgdi/blob/dd382ba8c135d491f8b55803925baba501973910/tf/aws/swisstopo-bgdi-builder/codebuild_projects/mf-chsdi3/buildspec.yml#L23C5-L23C22. And in order to make sure that the mocking is working, the key MUST BE removed before merging this PR so we can verify it. Would be nice to link the key removal PR to this PR.

Guess removal of the old secrets from CI needs to be done all together with the deployment of the new release of mf-chsdi3. Otherwise the CI will be broken for some time, which would not be a nice state.

hansmannj:
Please re-check the PR.
Will prepare another PR for removing the old secrets from CI soon

chsdi/lib/opentransapi/opentransapi.py
```diff
@@ -21,10 +23,22 @@ def get_departures(self, station_id, number_results=5, request_dt_time=False):
...
+        except ValueError:
+            # sometimes the timestamp of the OJP 2.0 API's response has 7 digits for the
+            # milliseconds. 6 are expected and only 6 can be handled by Python.
+            # Hence we need to safely truncate everything between the last . and
+            # the +01:00 part of the timestamp, e.g.:
+            # 2024-11-01T15:39:45.5348804+01:00
+            # Use regex to capture and truncate everything between the last '.' and
+            # the first '+' to 6 digits
+            truncated_date_time = re.sub(r'(\.\d{6})\d*(?=\+)', r'\1', str_date_time)
+            date_time = datetime.strptime(truncated_date_time, '%Y-%m-%dT%H:%M:%S.%f%z')
```
ltshb, line 41: This function don't need to be a class member, or make it a static method (`@staticmethod`), this way you can reuse it in the unittest instead of copying it.

tests/integration/test_stationboard.py
```diff
@@ -1,47 +1,159 @@
...
+    to_zone = tz.gettz('Europe/Zurich')
+
+    try:
+        date_time = datetime.strptime(str_date_time, fmt)
+    except ValueError:
+        truncated_date_time = re.sub(r'(\.\d{6})\d*(?=\+)', r'\1', str_date_time)
+        date_time = datetime.strptime(truncated_date_time, '%Y-%m-%dT%H:%M:%S.%f%z')
+    date_time_utc = date_time.replace(tzinfo=from_zone)
+    date_time_zurich = date_time_utc.astimezone(to_zone)
+    return date_time_zurich.strftime('%d/%m/%Y %H:%M')
```
ltshb, line 27: See previous comment, don't copy this, reuse it.

tests/integration/test_stationboard.py
```diff
@@ -1,47 +1,159 @@
...
-        resp.mustcontain('The id must be an integer')
-
-    def test_stationboard_wrong_limit(self):
-        params = {'limit': 'toto'}
-        resp = self.testapp.get('/stationboard/stops/8501120', params=params, status=400)
-        resp.mustcontain('The limit parameter must be an integer')
+        self.mock_api_key = "dummy_api_key"
+        self.mock_url = "https://dummy-url.com"
+
+    def generate_mock_response(self, departures, now):
```
ltshb, line 35: Don't need to be class method here.

ltshb (commented):
There is still some issues with the unittest in the CI

chsdi/lib/opentransapi/opentransapi.py
hansmannj: Good idea :+1:

hansmannj:
> There is still some issues with the unittest in the CI

Thanks! Forgot to push the updated Pipfile(.lock), should now be fixed

ltshb (commented):
Code looks good but still have an issue with the CI

tests/integration/test_stationboard.py
```diff
@@ -1,47 +1,112 @@
...
+    def test_stationboard_wrong_limit(self, mock_get_current_registry):
+        mock_get_current_registry.return_value = self.mock_registry
+        self.mock_request.params["limit"] = "lalala"
+        self.mock_request.matchdict = {"id": "123"}
+
+        with self.assertRaises(HTTPBadRequest) as e:
+            _ = TransportView(self.mock_request)
+        self.assertEqual(str(e.exception), "The limit parameter must be an integer.")
+
+    @patch('chsdi.views.stationboard.get_current_registry')
```
hansmannj, line 110: This has been an original test.
Mocking it might be complicated, since if I understand the code correctly, the callback is added outside `TransportView()` on a more global level. Not exactly sure how to mock this. Any ideas @rebert @ltshb ?

ltshb: I think to keep it very simple the only thing you need to mock is the API call to the external service, no need to mock anything inside mf-chsdi3, simply mock the API at the request level. Of course you need to make sure that all API call are done via the `requests` library. So in the example above I don't understand the need to mock `chsdi.views.stationboard.get_current_registry` ?

hansmannj: According to my sparse understanding of the old Pyramid code, the code of the tested classes needs values from the application's global registry (API key and URL). The registry itself is not available during the unit tests, hence I mocked it. But there might be more elegant ways to do that or probably ways to access the app's registry even when testing.

hansmannj: I'll probably stick to the old `testapp` approach, as in the current tests for those tests

hansmannj (approved):
LGTM, but I am the wrong one to approve, since I cannot approve my own code :wink:

ltshb (approved):

PR #4495, opened 2024-11-26
PB-1222 : link to object time=None fix

pakb:
Do not add time information for layer that aren't time enabled.

Also changing the URL scheme/structure so that it uses the new way of describing a pre-selected feature and a layer timestamp


rebert (approved):
merci

ltclm (approved):

PR #4499, opened 2024-11-28
adapted lark transformer to work with v0.12

boecklic:
In newer versions of lark (newer than 0.7.8 that we used previously) `Token` is directly subclassing `str`, previously it was `Token(Str)` (https://github.com/lark-parser/lark/blame/44a20c850ea7e0945a9db663a763abdcf1f86b88/lark/lexer.py#L113). This led to a different behavior of the index operator `[]`.


ltclm (approved):
thank you @boecklic 
had to change the base to [feat_PB-1241_akhet](https://github.com/geoadmin/mf-chsdi3/tree/feat_PB-1241_akhet)

PR #4504, opened 2024-12-05
PB-1269, PB-1241: Fix cross domain policy

ltshb:
With the removal of Akhet library static files are directly served by Apache
and not anymore by pyramid so the patching of cross domain policy header needs
to be done by apache.

See https://github.com/geoadmin/mf-chsdi3/pull/4498 and https://github.com/geoadmin/mf-chsdi3/pull/4024


adk-swisstopo (approved):

ltclm (approved):
thank you :+1:

rebert (approved):
merci

PR #4521, opened 2025-01-21
BGDIDIC-1317 SIL remove rasterization

marionb:


marionb:
OK thanks @faselm  there seemed to be a space that I missed to remove. Removing it fixed the test/lint :-)

PR #4549, opened 2025-02-26
Data bgdidic 1252 geometa standav

faselm:


chsdi/models/vector/stopo.py
```diff
@@ -1917,6 +1917,21 @@ class GeometaPNF(Base, Vector):
...
+    __template__ = 'templates/htmlpopup/metadata_standav.mako'
+    __bodId__ = 'ch.swisstopo-vd.geometa-standav'
+    __label__ = 'idn'
+    id = Column('bgdi_id', Integer, primary_key=True)
+    canton = Column('canton', Unicode)
+    idn = Column('id', Unicode)
+    quality = Column('quality', Integer)
+    the_geom = Column('the_geom', Geometry2D)
+
+register('ch.swisstopo-vd.geometa-standav', GeometaStandav)
```
rebert, line 1932: you could use GeometaStandav.__bodId__ instead, not to be redundant... but I know that we did not do it allover...

rebert (approved):
merci!

PR #4571, opened 2025-03-19
BGDIDIC-123: use views

faselm:
I have to use mat views finally...


ltclm (approved):

ltclm:
> Hi, can I merge this pr too? It's also for using the views for the tooltips for the notruf layers

sure, we can always replace matviews with views and vice versa.

ltclm:
thanks  @faselm  yes that's good to know :+1:

PR #4575, opened 2025-03-20
RELEASENOTES deploy-2025-05-14

rebert:
…isenbahnanlagen


ltclm:
@mariokeusen @GeoPhilo 
I've added  some info on the WMTS Swissimage transparency, wms getcap metadata links, and the fix of duplicate cache header.s in wmts responses PB-1597. Let me know what you think, and please feel free to tweak or remove anything that doesn't work!

mariokeusen (approved):

mariokeusen:
@GeoPhilo : since the 3 notruf _142_ layers are actually new layers, they're listed on top of the list as new layers. All other notruf below are actually updated/upgraded

PR #4576, opened 2025-03-24
BGDIDIC-1477: update legends

faselm:


rebert (approved):
merci

ltclm:
@faselm should be good now.

PR #4595, opened 2025-04-10
BGDIDIC-2923: update label field

faselm:


ltclm:
> Hi, I think the label should be changed here for the tooltip to work (bgdi_id can't be both the id and label)

you can set whatever you want as `__label__` . If the label is not set in the model, the primary key will be used.
the label here is used for labeling the features in the tooltip response.

the viewer is choosing the label information in this order:
https://github.com/geoadmin/web-mapviewer/blob/cf0692b5a57a5b4029990f3b17909a0b008a8f55/packages/mapviewer/src/api/features/features.api.js#L481

ltclm (approved):
thank you :+1:

PR #4605, opened 2025-05-09
BGDIDIC-676: use original url from delivered data

ltclm:
for the website url we are using exactly what has been delivered.


rebert (approved):
merci

PR #4609, opened 2025-05-23
BGDIDIC-939 update legend image

marionb:


marionb:
@belisas no they were not delivered in those languages which meas I used the fallback schema for the update. It is the same as what was there before the update.

PR #4633, opened 2025-07-07
BGDIDIC-1751: update model with new field names

faselm:


rebert (approved):
I assume, that you are going to keep the "old" attributes in the table until post deploy?

PR #4642, opened 2025-07-15
PB-1849: update from Buster to Bookworm.

adk-swisstopo:
Debian Buster stopped receiving security updates on [2022-06-30](https://www.debian.org/releases/buster/). It was removed from the main Debian mirror some time between [2025-07-09](https://web.archive.org/web/20250709234633/https://ftp.debian.org/debian/dists/) and 2025-07-15. This means that any build we have that is based on a Buster image and that attempts to run "apt-get" fails.

This change updates the [base image](https://hub.docker.com/_/python) from Buster to [Bookworm](https://www.debian.org/releases/bookworm/).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.9-slim-buster AS builder
+FROM python:3.9-slim-bookworm AS builder
 
 RUN apt-get update -qq \
     && DEBIAN_FRONTEND=noninteractive apt-get install -qq -y  \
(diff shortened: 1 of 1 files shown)


ltshb (approved):

ltshb:
Might be a good idea to also update the python version with its dependencies

adk-swisstopo:
> Might be a good idea to also update the python version with its dependencies

As per discussion, this is more likely to break something and is a little less urgent (Python 3.9 is still supported until October) so let's do that separately.

PR #4657, opened 2025-08-13
BGDIDIC-2865: format null values in haus_nummer attribute

ltclm:
whenever the identify result returns a null value in haus_nummer, the htmlpopup should show a hyphen.


rebert (approved):
merci

PR #4661, opened 2025-08-14
BGDIDIC-2980 Korridor Qualität

julianhafner:
- WMTS fails, therefore tooltip was not tested yet


marionb (commented):
Thanks for the work @julianhafner 

Here two thins to fix you tool-tip. I think with these two changes the CI will also be happy

PR #4676, opened 2025-09-02
PB-1892: Update to Python 3.13

msom:
PR Summary:
- Update to Python to 3.13
- Update all packages to the latest major release
- Keep SqlAlchemy <2 (doesn't seem worth the effort considering the long term plans)
- Remove the now‑obsolete `management` flag from GeoAlchemy2 (it was always False)
- Unpin all development dependencies

What i noticed:
- There are sometimes `sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) SSL connection has been closed unexpectedly` and this also happens on the master branch. I'm not exactly sure why the SSH-tunnel is flaky and how that could be resolved (or if this is even worth the effort investing). Updating to `psycopg 3` is not possible without updating to `SqlAlchemy 2` (which doesn't seem worth the effort).
- The current code is incompatible with `GeoAlchemy 0.18`, a possible fix might be https://github.com/geoadmin/mf-chsdi3/pull/4678. But to be honest, I can't figure out what changed in 0.18 that causes this to fail, but the necessary changes are recommeded anyway https://github.com/geoalchemy/geoalchemy2/issues/568.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -33,7 +33,7 @@ simplejson = "~=3.20"
 unidecode = "~=1.4"
 logging-utilities = "~=5.1"
 pyramid-exclog = "~=1.1"
-GeoAlchemy2 = "<0.18"  # tests are failing, not sure why
+GeoAlchemy2 = "~=0.18"
 setuptools = "<81"  # https://github.com/Pylons/pyramid/issues/3731
 
 [dev-packages]
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "72472240ea7b38836523fc6f2fb1e6f26a49028c074c9454f77baddb27f88565"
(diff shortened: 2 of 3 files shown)


ltshb (approved):
Thanks

Pipfile
```diff
@@ -4,56 +4,56 @@ verify_ssl = true
...
-SQLAlchemy = "~=1.4.48"
-Pillow = "~=9.2"
-polib = "~=1.1"
-psycopg2-binary = "~=2.9"
-pyproj = "==3.6.1"
+geojson = "~=3.2"
+lark_parser = "~=0.12"
+networkx = "~=3.5"
+papyrus = "<2.6"  # requires sqlalchemy>=2
+SQLAlchemy = "<2"  # we don't want to put in the effort to upgrade to 2
```
rebert, line 16: ok

rebert (approved):
merci

PR #4707, opened 2025-10-07
fix: enforce total limit across all models in identify endpoint

ltclm:
The identify endpoint was applying the limit parameter per model/table instead of across the total result set. For layers with multiple registered models (e.g., ch.swisstopo.vec200-transportation-strassennetz with 5 tables), this resulted in returning up to limit×models features instead of the expected total limit.

Changes:
- Modified _identify_db() to calculate total_limit and enforce it
- Updated _get_features_for_filters() to track features_yielded across all models and break early when total limit is reached
- Added remaining_limit calculation to distribute limit capacity across multiple models appropriately

This ensures that when limit=20 is specified, exactly 20 features are returned regardless of how many models/tables are registered for the layer.


ltclm:
as discussed this fix won't be applied.

PR #4720, opened 2025-10-24
BGDIDI-2998: Viewer settings for sora_200

belisas:
Adding the vector file, mako and legends for ch.bazl.intrinsisches-bodenrisiko_sora_200.


belisas:
Question: how do I properly resolve the mentioned conflicts?
I opened the current branch before the one for sora_100 was merged and it seems like there are problems because of that.

faselm (approved):

PR #4721, opened 2025-10-24
PB-1711: run docker image as geodata user

hbollon:
The dockerfile seems already designed to support running the image as `${USER}` (geodata) but final USER instruction was missing. This PR fix that.


rebert (approved):
merci

PR #4727, opened 2025-10-29
BGDIDIC-2837: Viewer settings for ch.vbs.ladestationen

belisas:
Updated (due to attribute changes):
- vector model vbs.py
- mako file vbs_ladestationen.mako

No changes to the legends.


rebert (approved):
merci

PR #4730, opened 2025-11-03
BGDIDIC-3000: convert to hours and minutes

rebert:


ltclm (changes requested):
thank you for that!
it is a nice implementation but im not sure if we should still implement such operations in the htmlpopup service.
i would rather suggest to ask the data owner to deliver publication ready data instead.

PR #4731, opened 2025-11-03
BGDIDIC-2990: ch.swisstopo.swisseo_ndvi_diff_v100 add legends

ltclm:


faselm (commented):
ch.swisstopo.swisseo_ndvi_z_v10_de.png seems at odd and unable to be displayed

PR #4734, opened 2025-11-05
BGDIDIC-1457: update legends

ltclm:


faselm (approved):
"Unable to render rich display" so I can't check the legends

PR #4841, opened 2026-04-29
BGDIDIC-1998: List entries as string in mako

belisas:


marionb:
@belisas the last commit does the trick. It is quite helpful to have a debugger ;-)

marionb (approved):
Code looks good now and tested it
👍

julianhafner (approved):
Looks good to me 👍

PR #4843, opened 2026-05-12
PB-2302: adapt the public transport stops to SLOID

hansmannj:
This needs to be deployed until beginning of June so that the public transport stop information requests will work after that date (when the OJP 2.0 API we use for that switched to SLOID support only).

Introduce a two-step LIR (location information request) + SER (stop event request) flow for public transport stationboard: numeric DiDok station IDs are first resolved to a SLOID via an OJP LocationInformationRequest, then the stop event request is made with the resolved identifier
- Use `PlaceRef/StopPointRef` in the LIR request for a deterministic ID-based lookup instead of a fuzzy name search
- Handle ythe case gracefully. where the LIR retursn DiDok instead of SLOID. This will most likely be the case untin beginning of June, until the OJP 2.0 API switched to SLOID: if the LIR returns a DiDok instead of a SLOID (API not yet upgraded), use it directly for SER. Once it returns SLOIDs, use those.
- Update and consolidate test mocks to reflect the two-step flow, rename helpers for clarity (`generate_mock_ser_response`, `generate_mock_lir_response`)


ltclm (changes requested):
Thanks @hansmannj! Just two comments:

1. **`test_stationboard_invalid_id` passes for the wrong reason now.** `"invalid_id"` has no `:`, so it hits LIR first — the single SER-shaped mock gets consumed there, parsing fails, and `OpenTransNoStationException` is raised from LIR, not SER. Could you add a proper LIR mock (or rename the test)?

2. **No test for the DiDok-fallback case.** The PR mentions LIR will keep returning DiDoks until June, but nothing covers it. Mind adding one where the LIR mock returns a colon-less ID and we assert SER is called with it (and ideally the warning is logged)?

ltclm (approved):
thank you 👍

boecklic (approved):
Thanks 👌 . 
If this needs to be deployed by beginning of Juni it should probably be merged into master directly? The date of the develop-deploy branch is July first...

PR #4856, opened 2026-07-06
swissgeocover2d: adding big legend PDFs and image overview (for legend)

procrastinatio:
As some PDF are quite big (even compressed), feel free to store them elsewhere, if too massive.


chsdi/static/images/legends/ch.swisstopo.geologie-swissgeocover2d_unconsolidated_fr_big.pdf
```diff
@@ -0,0 +1 @@
+ch.swisstopo.geologie-swissgeocover2d_unconsolidated_de_big.pdf
```
rebert, line 1: this looks like a text file ...

chsdi/static/images/legends/ch.swisstopo.geologie-swissgeocover2d_surfaces_en_big.pdf
```diff
@@ -0,0 +1 @@
+ch.swisstopo.geologie-swissgeocover2d_surfaces_de_big.pdf
```
rebert, line 1: There is no redicrect. You have to put the de version of the pdf here with the name ...surface_en_big.pdf

rebert (changes requested):
It looks almost good for me. I would put big.pdf allover and not text files called .pdf. There is no redirect to another file... 

If the document does not exist in a language I would make a copy f.ex. the _de_big.pdf copied to _rm_big.pdf

chsdi/static/images/legends/ch.swisstopo.geologie-swissgeocover2d_unconsolidated_fr_big.pdf
procrastinatio: @rebert  Apparently, this should create a symlink. The actual and only real file is `swissgeocover2d_bedrock_de_big.pdf‎`, the `rm`, `fr` and `en` version should reuse this file. Same for the other legends.

rebert (approved):
lgtm

PR #4858, opened 2026-07-20
BGDIDIC-124: remove che_date field

faselm:


ltclm (approved):
thank you, if needed, please also remove it from the translations/items table in the bod and from the wms getfeaturinfo 👍

PR #4867, opened 2026-08-18
BGDIDIC-3044: new legend PDF for swissgeocover2d

procrastinatio:
BGDIDIC-3044: updated  PDF legend for swissgeoocver2d and add missing Italian. Only German are real PDF, all other languages are symlinks to them.


rebert (approved):
thx

PR #4874, opened 2026-09-10
BGDIDIC-698: Create a Test env for Updates on BFE Side.

marionb:
This PR sets up new test URLs so that the BFE can test updates on this json layer.

NOTS: this PR should be reverted before the next deploy. It is only for test


rebert (approved):
ok, for testing. I'm happy with the "featureattachment" path. But we have to remember to remove the /test

ltclm (approved):
thank you, please already prepare the revert pr on the next milestone branch so that we dont forget to do the rollback.

marionb:
**NOTE:** this will revert this PR #4875

